PR #6, opened 2019-09-24
Add browser field and remove withAuth0

ijjk:
### Description

> Describe the purpose of this PR along with any background information and the impacts of the proposed change. For the benefit of the community, please do not assume prior context.

This PR adds usage of the `browser` field to `package.json` to allow using the `instance.browser.js` for browser builds. This removes the need for `withAuth0` build time configuration. It also moves `next` as a `peerDependency`/`devDependency` and removes `react` and `react-dom` as dependencies since they are not needed

> Provide details that support your chosen implementation, including: breaking changes, alternatives considered, changes to the API, etc.

No breaking changes as the `withAuth0` method has been left as a no-op with a deprecated comment. 

### Testing

> Describe how this can be tested by reviewers. Be specific about anything not tested and reasons why. If this library has unit and/or integration testing, tests should be added for new functionality and existing tests should complete without errors.

Remove `withAuth0` from a Next.js application using `@auth0/nextjs-auth0` and see that it works without.

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "eslint-import-resolver-typescript": "^1.1.1",
     "eslint-plugin-import": "^2.18.2",
     "jest": "^24.9.0",
+    "next": "^9.0.5",
     "nock": "^11.3.3",
     "prettier": "^1.18.2",
     "request": "^2.88.0",
@@ -66,10 +67,10 @@
     "@hapi/iron": "^5.1.1",
     "base64url": "^3.0.1",
     "cookie": "^0.4.0",
-    "next": "^9.0.5",
-    "openid-client": "^3.6.2",
-    "react": "^16.9.0",
-    "react-dom": "^16.9.0"
+    "openid-client": "^3.6.2"
+  },
+  "peerDependencies": {
+    "next": "^9.0.5"
   },
   "jest": {
     "testEnvironment": "node",
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "Next.js SDK for signing in with Auth0",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
-  "browser": "dist/instance.browser.js",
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -11,18 +11,10 @@ export function useAuth0(settings: IAuth0Settings): ISignInWithAuth0 {
   return require('./instance.node').default(settings);
 };
 
+/**
+ * @deprecated this is now a no-op and is no longer required
```
timneutkens: There's no need for deprecation considering it's not being used yet

package.json
```diff
@@ -54,6 +55,7 @@
     "eslint-import-resolver-typescript": "^1.1.1",
     "eslint-plugin-import": "^2.18.2",
     "jest": "^24.9.0",
+    "next": "^9.0.5",
```
timneutkens: Good catch!

package.json
```diff
@@ -4,6 +4,7 @@
   "description": "Next.js SDK for signing in with Auth0",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
+  "browser": "dist/instance.browser.js",
```
timneutkens: This should actually link to a file that implements the same exports as the `main` file, except for the browser.

ijjk: Updated

src/index.browser.ts
```diff
@@ -0,0 +1,7 @@
+/* eslint-disable */
+import IAuth0Settings from './settings';
+import { ISignInWithAuth0 } from './instance';
+
+export function useAuth0(settings: IAuth0Settings): ISignInWithAuth0 {
+  return require('./instance.browser').default(settings);
```
timneutkens: This `require` could just be an import right?

ijjk: Added in follow up PR https://github.com/auth0/nextjs-auth0/pull/9

PR #10, opened 2019-09-30
Rename useAuth0 to initAuth0 since it's not a hook

ijjk:
### Description

> Describe the purpose of this PR along with any background information and the impacts of the proposed change. For the benefit of the community, please do not assume prior context.
>
> Provide details that support your chosen implementation, including: breaking changes, alternatives considered, changes to the API, etc.
>
> If the UI is being changed, please provide screenshots.

Renames `useAuth0` to `initAuth0` to prevent confusion/conflicts with react hooks

### References

> Include any links supporting this change such as a:
>
> - GitHub Issue/PR number addressed or fixed
> - Auth0 Community post
> - StackOverflow post
> - Support forum thread
> - Related pull requests/issues from other repos
>
> If there are no references, simply delete this section.

https://github.com/zeit/next.js/pull/8802#discussion_r329375613

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,9 +38,9 @@ yarn add @auth0/nextjs-auth0
 And then create an instance of the Auth0 plugin (eg: under `/utils/auth0.js`):
 
 ```js
-import { useAuth0 } from '@auth0/nextjs-auth0';
+import { initAuth0 } from '@auth0/nextjs-auth0';
 
-export default useAuth0({
+export default initAuth0({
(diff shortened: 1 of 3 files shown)


sandrinodimattia:
Could you leave the existing export there too with a deprecation notice? We don't want to break people currently using the library.

ijjk:
Updated to deprecate `useAuth0` instead of removing it completely 👍

Issue #18, opened 2019-10-09
Handle clock skew with leeway option


qburnsFB:
### Description

Getting this error when running from my local server:

> id_token issued in the future, now 1570650460, iat 1570650461

Running on a pushed production server works as expected.

According to the [Auth0 ref](https://auth0.com/docs/libraries/auth0js/v9), we can set a leeway integer to deal with small amount of clock skew. However, this library doesn't seem to support it as an option.

I saw a [commit merged earlier today](https://github.com/auth0/nextjs-auth0/pull/13) that solved a similar issue with the audience option.  I tried following in their footsteps by adding settings.leeway to the handlers/login.js. I see it correctly added to the url when logging in, yet I still receive the error.

### Reproduction

Possibly hard to reproduce unless you manually change the time to be incorrect.

### Environment

> Please provide the following:

- **Version of this library used:** 0.3
- **Version of the platform or framework used, if applicable:** Next 9.0.5
- **Other relevant versions (language, server software, OS, browser):** Windows 10
- **Other modules/plugins/libraries that might be involved:** Running a custom HTTPS server for local development, but was working perfectly fine locally as of Friday afternoon.

qburnsFB:
Update: If I manually change the following line in:  `/node_modules/openid-client/lib/client.js` , I can get my app to work as expected.

this[CLOCK_TOLERANCE] = 0; **changed to**    this[CLOCK_TOLERANCE] = 30;

Obviously this is extremely hacky. Any ideas on how to resolve this?

qburnsFB:
Awesome! It seems to be working in my project. Thanks for the quick response and fix!

PR #20, opened 2019-10-10
Support id_token leeway

sandrinodimattia:
### Description

Add support for leeway when validating the `id_token`. This is necessary when the time on your server is running behind on Auth0.

### References

https://github.com/auth0/nextjs-auth0/issues/18


README.md
```diff
@@ -252,6 +254,26 @@ If the user is authenticated then your API route will simply execute, but if the
...
+```js
+import { initAuth0 } from '@auth0/nextjs-auth0';
+
+export default initAuth0({
+  ...
+  session: {
+    ...
+  },
+  oidcClient: {
+    clockTolerance: 10 // In seconds.
```
tusbar: Shouldn’t this be in milliseconds?

qburnsFB: According to [their docs](https://auth0.com/docs/libraries/auth0js/v9), it is in seconds.


> leeway (Integer) A value in seconds; leeway to allow for clock skew with regard to JWT expiration times.

sandrinodimattia: Correct, should be milliseconds to have a consistent way to express timespan (eg: on httpTimeout it's also MS).

sandrinodimattia: I updated the readme, thanks for letting me know.

PR #39, opened 2019-11-02
User permissions

### LINKED ISSUE #29: Ability to check permissions

> The library provides some helper functions like requireAuthentication for api routes, but currently there is no good way to check permissions from what I can tell.

> I'd like a helper function to secure api routes based on the current users permissions. It can be used in conjunction with requireAuthentication. Something like:

```
// Route requires authentication
export default auth0.requireAuthentication(
  // Check for permissions
  auth0.checkPermissions(
    async function user(req, res) {
      try {
        await auth0.handleProfile(req, res);
      } catch (error) {
        console.error(error);
        res.status(error.status || 500).end(error.message);
      }
    },
    // Permissions to be checked - If the defined permissions are not found, the request would be denied.
    ["view:profile"]
  )
);
```

> I've been able to implement a solution to decode session tokens and verify permissions using jsonwebtoken & jwks-rsa, but would like to have something incorporated into the nextjs-auth0 library.

karaggeorge:
By submitting a PR to this repository, you agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md). Please see the [contributing guidelines](https://github.com/auth0/open-source-template/blob/master/CONTRIBUTING.md) for how to create and submit a high-quality PR for this repo.

### Description

Fixes #29 

Adds a handler similar to `requireAthentication`: `requirePermissions`.

```js
export default requirePermissions(
	(req, res) => {...},
	['create:posts', 'edit:posts'],
	{ checkAllScopes: true }
);
```

Haven't used Auth0 scopes before, but when I created some roles and decoded the token it looks like it has two relevant fields:

```
{
	// ...
	scope: 'openid profile some:permission',
	permissions: ['some:permission', 'some:otherpermission']
}
```

Looks like the ones in the scope are the ones requested through the config, but the `permissions` field contains all the permissions included that user's role.

I saw that this package: https://github.com/Aulos/micro-jwt-authz used `scope`, so I used it here as well, but added the permissions as well.

### Testing

Haven't added tests yet. Waiting to see if this implementation is acceptable, and when it's agreed upon, I'll add the tests

### Checklist

Waiting for some discussion/approval on the implementation before I add the docs as well

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,8 @@
     "@hapi/iron": "^5.1.4",
     "base64url": "^3.0.1",
     "cookie": "^0.4.0",
+    "jsonwebtoken": "^8.5.1",
+    "jwks-rsa": "^1.6.0",
     "openid-client": "^3.7.4"
   },
   "peerDependencies": {
--- a/src/handlers/index.ts
+++ b/src/handlers/index.ts
@@ -4,6 +4,7 @@ import CallbackHandler from './callback';
 import ProfileHandler from './profile';
 import SessionHandler from './session';
 import RequireAuthentication from './require-authentication';
+import RequirePermissions from './require-permissions';
 
 export default {
   CallbackHandler,
@@ -11,5 +12,6 @@ export default {
   LogoutHandler,
   ProfileHandler,
   SessionHandler,
-  RequireAuthentication
+  RequireAuthentication,
+  RequirePermissions
 };
--- a/src/handlers/require-permissions.ts
+++ b/src/handlers/require-permissions.ts
@@ -1,0 +1,83 @@
+import { promisify } from 'util';
+import { NextApiResponse, NextApiRequest } from 'next';
+import jwt from 'jsonwebtoken';
+import jwksClient, { CertSigningKey, RsaSigningKey } from 'jwks-rsa';
+import { URL } from 'url';
+
+import { ISessionStore } from '../session/store';
+import IAuth0Settings from '../settings';
+
+export interface IApiRoute {
+  (req: NextApiRequest, res: NextApiResponse): Promise<void>;
+}
+
+export default function requirePermissions(settings: IAuth0Settings, sessionStore: ISessionStore) {
+  return (
+    apiRoute: IApiRoute,
+    expectedScopes: string[],
+    options = {
+      checkAllScopes: false
+    }
+  ): IApiRoute => async (req: NextApiRequest, res: NextApiResponse): Promise<void> => {
+    if (!req) {
+      throw new Error('Request is not available');
+    }
+
+    if (!res) {
+      throw new Error('Response is not available');
+    }
+
+    if (!settings.audience) {
+      throw new Error('Missing audience');
+    }
+
+    if (!Array.isArray(expectedScopes)) {
+      throw new Error('Parameter expectedScopes must be an array of strings representing the scsopes of the endpoint');
+    }
+
+
+    const session = await sessionStore.read(req);
+    if (!session || !session.user) {
+      res.status(401).json({
+        error: 'not_authenticated',
+        description: 'The user does not have an active session or is not authenticated'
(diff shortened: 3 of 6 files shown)


BjoernRave:
Would like to see this merged :)

BjoernRave:
Without this PR there is no way to have different accounts with different permissions, right?

karaggeorge:
@BjoernRave @sandrinodimattia 

I can update this PR with the recent changes and PRs that got merged if that's desired

BjoernRave:
@karaggeorge yea, that would be much appreciated :)

sandrinodimattia:
Before we continue with this PR I would like use to define the problems this is solving. Will API Routes really need this or are they typically just proxies to other places? (eg: GraphQL, REST APIs, ...)?

And should we support roles or permissions? Or both? Could we support roles and then you map them to permissions in your app? That might scale better (having a handfull of roles in your session as opposed to tens if not hundreds of permissions).

Can we first have an end to end example working showing the scenario?

karaggeorge:
@sandrinodimattia 

I agree. Like I mentioned in the initial PR, I haven't used either of these, so I don't know what the use case is. Once there's a clear direction I can either update this, or open a new one based on the recent changes

BjoernRave:
maybe I understand this wrong, but I would like to restrict access to some of my pages based on the permissions this user has. If I do `auth0.getSession(ctx.ctx.req)` there is no information about the users permissions

BjoernRave:
I was re-checking this PR and just realized, that it's not really usable with a grapqhl endpoint, because you will have one endpoint, but the required permissions would change depending on the query. Is there not a way to get access to the permissions through `auth0.getSession()` ?

9bits:
Any updates? Looks like the change was never merged into the master branch

FMGordillo:
Still waiting :c

glenmcrae:
Any updates on this one? Would like to see this merged. Is this repo being looked at by Auth0 staff? I am concerned about using Auth0 after seeing this package with a tonne of open issues and unmerged PR's.

karaggeorge:
Hi, it's been a while since I opened this PR, and I think the implementation of the project has changed a bit. I can look into it again and update it, but before I put in the effort again, I'd like to have some confirmation that this is a real need in the form that it is, or if not, what exactly do people need and what would the maintainers be willing to accept.

Referring to this comment above:

> Before we continue with this PR I would like use to define the problems this is solving. Will API Routes really need this or are they typically just proxies to other places? (eg: GraphQL, REST APIs, ...)?

> And should we support roles or permissions? Or both? Could we support roles and then you map them to permissions in your app? That might scale better (having a handfull of roles in your session as opposed to tens if not hundreds of permissions).

> Can we first have an end to end example working showing the scenario?

I personally didn't end up using this package directly (or Auth0), but used it as inspiration to do similar auth in Next.js. So, I don't know what information would be useful or not and in what format. If that is made clear however, I'd be glad to implement the changes 😄

adamjmcgrath:
Hi @karaggeorge  - thanks for your contribution

We've have a new Beta out and have decided that protecting a Next.js API Route with a JWT is out of scope for this library. 

But we have included an example of how you can do this with a 3rd party library here: https://github.com/auth0/nextjs-auth0/blob/beta/examples/kitchen-sink-example/pages/api/my/shows.ts

xadamxk:
> Hi @karaggeorge - thanks for your contribution
> 
> We've have a new Beta out and have decided that protecting a Next.js API Route with a JWT is out of scope for this library.
> 
> But we have included an example of how you can do this with a 3rd party library here: https://github.com/auth0/nextjs-auth0/blob/beta/examples/kitchen-sink-example/pages/api/my/shows.ts

Is there a way to use this library to retrieve roles now? Or is the 3rd party library still the only option?

PR #41, opened 2019-11-03
Support sending custom `state` via login handler

### LINKED ISSUE #36: Redirect after login
### Redirect users after login

As described in https://auth0.com/docs/users/guides/redirect-users-after-login

Is it possible to redirect a user to a defined page after login, and if so, how would you do it with this library? 
My use case is I have a user who tried to access a specific page which required authentication, and I want to redirect back to that page after login.

On the client-side I would have local stored a nonce, and retrieved it on the callback. What is the recommended way of doing this in the Next API - I believe they only run server-side?

claus:
### Description

Fixes #36 

This PR enables sending a custom `state` to the auth server. This way a redirect url can be stored as described in https://auth0.com/docs/protocols/oauth2/redirect-users and then retrieved in the callback handler in order to redirect users back to the originating url.

Previously, the state was created inside the login handler, with no possibility to access it from the outside. Now, the state can optionally be passed via `authParams`.

```js
import auth0 from '../../utils/auth0';
export default async function login(req, res) {
  try {
    await auth0.handleLogin(req, res, {
      authParams: {
        state: 'a custom state'
      }
    });
  } catch(error) {
    console.error(error)
    res.status(error.status || 400).end(error.message)
  }
}
```

Code in this PR has no breaking changes.

### Testing

- [x] This change adds test coverage for new/changed/fixed functionality (see `'should allow sending custom state to the authorization server'` in `login.test.ts`)

### Checklist

- [x] I have added documentation for new/changed functionality in this PR (see `README.md`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,7 @@ You can optionally send extra parameters to Auth0 to influence the transaction, 
 - Showing the login page
 - Filling in the user's email address
 - Exposing information to the custom login page (eg: to show the signup tab)
+- Using a custom `state`
 
 ```js
 import auth0 from '../../utils/auth0';
@@ -140,6 +141,7 @@ export default async function login(req, res) {
         login_hint: 'foo@acme.com',
         ui_locales: 'nl',
         scope: 'some other scope',
+        state: 'a custom state',
         foo: 'bar'
       }
     });
--- a/src/handlers/login.ts
+++ b/src/handlers/login.ts
@@ -30,6 +30,7 @@ export interface AuthorizationParameters {
   max_age?: string;
   prompt?: string;
   scope?: string;
+  state?: string;
   ui_locales?: string;
   [key: string]: unknown;
 }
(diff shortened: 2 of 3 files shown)


aorsten:
With this in place I believe I could migrate to this library - hope it goes through soon.

aorsten:
Any updates here by any maintainers? @sandrinodimattia

karaggeorge:
Should this follow the article's directions? So, instead of actually passing the state in the way it receives it (as a string), put it in an object, and encode it using a secret and then pass it on?

So, passing in a state can actually be an object, instead of just one url.

Or does this assume that the user of the package has already done that and they are passing in the encoded state?

sandrinodimattia:
@karaggeorge I would argue that serializing and deserializing objects to/from state is beyond the scope of this library. There's multiple ways to do this (base64url, semicolon delimited, ...) which I would leave up to the consumer of this library to handle.

aorsten:
Will you publish the newest version to npm? @sandrinodimattia

bromanko:
Am I reading it correct that this PR should have also updated [`ISigninWithAuth0`](https://github.com/auth0/nextjs-auth0/blob/fac8bf7d62ef8ac34bbef0cc66004b95b0acd8ea/src/instance.ts#L12) but it was missed?

claus:
Yeah, the `handleLogin` declaration in that interface is missing the third `options: LoginOptions` argument. It was missing before this PR already though.

bromanko:
OK - I can open a PR if you aren't already planning to do so.

claus:
Please do! 👍

mintuz:
How does this work with handleCallback, I am getting the error `state mismatch, expected K5y-1aEodfPB4tV3rdM9opSJLoV1ewyespIPowkj-MyfBO1fxqeEWLZp6dPnylQY, got:custom-state`

PR #38, opened 2019-11-02
Ability to refetch the user

karaggeorge:
By submitting a PR to this repository, you agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md). Please see the [contributing guidelines](https://github.com/auth0/open-source-template/blob/master/CONTRIBUTING.md) for how to create and submit a high-quality PR for this repo.

### Description

Fixes #23 

Allows an option to refetch the user, which also updates it on the session

### Testing

- [x] This change adds test coverage for new/changed/fixed functionality

### Checklist

- [x] I have added documentation for new/changed functionality in this PR or in auth0.com/docs
- [x] All active GitHub checks for tests, formatting, and security are passing
- [ ] The correct base branch is being used, if not `master`

PR #37, opened 2019-11-02
Refresh token automatically

karaggeorge:
By submitting a PR to this repository, you agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md). Please see the [contributing guidelines](https://github.com/auth0/open-source-template/blob/master/CONTRIBUTING.md) for how to create and submit a high-quality PR for this repo.

### Description

Fixes #7 (part of it)

When the refresh_token is saved in the session, at any point we try to read the session from the cookie, if the token has expired (also saving expires_at), we refresh the token, save the new session and return it.

The rest of the code doesn't need to have any idea of what happened

#### Parts that weren't implemented:

>  If the token is missing a scope, return null. If the token is for an other audience, return null.

I only focused on the refresh token functionality, no scope/audience checks

### Testing

- [x] This change adds test coverage for new/changed/fixed functionality

### Checklist

- [ ] I have added documentation for new/changed functionality in this PR or in auth0.com/docs
- [x] All active GitHub checks for tests, formatting, and security are passing
- [ ] The correct base branch is being used, if not `master`

----

I haven't added doc changes yet. I'm not sure if this needs its own setting (`autoRefresh`?) or if it's just something built in.

There wasn't much discussion on the issue, so I'll make the doc updates when I hear more here

PR #56, opened 2019-12-17
refresh token and refresh user cleanup integration

nodabladam:
Combines #37 and #38 for the purposes of helping with merge issues and fixing up some tests that are failing.  These PRs are critical for us.  Please merge so the community can continue to test this concept.   

### Testing
Tested using a manually published module: `npm install nextjs-auth0-testing`

### Checklist


src/handlers/profile.ts
```diff
@@ -21,6 +30,25 @@ export default function profileHandler(sessionStore: ISessionStore) {
       return;
     }
 
+    if (options && options.refetch) {
+      if (!session.accessToken) {
+        throw new Error('The access token needs to be saved in the session for the user to be fetched');
+      }
+
+      const { body: user } = await getAsync({
```
sandrinodimattia: All OIDC integrations should ideally be called through the OIDC client.

Could you do this instead:

```js
const client = await clientProvider();
const userInfo = await client.userinfo(access_token);
```

Here is an example in the callback handler: https://github.com/auth0/nextjs-auth0/blob/master/src/handlers/callback.ts#L37

src/session/cookie-store/index.ts
```diff
@@ -31,20 +36,39 @@ export default class CookieSessionStore implements ISessionStore {
       return null;
     }
 
+    const { expiresAt, refreshToken } = unsealed as ISession;
+
+    // Check if the token has expired
+    if (refreshToken && expiresAt && expiresAt * 1000 < Date.now()) {
```
sandrinodimattia: Can we  play it a bit safer here? There is an edge case where we might have some clock skew where our code assumes the token is still valid but in reality it's not.

Could we for example add a skew of 1 minute? Where instead of checking `expiresAt` we check `expiresAt - 1 minute`?

src/session/cookie-store/index.ts
```diff
@@ -60,12 +84,17 @@ export default class CookieSessionStore implements ISessionStore {
       persistedSession.refreshToken = refreshToken;
     }
 
+    if (this.settings.storeRefreshToken && refreshToken && expiresAt) {
```
sandrinodimattia: I think storing the `expiresAt` should be unrelated to the refresh token being present or not. The expiration of the access token can be tracked always, even if we don't store the RT.

src/handlers/profile.ts
```diff
@@ -21,6 +30,25 @@ export default function profileHandler(sessionStore: ISessionStore) {
...
+        baseUrl: `https://${settings.domain}`,
+        url: 'userinfo',
+        json: true,
+        headers: {
+          Authorization: `Bearer ${session.accessToken}`
+        }
+      });
+
+      await sessionStore.save(req, res, { ...session, user });
```
sandrinodimattia: Can we merge the current user in the session with the new claims we get from userinfo instead of replacing the whole user? There are cases where some claims are in the id_token but when they are not returned as part of userinfo. If that happens, we don't want to break the current session. With userinfo we just want to make sure we can get the newest info like updated email address or email_verified etc... and update that in the session.

src/handlers/session.ts
```diff
@@ -1,14 +1,14 @@
-import { IncomingMessage } from 'http';
+import { NextApiResponse, NextApiRequest } from 'next';
 
 import { ISession } from '../session/session';
 import { ISessionStore } from '../session/store';
 
 export default function sessionHandler(sessionStore: ISessionStore) {
-  return (req: IncomingMessage): Promise<ISession | null | undefined> => {
+  return (req: NextApiRequest, res: NextApiResponse): Promise<ISession | null | undefined> => {
```
sandrinodimattia: This is a breaking change. Can we make `res` optional somehow?

nodabladam: `CookieSessionStore` calls save which seems to require it.  I am not sure we can avoid that breaking change.

nodabladam:
I got it passing the tests but I am not at all sure I did it right.

src/handlers/profile.ts
nodabladam: I am not sure how to merge claims.  Could you help with that?

raguiar2:
Is this going to merge soon? I can't use auth0 without it as I need a working refresh token implementation.

src/handlers/profile.ts
nodabladam: @sandrinodimattia Could you help with the merging of claims?  I think that is the only thing holding this back.  It would be great to get this merged in so everyone could benefit from it.

sandrinodimattia: Ok looking into it!

sandrinodimattia:
Thanks for your contribution @nodabladam. I made a few changes to your PR to use the token cache and I then merged it as a separate PR (I didn't have access to push to yours). I also added the implementation to merge the existing user claims with the updated ones we get from user info.

This is now live in v0.10.0

PR #62, opened 2020-01-08
Update README.md

vassbence:
Since v0.7.0 an options parameter is required for the handleLogin() fn.

Fixed the error in the README.


sandrinodimattia:
Thanks for your PR @vassbence! This was not by design. I pushed v0.9.0 which makes the `options` parameter optional so we can go ahead and close this PR.

PR #68, opened 2020-01-16
chore: update dependencies

panva:
### Description

This PR updates some of the dependencies

1) it updates the development dependency use of `@panva/jose` to just `jose` (package got renamed)
2) it updates the minimal required version of `openid-client` to 3.11.0 ([changelog](https://github.com/panva/node-openid-client/compare/v3.7.4...v3.11.0)), the importance of this update is that ID Token's `iat` is no longer being checked to be in the past, this is unnecessary for an ID Token profile.

cc @sandrinodimattia


adamjmcgrath:
Superseded by #244

PR #81, opened 2020-02-17
Add redirectTo support

nicogarcia:
### Description
Adds support for custom redirect url at the beginning of the login transaction.

The `handleLogin` method now receives an optional `redirectTo` parameter and writes it to an `a0:redirectTo` cookie in the same way as `state` is written. Then, if the `handleCallback` method doesn't receive a `redirectTo` parameter directly, it will redirect to the value stored in the `a0:redirectTo` cookie.

#### Previously evaluated alternatives 
- Write the cookie outside of the SDK: this is not possible since the SDK sets the `Set-Cookie` header and overrides any previously set value.
- Use state to encode redirect url: this is non-trivial to do for users

### Testing

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,7 @@ You can optionally send extra parameters to Auth0 to influence the transaction, 
 - Filling in the user's email address
 - Exposing information to the custom login page (eg: to show the signup tab)
 - Using a custom `state`
+- Redirecting the user to a `redirectTo` url after the transaction is finished
 
 ```js
 import auth0 from '../../utils/auth0';
@@ -148,7 +149,8 @@ export default async function login(req, res) {
         scope: 'some other scope',
         state: 'a custom state',
         foo: 'bar'
-      }
+      },
+      redirectTo: '/custom-url'
     });
   } catch (error) {
     console.error(error);
--- a/src/handlers/callback.ts
+++ b/src/handlers/callback.ts
@@ -46,8 +46,8 @@ export default function callbackHandler(
     // Create the session.
     await sessionStore.save(req, res, session);
 
-    // Redirect to the homepage.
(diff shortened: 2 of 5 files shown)


nicogarcia:
I noticed that since the `redirectTo` value is not required if we use it in one call and then don't include it, the cookie will still be there if not expired.

Alternative solutions:
- Always set (with a default) like `a0:state`
- Clean the cookie in `handleCallback`

balazsorban44:
This would be a much needed, great addition! :slightly_smiling_face:

joeweale:
@sandrinodimattia Sorry to be a pain, can we get this PR reviewed? Would be a great addition and save a fork.

benjamintd:
This would indeed be a very useful addition. Thanks @nicogarcia for the PR !

BjoernRave:
@joeweale I already tried contacting @sandrinodimattia before, he doesn't seem to be around...

joeweale:
I've ended up handling this locally instead using Auth0 recommendation;

https://auth0.com/docs/protocols/oauth2/redirect-users

balazsorban44:
@joeweale do you maybe have a code example?🤞

joeweale:
I'll try and seperate it out Monday for you 👍

joeweale:
@balazsorban44 take a look at this gist;

https://gist.github.com/joeweale/e609e160eb31884211ab2ad605447ff1

included some bits I did to return a state to SSR which may be useful as well. 

Idea was; Complete data entry client side -> Login before submit -> on redirect submit previously entered client data

sandrinodimattia:
Apologies for the delay here! ✋ We'll try to push this out to NPM today

sandrinodimattia:
This was released under v0.11.0 yesterday

PR #92, opened 2020-03-31
Automatically redirect to redirectTo in login request

balazsorban44:
### Description
The `redirectTo` query string param from `req` is automatically infered, so this:

```js
auth0.handleLogin(req, res, {
    redirectTo: req.query.redirectTo
})
```

Can become this:

```js
auth0.handleLogin(req, res)
```

### References

This is supposed to be an improvement on #81.

### Testing

I can try adding a test for this, if necessary.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ You can optionally send extra parameters to Auth0 to influence the transaction, 
 - Filling in the user's email address
 - Exposing information to the custom login page (eg: to show the signup tab)
 - Using a custom `state`
-- Redirecting the user to a `redirectTo` url after the transaction is finished
+- Redirecting the user to a `redirectTo` url after the transaction is finished (`redirectTo` is infered from `req` by default. If not present there either, the default is `'/'`.)
 
 ```js
 import auth0 from '../../utils/auth0';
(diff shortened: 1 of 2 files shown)


balazsorban44:
Would it also be a good idea for security's sake if `redirectTo` that comes from `req.query` is sanitized first, so only a relative path is allowed? Or is it already done somewhere?

balazsorban44:
ping @sandrinodimattia :)

sandrinodimattia:
Thanks for your PR. It was slightly more work to support this use case, so I'll be closing this one. The final implementation is here: https://github.com/auth0/nextjs-auth0/commit/ad5487ab31bd230e1a9696dfb81e1eeca4b99ed3

And was just published as `v0.13.0`

The basic example was also updated, `withAuth` will capture the current path and add it to the redirect to `/api/login`:

- https://github.com/auth0/nextjs-auth0/blob/master/examples/basic-example/components/with-auth.jsx
- https://github.com/auth0/nextjs-auth0/blob/master/examples/basic-example/components/login-redirect.jsx
- https://github.com/auth0/nextjs-auth0/blob/master/examples/basic-example/pages/protected-page.jsx#L26

balazsorban44:
Awesome! Just a question. Will it also preserve hashes in the url? Eg. if I am at http://exanple.com/foo#bar, will I be redirected to exactly the same place, or #bar is removed?

PR #98, opened 2020-04-12
Updated README to match session guards in examples

TillaTheHun0:
### Description

Just an update to the README.

The current example in README does not have sufficient guards on calls to `auth0.getSession()`. I assumed it was copy/pasta code, but it can produce an error out of the box when `getSession()` [returns null](https://github.com/auth0/nextjs-auth0/blob/e724719f7be191e122ba664b16d07926aa18d980/src/session/cookie-store/index.ts#L30). I just matched the README to what is in the example projects, which _does_ use a guard.


sandrinodimattia:
Looks good, thank you

PR #102, opened 2020-05-08
Fix #100 - Use end_session_endpoint if available from issuer

### LINKED ISSUE #100: Using client.endSessionUrl instead of hardcoded url for logout

I see that logout handler has hardcoded url, and not the one that could be generated from the client.endSessionUrl. If that would be the case, then it could be used for various providers. I am wondering would you have something like that in mind, or maybe you would prefer that it stays coupled to this specific logout endpoint of auth0 implementation? 

Here is a quick draft of what could be achieved.
createLogoutUrl could be modified to make use of client.endSessionUrl, or to be replaced by it.
```
function createLogoutUrl(settings: IAuth0Settings, client: Client, idToken: string): string {
  return client.endSessionUrl({
    id_token_hint: idToken,
    post_logout_redirect_uri: encodeURIComponent(settings.postLogoutRedirectUri)
  });
}
```
In handlers/logout.ts params that this modified function would need could be declared :
```
    const client = await clientProvider();
    const tokenCache = tokenCacheHandler(clientProvider, sessionStore)(req, res);
    const { idToken } = await tokenCache.getIdToken();
    if (!idToken) {
      throw new Error('No id token available to logout user');
    }
```

In tokens/session-token-cache.ts could have additional function for returning idToken, e.g.
```
  async getIdToken(): Promise<IdTokenResponse> {
    const session = await this.store.read(this.req);
    if (!session) {
      throw new IdTokenError('invalid_session', 'The user does not have a valid session.');
    }
    if (!session.idToken) {
      throw new IdTokenError('invalid_session', 'The user does not have a valid id token.');
    }
    return {
      idToken: session.idToken
    };
  }
```

reicheltp:
### Description

Solves the problem described in #100.

This PR enhance `auth0.handleLogout(req, res)` behavior by using the `end_session_endpoint` from discovery endpoint if available. This still falls back to use the default Auth0 logout URL, thus existing behavior is not broken.

### References

- https://openid.net/specs/openid-connect-session-1_0.html#RPLogout

### DIFF
--- a/src/handlers/logout.ts
+++ b/src/handlers/logout.ts
@@ -2,7 +2,9 @@ import { IncomingMessage, ServerResponse } from 'http';
 
 import IAuth0Settings from '../settings';
 import { setCookies } from '../utils/cookies';
+import { IOidcClientFactory } from '../utils/oidc-client';
 import CookieSessionStoreSettings from '../session/cookie-store/settings';
+import { ISessionStore } from '../session/store';
 
 function createLogoutUrl(settings: IAuth0Settings): string {
   return (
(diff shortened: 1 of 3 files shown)


sandrinodimattia:
Thanks for this PR! Will be included in the next release

PR #115, opened 2020-05-20
Refresh token with accessTokenRequest scopes

jamesharv:
### Description

Adds support for requesting different scopes during a refresh token request, as supported by Auth0's Authentication API: https://auth0.com/docs/api/authentication#refresh-token.

This seems to be a missing feature of the library, and I need it for my project, so I thought I'd contribute. Let me know if any issues.

For example:

Given the client requested scopes `read:foo write:foo delete:foo` on login
When the client performs a refresh token request and requests only `read:foo write:foo`
Then the refresh token request should include the new scopes, so that the access_token / id_token returned only includes the new scopes as requested

It is already possible to specify a set of required scopes when getting the accessToken from the `TokenCache` (see below), so this is not an interface change. If no scopes are specified, the behaviour is the same as before.

```typescript
auth0.tokenCache(req, res).getAccessToken({ scope: ["read:foo", "write:foo"] });
```

Also note that `getAccessToken()` already enforces the constraint that you may only request a subset of the scopes requested during initial login.

### Testing

### Checklist

(Not sure this is needed, since it's not a change to the interface)

### DIFF
--- a/src/tokens/session-token-cache.ts
+++ b/src/tokens/session-token-cache.ts
@@ -75,7 +75,8 @@ export default class SessionTokenCache implements ITokenCache {
     // Adding a skew of 1 minute to compensate.
     if (session.refreshToken && session.accessTokenExpiresAt * 1000 - 60000 < Date.now()) {
       const client = await this.clientProvider();
-      const tokenSet = await client.refresh(session.refreshToken);
+      const extras = accessTokenRequest ? { exchangeBody: { scope: accessTokenRequest.scopes?.join(" ") } } : undefined;
+      const tokenSet = await client.refresh(session.refreshToken, extras);
 
       // Update the session.
       const newSession = getSessionFromTokenSet(tokenSet);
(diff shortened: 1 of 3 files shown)


jamesharv:
It's occurred to me that this may not be desirable in all situations. For example, if a user logs in and grants 4 scopes, and then the application refreshes the token at some point with only 3 of those scopes, then the user would be required to log in again if the application needed all 4 scopes at some point.

So maybe it's worth adding an option to the `AccessTokenRequest` interface which enables this behaviour. Eg.

```typescript
interface AccessTokenRequest {
  scopes?: Array<string>;
  refreshScopes?: boolean;
}
```

Happy to do more work on this if you have other suggestions.

Widcket:
Hi @jamesharv, thanks for the PR. In the recently released 1.0.0 beta, we've heavily reworked the session logic, so these changes no longer apply.

Issue #109, opened 2020-05-15
PKCE Support


reicheltp:
### Describe the problem you'd like to have solved

The OAuth 2.0 Security Best Current Practice recommends the usage of PKCE for the authorization code grant: https://tools.ietf.org/html/draft-ietf-oauth-security-topics-15#section-2.1.1

### Describe the ideal solution

If the discovery document shows that the endpoint supports PKCE, `auth0.handleLogin` should add the `code_challenge` and not use `state`.

### Additional context
- https://tools.ietf.org/html/rfc8414#section-2 at `code_challenge_methods_supported` 
- https://github.com/panva/node-openid-client#authorization-code-flow

Widcket:
Hi @reicheltp, this is now fixed in the new [v1.0.0-beta.0](https://github.com/auth0/nextjs-auth0/releases/tag/v1.0.0-beta.0) release.

reicheltp:
This is awesome 👍 I'll try out next week! 😃

balazsorban44:
@Widcket did you use a cookie to store the `code_verifier`?

Widcket:
@balazsorban44 yup, take a look at [this file](https://github.com/auth0/nextjs-auth0/blob/beta/src/auth0-session/transient-store.ts).

PR #116, opened 2020-05-20
Create Auth0 SSR Apollo Hasura example

### LINKED ISSUE #114: The code quality of the examples is low

There's IMO a number of flaws in the examples:

- It's possible to accidentally use the global user from 'fetchUser' so that you [log in others inadvertently](https://owasp.org/www-project-api-security/) (API2)
- You're printing [internal errors on failure](https://owasp.org/www-community/Improper_Error_Handling)
- You use the error handler to set the status code, when IRL the status code might already have been set by the auth0 lib; causing an exception that crashes the server because headers already have been sent
- The components' files are not properly capitalised (e.g. `Layout.jsx`) and it would be nice, had they been TypeScript, and it ought to be `withAuth.jsx` not `with-auth.jsx` (niggles only)
- E.g. `login-redirect.jsx` only redirects client-side, not server-side, potentially breaking the back button
- In e.g. `with-auth.jsx` you do a `if (!ctx.req) {` instead of the more common `if (typeof window !== 'undefined') {` — because this is basically just a check if we're running in the browser, so it should clearly be written as such
- In `with-auth` you overwrite any/all getInitialProp values that pages might have
- It's not a well written HoC; instead of returning `div > InnerComponent`, it should return a wrapper around what it received — and Layout should absolutely not require you to pass two generic properties `loading` and `user`
- You should split up `user.js` into its different parts
- NextJS now includes a `fetch` global, so you can remove the `isomorphic-unfetch` import
- Never compare to `undefined` in JS, since it's a valid variable name; so `userState === undefined` is buggy and should be `typeof userState === 'undefined'`
- The exported `UserProvider` is in fact NOT a user provider; it's some sort of wrapper around it
- You have extra `isMounted` checks when you don't need to
- There's no way to force a user refresh; once set globally in the app, it's there forever
- The pattern with `import config from './config'` is just strange and completely needless since the indirection doesn't add expressive power

Fixing all the above. I've done it for now, I can file a PR; but that includes TypeScript typings for everything; looks like this:

image

There are some not-so-nice bugs in there.

## Question

Would you accept a PR that fixes this but turns it into TypeScript?

haf:
### Description

Fixes #114 

Supercedes #113 

It should be noted that 6700 LoC is the yarn.lock file, and the deletions is the previous example that I bootstrapped this one from.

The TLS certificates are optional, requiring `cfssl` to generate and having a script to do so built in; but I've found that in dev teams and for examples, having them in leads to smoother onboarding, so they're more useful in git than not.

There are two `@ts-ignore` statements; these can be removed when you either merge #113 or this PR and release a new version; I was never able to get `yarn link` to work in a PR, nor able to get `npm` to reference the package as `../..` instead of `"^0.13.0"`. The ignores can then be removed.

I use yarn, because I don't want to keep `rm -rf node_modules`. Anecdotal evidence; the last 4 months in a dev team; team came from npm before, with yarn in this new project, removing the `node_modules` directory was never needed for the duration of the project. I prefer this consistency.

You can run/test this PR in three forms:

- `yarn && yarn dev` — simple, localhost form
- `make deploy_dev` — deploy on k8s and then you port-forward (but have to make sure your Auth0 matches this forward)
- the above + setting up an ingress, pointing to the certificates I added and having a much more realistic test environment that can include HSTS, certificate pinning, and realistic values for all configuration variables — this is how my teams do their development

Since [everything](https://hasura.io/blog/add-authentication-and-authorization-to-next-js-8-serverless-apps-using-jwt-and-graphql/) [I've](https://medium.com/@xfor/apollo-next-js-refresh-token-authentication-flow-15e5f45df5a3) [found](https://www.google.com/search?hl=en&q=hasura+apollo+ssr+authentication&oq=hasura+apollo+ssr+auth) [online](https://www.google.com/search?hl=en&q=nextjs+apollo+auth0++site%3Astackoverflow.com) [does it wrong](https://stackoverflow.com/a/58091062) this could potentially be a pretty good pattern to document. (It either only does SSR, it uses cookies, not bearer tokens, has no answers, has wrong answers, it only does client-side, or it has race conditions [like this repo])

@reicheltp Would you fix https://github.com/auth0/nextjs-auth0/issues/109 as part of this PR? I haven't investigated the oidc-client lib and what it's capable of, nor how the signing key/nonces would be saved between stateless replicas or for serverless deployments.

### Testing

I'm not sure how you handle your examples/ testing, but I still think this PR makes them better than what is available right now. Perhaps you could move this lib out of prerelease state as part of merging.

Another curious observation during the writing of this PR, was that if you link to the source directly, client-side rendering in the next dev environment doesn't work, because it loads/typechecks the node-only OIDC-client library.

### Checklist


haf:
@james-ff @bjoernrave @mrskish Now please test this PR, so we can merge it and so we can reduce the latency between changes and feedback!

examples/with-apollo-hasura-ssr/components/Layout.tsx
```diff
@@ -0,0 +1,38 @@
+import Head from 'next/head'
+import { ReactNode } from 'react'
+import Header from './Header'
+
+type Props = Readonly<{
+  loading?: boolean;
+  children: ReactNode;
+}>
+
+const Layout = ({ children }: Props) => (
```
MrShish, line 10: cool, looks much cleaner without the user attributes!

reicheltp:
@haf The PKCE Support (#109) is only an implementation detail and do neither modify the public API nor requires any changes in the examples. Thus, I think we should target this with a separate PR.

examples/with-apollo-hasura-ssr/lib/withAuth.tsx
```diff
@@ -0,0 +1,111 @@
...
+        ...pageProps,
+        user: 'user' in pageProps ? pageProps.user : await fetchUserAtClient(),
+      }
+    }
+
+    // @ts-ignore
+    const session = await auth0.getSession(req)
+
+    // Server-side, unauthenticated, just returns the location header
+    if (!session || !session.user) {
```
reicheltp, line 87: This is also true, if the session is expired. Is this the expected behavior? Otherwise, I would propose to validate the expiration date. 

```suggestion
    const isExpired = Date.now() > session?.accessTokenExpiresAt * 1000 - 60000
    if (!session || !session.user || isExpired) {
```

haf: Maybe we should move `withAuth` into the library proper as well?

haf: Also, you can put this extra if inside the parent if, since you qualify `session` as non-null in the leftmost disjunctions.

BjoernRave:
I think this example is maybe a bit too advanced. Especially bringing in kubernetes here will be overwhelming for many. 

I also don't think hasura is necessary, a simple graphql or REST api would suffice.

We might want to have one example for graphql and one for REST seperately

haf:
This PR is done anyway. Merge it then and then refactor to separate examples; it gets easier when you have the `withAuth` and the `UserProvider` built into the core, also.

m-torin:
> I also don't think hasura is necessary, a simple graphql or REST api would suffice.
I agree; core examples should be simple enough to fit generic use cases so people learn, not solve all their problems; that's what boilerplates are for. This certainly is a good complex boilerplate though.

jpwesselink:
Even though this example touches Hasura and Kubernetes I think it is very helpful. Thanks for this!

haf:
@jpwesselink You're welcome! :)

Widcket:
Hi @haf, this example is far too complex to be a core example. And with all the changes in the 1.0.0 beta, it's outdated now as well. Thanks anyway for wanting to demonstrate good practices on using this library with Apollo Client. No doubt many will still find it useful.

Issue #126, opened 2020-06-01
encodeing postLogoutRedirectUri twice 


Jamesglaschu:
# Description

> In the logout handler, where creating the endsession url it is using encodeURIComponent which causes the url for the return address to be encoded twice because client.endSessionUrl() also encodes the url, which messes up on some logout systems.

      `endSessionUrl = client.endSessionUrl({
        id_token_hint: session ? session.idToken : undefined,
        post_logout_redirect_uri: encodeURIComponent(settings.postLogoutRedirectUri)
      });`
what it should be 
      `endSessionUrl = client.endSessionUrl({
        id_token_hint: session ? session.idToken : undefined,
        post_logout_redirect_uri: settings.postLogoutRedirectUri
      });`

Jamesglaschu:
This has now been fixed

PR #127, opened 2020-06-01
fix for encodeing postLogoutRedirectUri twice

Jamesglaschu:
This pr fixes an issue in auth0.handleLogout(req, res) where the post redirect url was encoded twice causing issues logging out on identity. This still falls back to use the default Auth0 logout URL, thus existing behavior is not broken.

### References
Solves the problem described in #126 

### Testing

### Checklist


chriskolenko:
Can confirm this fixes the double encoding problem for logout.

sandrinodimattia:
Thank you 👍  It introduces an other small issue, but I'll fix it separately.

PR #133, opened 2020-06-12
Keeps previously set cookies

samuelhnrq:
### Description

Keeps any previously set cookies on the cookies helper in order to allow more cookies to be set other than the session one, e.g. on the callback route

### Testing

### Checklist


sandrinodimattia:
Thanks for this PR, looks great 👍  Will be part of the next release

PR #151, opened 2020-08-04
Allow setting insecure cookie 

bjarkebm:
### Description

Currently, the cookie set is always Secure unless running on localhost or when `NODE_ENV` is not `production`.
When deploying code to a non-https site that has `NODE_ENV=production`, setting the cookie will fail.
Our use-case is deploying PRs to a dynamically created environment, where we test that everything is working as in production (thus, we have `NODE_ENV=production`), but the environment is a random url, and so cannot have https (at least easily).

The proposed solution is to allow setting an environment variable that turns off setting the cookie as secure, `AUTH0_ALLOW_INSECURE_COOKIE=true`

### Testing

### Checklist

### DIFF
--- a/src/utils/cookies.ts
+++ b/src/utils/cookies.ts
@@ -67,6 +67,10 @@ function isSecureEnvironment(req: IncomingMessage): boolean {
     return false;
   }
 
+  if (process.env.AUTH0_ALLOW_INSECURE_COOKIE === 'true') {
+    return false;
+  }
+
   const host = (req.headers.host.indexOf(':') > -1 && req.headers.host.split(':')[0]) || req.headers.host;
   if (['localhost', '127.0.0.1'].indexOf(host) > -1) {
     return false;
--- a/src/utils/cookies.ts
+++ b/src/utils/cookies.ts
@@ -67,7 +67,7 @@ function isSecureEnvironment(req: IncomingMessage): boolean {
     return false;
   }
 
-  if (/^true$/i.test(process.env.AUTH0_ALLOW_INSECURE_COOKIE)) {
+  if (/^true$/i.test(process.env.AUTH0_ALLOW_INSECURE_COOKIE as string)) {
     return false;
   }
 
--- a/src/utils/cookies.ts
+++ b/src/utils/cookies.ts
@@ -67,7 +67,7 @@ function isSecureEnvironment(req: IncomingMessage): boolean {
     return false;
   }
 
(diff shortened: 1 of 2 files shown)


src/utils/cookies.ts
```diff
@@ -67,6 +67,10 @@ function isSecureEnvironment(req: IncomingMessage): boolean {
     return false;
   }
 
+  if (process.env.AUTH0_ALLOW_INSECURE_COOKIE === 'true') {
```
sandrinodimattia, line 70: Would it be possible to make this code slightly more resilient? Eg:

```suggestion
  if (/^true$/i.test(process.env.AUTH0_ALLOW_INSECURE_COOKIE)) {
```

tests/utils/cookies.test.ts
```diff
@@ -138,6 +138,47 @@ describe('cookies', () => {
           ]);
         });
       });
+
+      describe('when NODE_ENV is not production', () => {
+        test('should not set a secure cookie on the response', async () => {
+          const { req, res } = getRequestResponse();
+          req.headers.host = 'www.acme.com';
+
+          process.env.NODE_ENV = 'development';
```
sandrinodimattia, line 147: This line of code is changing the environment variable and will potentially have side effects for other tests. Would it be possible to include some clean up logic, like this for example:

```js
  beforeEach(() => {
    initialEnvVar = process.env. NODE_ENV
  })
  afterEach(() => {
    process.env. NODE_ENV = initialEnvVar
  })
```

sandrinodimattia (changes requested):

tests/utils/cookies.test.ts
bjarkebm: (Sorry for the late followup)
Regarding NODE_ENV, this is already handled at the outer 'cookies' level:
```javascript

const originalEnv = process.env.NODE_ENV;
// ...
describe('cookies', () => {
  // ...
  afterEach(() => {
    timekeeper.reset();
    process.env.NODE_ENV = originalEnv;
  });
```
But you are right, I do add the `AUTH0_ALLOW_INSECURE_COOKIE` to `process.env` so should clean that up.

bjarkebm:
I decided to rearrange the tests a little to make it clearer when the tests are for production environment (all tests but one), and for when they are not-production (just one test)

sandrinodimattia:
Excellent, thank you! 🙇  Will be available in the next release.

sandrinodimattia (approved):

PR #162, opened 2020-08-31
Fix #161 getAccessToken clears custom session.user attributes set via…

### LINKED ISSUE #161: getAccessToken clears custom session.user attributes set via onUserLoaded
### Description

As per the README, onUserLoaded can be used to set custom session.user attributes (eg `session.user.age`). However these attributes are removed when the accessToken is refreshed.

This can be observed here https://github.com/auth0/nextjs-auth0/blob/411896ba5044bbeda1b362f7f85b5ab8128d12cd/src/tokens/session-token-cache.ts#L83-L88 where it saves to the session store, overriding the `session.user` key.

This appears unintentional, as later on it attempts to preserve the `session.user` (however its already been overwritten via the `getAccessToken` function

https://github.com/auth0/nextjs-auth0/blob/411896ba5044bbeda1b362f7f85b5ab8128d12cd/src/handlers/profile.ts#L31-L50

### Reproduction

1. Create a callback onUserLoaded that sets a custom `session.user` attribute (such as the example in the README that sets `session.user.age`

```
await auth0.handleCallback(req, res, {
      onUserLoaded: async (req, res, session, state) => {
        return {
          ...session,
          user: {
            ...session.user,
            age: 20
          }
        };
      }
    });
```

2. Force a token refresh

```
  const tokenCache = auth0.tokenCache(request, response);
  await tokenCache.getAccessToken({
    refresh: true,
  });
```

3. Observe the users session

```
console.log(session);
```

The custom attribute will have been removed

### Environment

Version `0.16.0`

marklawlor:
### Description

Fixes #161. I left this code in `profile.ts` but it might now be redundant code (unless its used by another code path?) https://github.com/auth0/nextjs-auth0/blob/411896ba5044bbeda1b362f7f85b5ab8128d12cd/src/handlers/profile.ts#L40-L48  

@sandrinodimattia Let me know if this code should be removed

### Testing

### Checklist

### DIFF
--- a/src/tokens/session-token-cache.ts
+++ b/src/tokens/session-token-cache.ts
@@ -84,6 +84,10 @@ export default class SessionTokenCache implements ITokenCache {
       const newSession = getSessionFromTokenSet(tokenSet);
       await this.store.save(this.req, this.res, {
         ...newSession,
+        user: {
+          ...session.user,
+          ...newSession.user
+        },
         refreshToken: newSession.refreshToken || session.refreshToken
       });
 
--- a/tests/tokens/session-token-cache.test.ts
+++ b/tests/tokens/session-token-cache.test.ts
@@ -1,5 +1,6 @@
 import jose from '@panva/jose';
 
+import { IncomingMessage, ServerResponse } from 'http';
 import getRequestResponse from '../helpers/http';
 import getClient from '../../src/utils/oidc-client';
 import { withApi } from '../helpers/default-settings';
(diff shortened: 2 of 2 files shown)


Widcket:
Hi @marklawlor, thanks for the PR. However, in the new [v1.0.0-beta.0](https://github.com/auth0/nextjs-auth0/releases/tag/v1.0.0-beta.0) release we've heavily reworked the session logic, so this no longer applies. Please try it out, your feedback is welcome.

PR #171, opened 2020-09-23
Less specific tokenCacheHandler return type

Mathdrquinn:
### Description

**Update the return type of `tokenCacheHandler` to be less restrictive.**

The invocation of `tokenCache` from `tokenCacheHandler` is necessarily restricted to taking a `NextApiRequest` and an `NextApiResponse`.

It can be less specific and take an IncomingMessage and ServerResponse, which is all  the called `SessionTokenCache` requires/

In doing so the token cache handler can be utilized in Next.js's `getServerSideProps` without type errors by passing the [`req` and `res` from `context` of type `GetServerSidePropsContext`](https://github.com/vercel/next.js/blob/master/packages/next/types/index.d.ts#L111).

I encountered this as a type error when trying to refresh a token within `getServerSideProps` by calling:

```typescript
export const getServerSideProps = (context: GetServerSidePropsContext) => {
    const { req, res } = context;
    const tokenCache = auth0.tokenCache(req, res); // <- Type error here: Argument of type 'IncomingMessage' is not assignable to parameter of type 'NextApiRequest'.
    // ...
};
```

### References

None.

### Testing

None.

### Checklist

- [ ] ~I have added documentation for new/changed functionality in this PR or in auth0.com/docs~

### DIFF
--- a/src/handlers/token-cache.ts
+++ b/src/handlers/token-cache.ts
@@ -1,4 +1,4 @@
-import { NextApiRequest, NextApiResponse } from 'next';
+import { IncomingMessage, ServerResponse } from 'http';
 
 import { ISessionStore } from '../session/store';
 import { ITokenCache } from '../tokens/token-cache';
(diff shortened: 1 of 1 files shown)


Mathdrquinn:
@sandrinodimattia , is there anything I can do to raise up this work?  Did I not follow guidelines?

Mathdrquinn:
@sandrinodimattia is there another Auth0 team member I can ping to look at this work?

Widcket:
Hi @Mathdrquinn, thanks for the PR. In the recently released 1.0.0 beta, we've completely reworked the session logic, so these changes no longer apply.

Mathdrquinn:
@Widcket This work would be really helpful to those not yet ready for the migration work needed to move to v1.X and some, like myself, would wait while the project is still in beta.

Would you please reconsider?

Issue #128, opened 2020-06-03
storeAccessToken breaks login after 0.7.0


tomclark:
### Description

I'm having an issue where if I add `storeAccessToken` to the session configuration, it results in the session not being created at all in versions of the plugin later than 0.7.0. My application goes through the login sequence with Auth0, is issued a token (I can see this happening successfully in the Auth0 logs), but when Auth0 redirects back to the app, there is no session at all and I'm not logged in.

This only seems to be an issue on versions later than 0.7.0. If I use 0.7.0, it works as expected and the access token is stored in the session.

It's only the access token that causes this to fail. If I include the ID token, the session is created and I can see the ID token in the state, but adding the access token breaks login completely.

I've tried using both a regular web application and a machine-to-machine application, and I've also tried adding some of the legacy grant types:

![image](https://user-images.githubusercontent.com/784012/83683572-866f0680-a5dd-11ea-9bea-ac9ca5cb4048.png)

They're unticked in this screenshot, but I've tried it with all of them ticked as well with the same failed response.

I've tried using the code to specifically get an access token, but it fails:

```javascript
const tokenCache = await auth0.tokenCache(req, res);
const { accessToken } = await tokenCache.getAccessToken();
```

This is no doubt because - as the documentation clearly says - you must mark storeAccessToken true in the settings, but I obviously can't do that since that's what causes the problem. The tokenCache code doesn't appear to be in versions as old as 0.7.0 anyway.

I'm entirely open to the possibility (probability!) that I've not set something up correctly - be it my application in Auth0 or the config of my app or something I'm missing completely - but I'm struggling to see what the issue is, and I'd be very grateful for any assistance from the community. I can obviously continue to use 0.7.0 in the meantime, but it'd be good to upgrade to the latest version.

Thanks very much in advance for any assistance.

### Reproduction

Create a new application using `yarn create next-app --example auth0`, and then upgrade the version of @auth0/nextjs-auth0 to latest (0.13.0), and add storeAccessToken to the module that stores the Auth0 information (auth0.js):

```javascript
import config from './config'
import { initAuth0 } from '@auth0/nextjs-auth0'

export default initAuth0({
  clientId: config.AUTH0_CLIENT_ID,
  clientSecret: config.AUTH0_CLIENT_SECRET,
  scope: config.AUTH0_SCOPE,
  domain: config.AUTH0_DOMAIN,
  redirectUri: config.REDIRECT_URI,
[...]

PR #173, opened 2020-09-25
Split session cookie into multiple cookies when content is too long

mrbig:
Content of the session can easily outgrow 4090 bytes which is the limit for the size of a single cookie. This happens more easily when idToken and refreshToken are stored as well.

### Description

To overcome this issue this patch will split the session cookie into multiple cookies, and will reintegrate them upon session read.

This is done by adding a postfix to the chosen cookie name:
[cookiename].c will contain the number of cookies to look for. If missing it's assumed to be 1
[cookiename].[number] will contain part of the encrypted session.

This is proposal to fix the issue, please feel free to add any thoughts or changes.

### References

This PR fixed issue #128 

### Testing

This issue can easily be reproduced by turning on all all of `storeAccessToken`, `storeIdToken`, `storeRefreshToken`, and adding a few hundred bytes of claims to these tokens.

### Checklist

### DIFF
--- a/src/session/cookie-store/index.ts
+++ b/src/session/cookie-store/index.ts
@@ -28,7 +28,14 @@ export default class CookieSessionStore implements ISessionStore {
     const cookieCount = cookies[`${cookieName}.c`] || 1;
     let cookie = '';
     for (let i = 0; i < cookieCount; i += 1) {
-      const cookiePart = cookies[`${cookieName}.${i}`];
+      let cookiePart = cookies[`${cookieName}.${i}`];
+
+      // derecated: this could be removed later, once compatibility with legacy store is not needed
+      if (i == 0 && !cookiePart) {
+        cookiePart = cookies[`${cookieName}`];
+      }
+      // end of deprecated code
+
       if (!cookiePart || cookiePart.length === 0) {
         return null; // missing or broken cookie part
       }
--- a/src/session/cookie-store/index.ts
+++ b/src/session/cookie-store/index.ts
@@ -4,7 +4,7 @@ import { IncomingMessage, ServerResponse } from 'http';
 import { ISessionStore } from '../store';
 import Session, { ISession } from '../session';
 import CookieSessionStoreSettings from './settings';
-import { setCookie, parseCookies } from '../../utils/cookies';
+import { setCookies, parseCookies } from '../../utils/cookies';
 
 export default class CookieSessionStore implements ISessionStore {
   private settings: CookieSessionStoreSettings;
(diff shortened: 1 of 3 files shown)


src/session/cookie-store/index.ts
```diff
@@ -25,12 +25,17 @@ export default class CookieSessionStore implements ISessionStore {
     const { cookieSecret, cookieName } = this.settings;
 
     const cookies = parseCookies(req);
-    const cookie = cookies[cookieName];
-    if (!cookie || cookie.length === 0) {
-      return null;
+    const cookieCount = cookies[`${cookieName}.c`] || 1;
```
RyGuyM, line 28: Any thoughts on making this backwards compatible? So instead of defaulting to `1` it uses `const cookie = cookies[cookieName];` as the cookie when there is no cookie with the count present. That way existing sessions would be valid even after the library upgrade.

mrbig: A had this on my mind, but on the other hand it would complicate the code quite a bit. And at the end everyone only has to restart their sessions maximum once after switching to the new version.

johansundwall: @mrbig which could be unacceptable for some users.
It would need to be either backwards compatible or a breaking change.

mrbig: Ok, added some backward compatibility code and tests

src/session/cookie-store/index.ts
```diff
@@ -71,15 +83,32 @@ export default class CookieSessionStore implements ISessionStore {
...
+    do {
+      cookies.push({
+        name: `${cookieName}.${cookies.length}`,
+        value: encryptedSession.slice(start, start + 4000),
+        path: cookiePath,
+        maxAge: cookieLifetime,
+        domain: cookieDomain,
+        sameSite: cookieSameSite
+      });
+      start += 4000;
```
johansundwall, line 98: I still seem to get too large cookies with 4000.

mrbig: Maximum cookie size per [rfc](https://tools.ietf.org/html/rfc6265) is 4096 bytes including cookie name, value and attributes. So yes, with long name, or additional properties it can be exceeded. @johansundwall do you have a recommendation what we should set here?

xiaomaini: I would wait till the maintainers give the green light on the possibility of merging this PR.

Maybe something like this https://stackoverflow.com/a/57071072/1593363 would be more reliable when splitting into chunks.

mrbig: You are right, this value can be tuned any time. Since the rebuild code does not depend on the slice size, it could be changed even on a running application.

Your second remark is about the Buffer class instead of strings?

Widcket:
Hi @mrbig, really appreciate the PR here. We have added cookie stitching in the [v1.0.0-beta.0](https://github.com/auth0/nextjs-auth0/releases/tag/v1.0.0-beta.0) release. Please try it out, would love to have your feedback.

PR #187, opened 2020-11-02
Add frontend hook

Widcket:
### Description

This PR adds a frontend hook to fetch the User Profile from the React side of the app. The sample apps have been updated to use the provided hook.

### Testing

### Checklist


examples/api-call-example/components/with-auth.jsx
```diff
@@ -1,42 +1,41 @@
...
+      const result = response.ok ? await response.json() : null;
+
+      return { user: result };
     }
 
-    render() {
-      if (!this.props.user) {
-        return <RedirectToLogin />;
-      }
+    const session = await auth0.getSession(ctx.req, ctx.res);
```
Widcket, line 27: The TS type for `getSession` only takes the request, but if the response is not passed it fails at runtime.

examples/typescript-example/components/with-auth.tsx
```diff
@@ -1,48 +1,49 @@
...
-import { fetchUser } from '../lib/user';
 import createLoginUrl from '../lib/url-helper';
 import RedirectToLogin from '../components/login-redirect';
 
 type AuthenticatedProps = {
-  user?: any;
-  loading: boolean;
-  children: React.ReactChild | React.ReactChildren;
+  user?: UserProfile;
+  children: React.ReactNode;
```
adamjmcgrath, line 11: children should optional https://github.com/DefinitelyTyped/DefinitelyTyped/blob/4486794741ccb9f18f562fefade043ac31e817de/types/react/index.d.ts#L822

You could do `type AuthenticatedProps = React.PropsWithChildren & { user?: UserProfile }`

Which should resolve your issues on L32 and L45

src/hooks/use-user.tsx
```diff
@@ -0,0 +1,48 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+import React, { ReactElement, useState, useEffect, useContext, createContext } from 'react';
+
+declare const fetch: any;
```
adamjmcgrath, line 4: Can you add `lib: ['dom']` to the tsconfig, then you wont need to declare fetch https://github.com/auth0/auth0-react/blob/master/tsconfig.json#L4

adamjmcgrath:
lgtm 👍 just a couple of TS things

We can look at the failing tests later

src/hooks/use-user.tsx
adamjmcgrath: Or you may need to add something to eslint https://github.com/vercel/next.js/blob/canary/examples/with-typescript-eslint-jest/.eslintrc.json#L15

Widcket: Added `"browser": true` to eslint config. BTW, the fetch polyfill comes bundled with Next.js 9.4.0+. Currently the SDK requires 9.3.6. If we're going to keep that minimum version, we need to use `node-fetch` or similar here.

examples/typescript-example/components/with-auth.tsx
Widcket: Changed in https://github.com/auth0/nextjs-auth0/pull/187/commits/5592ca5fe476d1148b8493c7feb81289e251fd15.

src/hooks/use-user.tsx
adamjmcgrath: Let's go with 9.4 for min version

adamjmcgrath (approved):

PR #189, opened 2020-11-04
Add withAuth HOC [SDK-2120]

Widcket:
### Description

This PR adds the `withAuth` frontend HOC, that allows to fetch the user profile on server-side rendered pages.

### Testing

### Checklist


src/hooks/with-auth.tsx
```diff
@@ -0,0 +1,67 @@
...
+  render: () => React.ReactElement;
+};
+
+class RedirectToLogin extends Component<RedirectToLoginProps> {
+  public constructor(props: RedirectToLoginProps) {
+    super(props);
+  }
+
+  public componentDidMount(): void {
+    window.location.assign(createLoginUrl(Router.pathname));
```
Widcket, line 19: Since `useEffect` runs after the render, by using a class-based component with `componentDidMount` we're avoiding one extra render.

src/hooks/with-auth.tsx
```diff
@@ -0,0 +1,67 @@
...
+    if (!session || !session.user) {
+      context.res?.writeHead(302, {
+        Location: createLoginUrl(context.req.url)
+      });
+      context.res?.end();
+
+      return { user: null, children: undefined };
+    }
+
+    return { user: session.user as NullableUserProfile, children: undefined };
```
Widcket, line 63: Session uses a `Claim` type for the profile.

src/hooks/with-auth.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  Authenticated.getInitialProps = async (context: NextPageContext): Promise<AuthenticatedProps> => {
+    if (!context.req) {
+      const response = await fetch('/api/me');
+      const result = response.ok ? await response.json() : null;
+
+      return { user: result, children: undefined };
+    }
+
+    const session = await instance.getSession(context.req as NextApiRequest, context.res as NextApiResponse);
```
adamjmcgrath, line 52: `context.req` in `getInitialProps` is an `IncomingMessage`, so we need to update the types od `getSession` to accept `IncomingMessage | NextApiRequest` (and the same for `ServerResponse` and `NextApiResponse`)

Also `getSession` will error if `context.res` is undefined, so we need to figure out when that would be and handle those cases

src/hooks/with-auth.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+    return <InnerComponent {...props} user={user} />;
+  };
+
+  Authenticated.getInitialProps = async (context: NextPageContext): Promise<AuthenticatedProps> => {
+    if (!context.req) {
+      const response = await fetch('/api/me');
+      const result = response.ok ? await response.json() : null;
+
+      return { user: result, children: undefined };
```
adamjmcgrath, line 49: `children` is optional so we don't have to provide it

src/hooks/with-auth.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { NullableUserProfile, useUser } from './use-user';
+import { ISignInWithAuth0 } from '../instance';
+import { createLoginUrl } from '../utils/url-helpers';
+
+type RedirectToLoginProps = {
+  render: () => React.ReactElement;
+};
+
+class RedirectToLogin extends Component<RedirectToLoginProps> {
+  public constructor(props: RedirectToLoginProps) {
```
adamjmcgrath, line 14: you don't need to specify the constructor if it's just calling `super(args)`

src/hooks/with-auth.tsx
```diff
@@ -0,0 +1,67 @@
...
+    return this.props.render();
+  }
+}
+
+type AuthenticatedProps = React.PropsWithChildren<{ user: NullableUserProfile }>;
+
+export default function withAuth(
+  InnerComponent: React.ElementType,
+  redirect: () => React.ReactElement,
+  instance: ISignInWithAuth0
```
adamjmcgrath, line 32: I wonder if it would be better for `ISignInWithAuth0` to have a `withAuth` wrapper like the handlers (rather than having to pass the instance in) eg


```js
// profile-ssr.jsx
import { withAuth } from '../lib/auth0';
```

src/hooks/with-auth.tsx
```diff
@@ -0,0 +1,67 @@
...
+  public render(): React.ReactElement {
+    return this.props.render();
+  }
+}
+
+type AuthenticatedProps = React.PropsWithChildren<{ user: NullableUserProfile }>;
+
+export default function withAuth(
+  InnerComponent: React.ElementType,
+  redirect: () => React.ReactElement,
```
adamjmcgrath, line 31: Can we go for an options object here fro the other params? (like https://github.com/auth0/auth0-react/blob/master/src/with-authentication-required.tsx#L75)

It gives us more flexibility when designing the API

src/hooks/with-auth.tsx
Widcket: In `next` they're [also casting](https://github.com/vercel/next.js/blob/8277d4d5e305dad03fae6738e8eee07dc686b647/packages/next/next-server/server/next-server.ts#L823), so it seems safe. But it's probably better to use a union type.

Widcket: > Also getSession will error if context.res is undefined, so we need to figure out when that would be and handle those cases
I'll look into that 👍🏼

src/hooks/with-auth.tsx
Widcket: Good catch, I'll remove it. Thanks!

src/hooks/with-auth.tsx
Widcket: Yup, I'll change it.

src/hooks/with-auth.tsx
Widcket: Good idea, so the the instance would be providing the HOC (and it would already have the instance 'closed over').

src/hooks/with-auth.tsx
Widcket: But if the instance will be providing the HOC, as in :

```js
// profile-ssr.jsx
import { withAuth } from '../lib/auth0';
```

does it still make sense to use an options object? Because it will only be two params now (the inner component, and the render prop).

src/hooks/with-auth.tsx
```diff
@@ -0,0 +1,67 @@
...
+    const { user } = useUser();
+
+    if (!user) {
+      return <RedirectToLogin render={redirect} />;
+    }
+
+    return <InnerComponent {...props} user={user} />;
+  };
+
+  Authenticated.getInitialProps = async (context: NextPageContext): Promise<AuthenticatedProps> => {
```
adamjmcgrath, line 44: Also, since we're targeting 9.4+ we should probably use `getServerSideProps`

> If you're using Next.js 9.3 or newer, we recommend that you use getStaticProps or getServerSideProps instead of getInitialProps.
https://nextjs.org/docs/api-reference/data-fetching/getInitialProps

src/hooks/with-auth.tsx
```diff
@@ -0,0 +1,67 @@
...
+    if (!context.req) {
```
adamjmcgrath, line 45: I'm not sure about this either, `context.req` will be falsey when rending the static page at build time, so we shouldn't be checking the session here - let's have a chat about this, because I'm not sure how this should work at the moment

src/hooks/with-auth.tsx
Widcket: Just tried the union type: `Argument of type 'ServerResponse | undefined' is not assignable to parameter of type 'ServerResponse | NextApiResponse<any>'`. Will have to keep the cast.

PR #191, opened 2020-11-11
[SDK-2057] Express mw tests

adamjmcgrath:
Add tests for auth0-session mw

- Essentially, these tests are https://github.com/auth0/express-openid-connect/tree/master/test running against Node.js `http` primitive server rather than express.
- I also changed the name of `TransientHandler` to `TransientStore` since that seemed to make more sense.
- Added custom telemetry to the client
- removed some unreachable code


examples/typescript-example/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Typescript example for nextjs-auth0",
   "main": "index.js",
   "scripts": {
-    "dev": "NODE_OPTIONS='--inspect' next dev",
```
adamjmcgrath, line 7: I think this is what was causing the flakey cypress tests

davidpatrick (approved):

Issue #139, opened 2020-07-06
Redirect based on the session


serendipity1004:
**Please do not report security vulnerabilities here**. The [Responsible Disclosure Program](https://auth0.com/whitehat) details the procedure for disclosing security issues.

**Thank you in advance for helping us to improve this library!** Your attention to detail here is greatly appreciated and will help us respond as quickly as possible. For general support or usage questions, use the [Auth0 Community](https://community.auth0.com/) or [Auth0 Support](https://support.auth0.com/). Finally, to avoid duplicates, please search existing Issues before submitting one here.

By submitting an Issue to this repository, you agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

The problem I would like to have solved is function to be able to redirect to a different route inside `onUserLoaded` function in `handleCallback` method. `onUserLoaded` is definitely useful for doing some database transactions before redirection but currently we cannot perform a redirection to different routes depending on the session. If we wanted to perform a progressive profiling but if some of the steps are absolutely required after signup, only users who have not yet filled those information should be redirected to the form whereas users who have fulfilled the requirements should be redirected to the home page.

### Describe the ideal solution

> A clear and concise description of what you want to happen.

I think it would be most ideal if we can include `redirectTo: '/route'/` in the session return statement of `onUserLoaded` method.

## Alternatives and current work-arounds

> A clear and concise description of any alternatives you've considered or any work-arounds that are currently in place.

current work around would be to check the session every time in the client and redirect user to profiling page.

### Additional context

> Add any other context or screenshots about the feature request here.

mcky:
+1 on this, one of our integrations requires generating a token for the user during the login callback and then redirecting the user to that URL to log them in. The problem is that we won't know the URL to redirect to until after `onUserLoaded` has ran. I've locally patched the package to return extra options from `onUserLoaded`

```js
return {
  session,
  options: {
    redirectTo: claimUrl,
  },
};
```

and patched `handlers/callback`

```js
let callbackOptions;
if (options && options.onUserLoaded) {
    // session = yield options.onUserLoaded(req, res, session, decodedState);
    const { session: _session, options: _options } = yield options.onUserLoaded(req, res, session, decodedState);
    // Return the options here ^^
    session = _session
    callbackOptions = _options

}
// Create the session.
yield sessionStore.save(req, res, session);
[...]

PR #202, opened 2020-12-03
Allow function in redirectTo in callback handler

steelbrain:
Right now the API accepts a string `redirectTo` in callback. This doesn't work when you want to do conditional callbacks. Current workaround is to add another "hop" url in the mix. That hop url houses the logic for where to send the user based on their session data. Think homepage vs onboarding.

With this PR change applied, we can remove that extra redirect hop and make the redirection decision in the callback API request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -407,9 +407,6 @@ await auth0.handleCallback(req, res, {
     // "new_user" is a theoretical field here, you should replace it
     // with something else in your implementation
     return session.new_user ? '/onboarding': '/dashboard'
-  },
-  onUserLoaded: async (req, res, session, state) => {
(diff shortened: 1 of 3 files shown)


adamjmcgrath:
Hi @steelbrain - thanks for your contribution

I'd be happy to accept this if you raised it against the `beta` branch (note the callback option is called `returnTo` in the beta)

adamjmcgrath:
fixes: #139

steelbrain:
I checked out the new beta, this PR is no longer necessary.

image

We can modify the `openIdState` variable from `afterCallback`

adamjmcgrath:
Ah - good catch! thanks @steelbrain

PR #203, opened 2020-12-07
feat(profile): Allow refresh without sending json response

lukevmorris:


lukevmorris:
Apologies -- I opened this versus the wrong base repo

PR #212, opened 2020-12-22
Fix defaultValue fallback in typescript example

KATT:
### Description

- Fix fallback on defaultValue

### Testing

Try running the example with a missing `REDIRECT_URI` - will give a weird error from Auth0 that is hard to understand as it didn't fallback on the default.


Widcket:
Hi @KATT, thanks for the PR. However, in the new [v1.0.0-beta.0](https://github.com/auth0/nextjs-auth0/releases/tag/v1.0.0-beta.0) release it's no longer necessary to create an SDK instance, so this no longer applies.

PR #229, opened 2021-01-06
Added error handling to useUser [SDK-2236]

Widcket:
### Description

This PR adds an `error` property to the return object of the `useUser` hook, that will contain an `Error` instance when the `fetch()` or `response.json()` calls throw.
The tests, documentation and example apps were updated accordingly.

### Testing

### Checklist


tests/fixtures/frontend.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { UserProfile, UserProvider } from '../../src';
+import { UserProvider, UserProviderProps, UserProfile } from '../../src/frontend';
```
adamjmcgrath, line 2: Could you export `UserProviderProps` from:

https://github.com/auth0/nextjs-auth0/blob/beta/src/index.ts#L95
and
https://github.com/auth0/nextjs-auth0/blob/beta/src/index.browser.ts#L5-L12

Then the developer will be able to do `import { UserProviderProps } from '@auth0/nextjs-auth0';`

And you can leave this import as `'../../src'`

src/frontend/with-page-auth-required.tsx
```diff
@@ -66,18 +81,26 @@ export type WithPageAuthRequired = <P extends object>(
...
+      if ((user && !error) || isLoading) return;
 
       (async (): Promise<void> => {
         await router.push(`${loginUrl}?returnTo=${returnTo}`);
       })();
-    }, [user, isLoading, router, loginUrl, returnTo]);
+    }, [user, error, isLoading]);
+
+    if (error) return onError(error);
+    else if (user) return <Component {...props} />;
```
adamjmcgrath, line 101: this can be just an `if`

adamjmcgrath (changes requested):

tests/fixtures/frontend.tsx
Widcket: Changed in 6ceda646787bd2c0676ecc0a288522b2645e49de.

src/frontend/with-page-auth-required.tsx
Widcket: Changed in 6ceda646787bd2c0676ecc0a288522b2645e49de.

adamjmcgrath (approved):

PR #234, opened 2021-01-12
Make smoke tests more resilient

Widcket:
### Description

- The failing smoke tests are now retried 3 times. This helps deal with flaky tests without outright failing the CI job.
- The smoke tests no longer require a login before each test, as Cypress provides a way to [keep the cookies](https://docs.cypress.io/api/cypress-api/cookies.html). So it's now down to one login (at the beginning) plus one logout (at the end) vs 3 pairs. This reduces the chances of flaky tests, and makes the smoke tests run faster.
- The ids were replaced with `data-testid` attributes.

Also:
- Removed usage of `React.FunctionComponent` in the Kitchen Sink example, as [it's now discouraged](https://github.com/facebook/create-react-app/pull/8177) (mainly due to the implicit children).
- Removed unnecessary usage of `isLoading` in the **header** component and **shows** page of the example apps. The **profile** page [was already OK](https://github.com/auth0/nextjs-auth0/blob/beta/examples/kitchen-sink-example/pages/profile.tsx).

### Testing

### Checklist

### DIFF
--- a/cypress/integration/smoke.test.ts
+++ b/cypress/integration/smoke.test.ts
@@ -9,7 +9,7 @@ describe('smoke tests', () => {
   beforeEach(() => {
     cy.visit('/');
     cy.get('[data-testid=login]').click();
-    cy.get('input[name=username]').focus().clear().type(EMAIL);
+    cy.get('input[name=email], input[name=username]').focus().clear().type(EMAIL);
     cy.get('input[name=password]').focus().clear().type(PASSWORD);
     cy.get('button[name=submit], button[name=action]').click();
   });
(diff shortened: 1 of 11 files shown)


cypress/integration/smoke.test.ts
```diff
@@ -5,41 +5,34 @@ if (!EMAIL || !PASSWORD) {
...
-    loginToAuth0();
+  after(() => {
+    cy.get('[data-testid=logout]').click();
+  });
 
+  it('should do basic login and show user', () => {
     cy.url().should('eq', `${Cypress.config().baseUrl}/`);
-    cy.get('#profile').contains(EMAIL);
-    cy.get('#logout').click();
-    cy.get('#login').should('exist');
```
adamjmcgrath, line 23: Can we include a test for a successful logout?

Widcket: Done, thanks for pointing this out.

adamjmcgrath (approved):

PR #246, opened 2021-01-20
Added checkSession method to useUser hook [SDK-2256]

Widcket:
### Description

This PR adds a method `checkSession` to the `useUser` hook that allows checking the session status from the frontend.

https://user-images.githubusercontent.com/5055789/105118105-8bb98c80-5aac-11eb-8bfd-03341a7b7cab.mov

### Testing

### Checklist

### DIFF
--- a/src/frontend/use-user.tsx
+++ b/src/frontend/use-user.tsx
@@ -64,7 +64,25 @@ export type UserProviderProps = React.PropsWithChildren<{ user?: UserProfile; pr
 /**
  * @ignore
  */
-const User = createContext<UserContext>({} as UserContext);
+const missingUserProvider = 'You forgot to wrap your app in <UserProvider>';
+
+/**
+ * @ignore
+ */
+const User = createContext<UserContext>({
+  get user(): never {
+    throw new Error(missingUserProvider);
+  },
+  get error(): never {
+    throw new Error(missingUserProvider);
+  },
+  get isLoading(): never {
+    throw new Error(missingUserProvider);
+  },
+  checkSession: (): never => {
+    throw new Error(missingUserProvider);
+  }
+});
 
 /**
  * The `useUser` hook, which will get you the {@link UserProfile} object from the server side session by requesting it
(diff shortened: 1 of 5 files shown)


src/frontend/use-user.tsx
```diff
@@ -109,21 +110,22 @@ export default ({
...
-      } finally {
-        setIsLoading(false);
-      }
-    })();
+      });
+  }, [profileUrl]);
+
+  useEffect((): void => {
+    if (user) return;
+    checkSession().finally(() => setIsLoading(false));
```
adamjmcgrath, line 127: Let's stick with using async/await everywhere so

```js
  const checkSession = useCallback(async (): Promise<void> => {
    try {
      const response = await fetch(profileUrl);
      setUser(response.ok ? await response.json() : undefined);
      setError(undefined);
    } catch (_e) {
      setUser(undefined);
      setError(new Error(`The request to ${profileUrl} failed`));
    }
  }, [profileUrl]);

  useEffect((): void => {
    if (user) return;
    (async (): Promise<void> => {
      await checkSession();
      setIsLoading(false);
    })();
  }, [user]);
```

It's easier to debug and passing promises into promises like `promise.then(async () => {})` is generally considered an anti pattern

src/frontend/use-user.tsx
```diff
@@ -63,7 +64,7 @@ export type UserProviderProps = React.PropsWithChildren<{ user?: UserProfile; pr
 /**
  * @ignore
  */
-const User = createContext<UserContext>({ isLoading: false });
+const User = createContext<UserContext>({} as UserContext);
```
adamjmcgrath, line 67: We can use the initial context to warn the user when they forget to include the `Provider`, eg

```js
const User = createContext<UserContext>({
  isLoading: false,
  checkSession: () => Promise.reject('You forgot to wrap your app in <UserProvider>')
});
```

We do the same in react https://github.com/auth0/auth0-react/blob/master/src/auth0-context.tsx#L141-L156

adamjmcgrath (changes requested):

tests/frontend/use-user.test.tsx
```diff
@@ -64,7 +66,7 @@ describe('context wrapper', () => {
   });
 
   test('should use a custom profileUrl', async () => {
-    const fetchSpy = jest.fn();
+    const fetchSpy = jest.fn().mockReturnValue({ then: () => Promise.resolve() });
```
adamjmcgrath, line 69: ` const fetchSpy = jest.fn().mockResolvedValue({})`

I would always use native promises rather than try and create mocked thenables

adamjmcgrath (approved):

PR #252, opened 2021-01-21
propose change to nextjs ^10.0.5 on peerDependencies

cebrakadabra:
Hey all,
I was testing the nextjs-auth0 package on nextjs ^10.0.5 and could verify it's implementation.
I also saw, that somebody is working already in a branch (beta) that has a >=10 range. 

Depending on the release of the beta, this could be moved in already based on the tests I did as there were no breaking changes between nextjs9 and nextjs10.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "openid-client": "^3.15.0"
   },
   "peerDependencies": {
-    "next": "^9.0.0"
+    "next": "^10.0.5"
   },
   "jest": {
     "testEnvironment": "node",


Widcket:
Hi @cebrakadabra, thanks for the PR. However, we cannot drop support for Next.js 9, as that would be a breaking change on this library. And the first beta release is out already, feel free to [check it out](https://github.com/auth0/nextjs-auth0/releases/tag/v1.0.0-beta.0).

PR #253, opened 2021-01-24
Simplified configuration of custom URLs [SDK-2264]

Widcket:
### Description

If the developer set up their login handler URL to be something other than the default `/api/auth/login`, then they would have to pass the custom value on every call to `withPageAuthRequired`. Same if they wanted to customize the URL the user would be redirected to after login (the `returnTo` URL).

This PR moves the configuration of custom URLs out of `withPageAuthRequired` to the `UserProvider` on the client-side and the config object on the server-side. Plus it's now also possible to set them via environment variables that are accessible both on the server and the client, enabling the developer to just set them once in a single place. This results in an improved developer experience.

Also, it's possible to customize the profile handler URL via environment variable as well, saving the developer from having to pass it to the `UserProvider`.

### Testing

### Checklist

- [X] I have added documentation for new/changed functionality in this PR or in auth0.com/docs

### DIFF
--- a/src/auth0-session/get-config.ts
+++ b/src/auth0-session/get-config.ts
@@ -164,7 +164,7 @@ export const get = (params: ConfigParameters = {}): Config => {
     ...params
   };
 
-  const { value, error, warning } = paramsSchema.validate(config);
+  const { value, error, warning } = paramsSchema.validate(config, { allowUnknown: true });
   if (error) {
     throw new TypeError(error.details[0].message);
   }
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,6 +1,6 @@
-import { IncomingMessage } from 'http';
 import { AuthorizationParameters as OidcAuthorizationParameters } from 'openid-client';
-import { LoginOptions, DeepPartial } from './auth0-session';
+
+import { DeepPartial, Config as SessionLayerConfig } from './auth0-session';
 
 /**
  * ## Configuration properties.
@@ -39,6 +39,8 @@ import { LoginOptions, DeepPartial } from './auth0-session';
  * - `AUTH0_IDP_LOGOUT`: See {@link idpLogout}
  * - `AUTH0_ID_TOKEN_SIGNING_ALG`: See {@link idTokenSigningAlg}
  * - `AUTH0_LEGACY_SAME_SITE_COOKIE`: See {@link legacySameSiteCookie}
+ * - `NEXT_PUBLIC_AUTH0_LOGIN`: See {@link Config.routes}
+ * - `NEXT_PUBLIC_AUTH0_POST_LOGIN_REDIRECT`: See {@link Config.routes}
  * - `AUTH0_POST_LOGOUT_REDIRECT`: See {@link Config.routes}
  * - `AUTH0_CALLBACK`: See {@link Config.routes}
  * - `AUTH0_AUDIENCE`: See {@link Config.authorizationParams}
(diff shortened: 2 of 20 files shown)


src/auth0-session/get-config.ts
```diff
@@ -164,7 +164,7 @@ export const get = (params: ConfigParameters = {}): Config => {
     ...params
   };
 
-  const { value, error, warning } = paramsSchema.validate(config);
+  const { value, error, warning } = paramsSchema.validate(config, { allowUnknown: true });
```
Widcket, line 167: Otherwise the validation would fail because of the new extra fields.

src/config.ts
```diff
@@ -39,6 +39,8 @@ import { LoginOptions, DeepPartial } from './auth0-session';
  * - `AUTH0_IDP_LOGOUT`: See {@link idpLogout}
  * - `AUTH0_ID_TOKEN_SIGNING_ALG`: See {@link idTokenSigningAlg}
  * - `AUTH0_LEGACY_SAME_SITE_COOKIE`: See {@link legacySameSiteCookie}
+ * - `NEXT_PUBLIC_AUTH0_LOGIN`: See {@link Config.routes}
+ * - `NEXT_PUBLIC_AUTH0_POST_LOGIN_REDIRECT`: See {@link Config.routes}
```
Widcket, line 43: Not adding `NEXT_PUBLIC_AUTH0_PROFILE ` here as it's not used server-side.

src/config.ts
```diff
@@ -82,180 +84,35 @@ import { LoginOptions, DeepPartial } from './auth0-session';
...
+     * Relative path to the login handler.
      */
-    login: string | false;
+    login: string;
+
+    /**
+     * Either a relative path to the application or a valid URI to an external domain.
+     * The user will be redirected to this after a login has been performed.
+     */
+    postLoginRedirect: string;
```
adamjmcgrath, line 101: If we're going to create a `postLoginRedirect` option, then it should work the same as `postLogoutRedirect` - in that it is the default in `auth0-session#logout` if `returnTo` is not set https://github.com/auth0/nextjs-auth0/blob/beta/src/auth0-session/handlers/logout.ts#L17

adamjmcgrath (changes requested):

src/frontend/use-config.tsx
```diff
@@ -0,0 +1,22 @@
...
+  returnTo?: string;
+};
+
+const Config = createContext<ConfigContext>({});
+
+export type ConfigProviderProps = React.PropsWithChildren<ConfigContext>;
+export type UseConfig = () => ConfigContext;
+export const useConfig: UseConfig = () => useContext<ConfigContext>(Config);
+
+export default ({
```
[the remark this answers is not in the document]
Widcket: The `useConfig` hook is not public, is internal to the SDK.

src/frontend/with-page-auth-required.tsx
```diff
@@ -81,12 +63,8 @@ export type WithPageAuthRequired = <P extends object>(
...
   return function withPageAuthRequired(props): JSX.Element {
     const router = useRouter();
-    const {
-      returnTo = router.asPath,
-      onRedirecting = defaultOnRedirecting,
-      onError = defaultOnError,
-      loginUrl = '/api/auth/login'
-    } = options;
+    const { onRedirecting = defaultOnRedirecting, onError = defaultOnError } = options;
+    const { loginUrl, returnTo } = useConfig();
```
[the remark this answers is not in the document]
Widcket: Good point, I'll re-add the config option so it's customizable by route. But still should be able to be centralizable for the majority of users that won't need route-specific overrides.

src/auth0-session/get-config.ts
Widcket: The issue with that approach is that you can't really have a configuration object that is any different from the one the underlying session library uses. That's not flexible and is not really an option, as it's a hard coupling with that package. I don't think that being able to catch misspellings is worth trading off for that flexibility.

src/config.ts
Widcket: Good catch, I'll add it.

src/index.ts
```diff
@@ -53,15 +53,15 @@ function getInstance(): SignInWithAuth0 {
 }
 
 export const initAuth0: InitAuth0 = (params) => {
-  const config = getConfig(getParams(params));
+  const config = getConfig(getParams(params)) as Config;
```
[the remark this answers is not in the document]
Widcket: This is related to coupling this config object to the session package's one. As we're extending that config object, that should not be an issue because it still contains all the properties from the session package config.

src/helpers/with-page-auth-required.ts
```diff
@@ -90,22 +92,25 @@ export type WithPageAuthRequired = {
 /**
  * @ignore
  */
-export default function withPageAuthRequiredFactory(sessionCache: SessionCache): WithPageAuthRequired {
+export default function withPageAuthRequiredFactory(config: Config, sessionCache: SessionCache): WithPageAuthRequired {
```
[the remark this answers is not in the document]
Widcket: I'll use the environment variables directly then.

src/frontend/use-config.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { ReactElement, useContext, createContext } from 'react';
```
adamjmcgrath, line 1: Can you exclude this file from the API Docs https://github.com/auth0/nextjs-auth0/blob/beta/typedoc.js#L4

Widcket: Sure 👍🏼

src/config.ts
```diff
@@ -82,180 +84,35 @@ import { LoginOptions, DeepPartial } from './auth0-session';
...
-  auth0Logout: boolean;
-
-  /**
-   *  URL parameters used when redirecting users to the authorization server to log in.
-   *
-   *  If this property is not provided by your application, its default values will be:
-   *
-   * ```js
-   * {
-   *   response_type: 'code',
```
adamjmcgrath, line 111: We'll need to keep these docs for most of these properties because the defaults are different.

If you generate the docs now, it'll say the default `response_type` is `id_token` when it should be `code` https://github.com/auth0/nextjs-auth0/blob/beta/src/auth0-session/config.ts#L40

I also wanted to add the corresponding env vars to each config option, eg

```js
    /**	
   * To opt-out of sending the library and node version to your authorization server	
   * via the `Auth0-Client` header. Default is `true	
   * You can also use the `AUTH0_IDP_LOGOUT` env var
   */	
  enableTelemetry: boolean;
```

Also `baseURL` has some Vercel specific docs

src/auth0-session/get-config.ts
adamjmcgrath: I'll come up with a solution for this and https://github.com/auth0/nextjs-auth0/pull/253#discussion_r563631152 and share it with you

src/frontend/with-page-auth-required.tsx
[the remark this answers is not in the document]
adamjmcgrath: I'm going to insist that we don't have a centralised `returnTo` option, for these reasons:

1. We don't have one in auth0-react's [withAuthenticationRequired](https://github.com/auth0/auth0-react/blob/master/src/with-authentication-required.tsx#L73), auth0-angular's [AuthGuard](https://github.com/auth0/auth0-angular/blob/master/projects/auth0-angular/src/lib/auth.guard.ts#L45) or express-openid-connect's [requiresAuth](https://github.com/auth0/express-openid-connect/blob/master/middleware/requiresAuth.js#L11) - and AFAIK, we've never had a request for it
2. It wouldn't be obvious from the name what a global `returnTo` option would be for (we accept a `returnTo` option in `handleLogin`, `handleLogout` and `withPageAuthRequired`)
3. `withPageAuthRequired` has a default behaviour - returning to the url the user was trying to visit, adding a global option would disable this and be a source of confusion for developers
4. We don't have a good use case for it
5. It's easy to add this option if user's ask for it, but difficult to take away without a breaking change

src/helpers/with-page-auth-required.ts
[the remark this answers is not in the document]
adamjmcgrath: Using environment variables directly won't work for the case where user's create their own instance of the SDK using `initAuth0({ routes: { loginUrl: ... }})`

src/auth0-session/get-config.ts
adamjmcgrath: I've put a branch here https://github.com/auth0/nextjs-auth0/tree/split-config

It should make it easier for you to add config that isn't used by the base `auth0-session` layer https://github.com/auth0/nextjs-auth0/blob/split-config/src/config.ts#L460 - I'll add a few more tests to the config module and raise a PR

Widcket:
Superseded by https://github.com/auth0/nextjs-auth0/pull/262, https://github.com/auth0/nextjs-auth0/pull/263 and https://github.com/auth0/nextjs-auth0/pull/268.

Issue #154, opened 2020-08-09
Document best practices for using nextjs-auth0 with a nextjs production build.


martaver:
### Description

During a production build, `nextjs` creates optimised builds for each page. Naturally, these pages won't authenticate as there is no user, however auth0 will require configuration variables, such as `AUTH0_DOMAIN`, `AUTH0_CLIENT` and so on...

These are secrets that should be injected into the application at runtime, and it strikes me that they shouldn't be necessary at build time and that would be bad practice.

Is there an official take on this?

invasionofsmallcubes:
You know, I actually had the same doubt when I've read the tutorial. Seems weird write?
I was thinking to try a build passing the variables from environment but it's not clear if it will work at runtime on Vercel.

jordie23:
It's because in the example the `initAuth0` function is executed as soon as the file is imported.

i.e. this is generally used in the examples:

```javascript
import { initAuth0 } from '@auth0/nextjs-auth0';

export default initAuth0({
  domain: '<AUTH0_DOMAIN>'
  clientId: '<AUTH0_CLIENT_ID>',
  clientSecret: '<AUTH0_CLIENT_SECRET>',
  audience: 'https://api.mycompany.com/',
  scope: 'openid profile',
  redirectUri: 'http://localhost:3000/api/callback',
  postLogoutRedirectUri: 'http://localhost:3000/',
  session: {
    cookieSecret: '<RANDOMLY_GENERATED_SECRET>',
    cookieLifetime: 60 * 60 * 8,
    cookieDomain: 'https://mycompany.com',
    storeAccessToken: true
  }
});
```

I changed it to:

```javascript
import { initAuth0 } from '@auth0/nextjs-auth0'

let auth0 = null

export default () => {
  if (!auth0) {
    auth0 = initAuth0({
      domain: '<AUTH0_DOMAIN>'
      clientId: '<AUTH0_CLIENT_ID>',
      clientSecret: '<AUTH0_CLIENT_SECRET>',
      audience: 'https://api.mycompany.com/',
      scope: 'openid profile',
      redirectUri: 'http://localhost:3000/api/callback',
      postLogoutRedirectUri: 'http://localhost:3000/',
      session: {
        cookieSecret: '<RANDOMLY_GENERATED_SECRET>',
        cookieLifetime: 60 * 60 * 8,
        cookieDomain: 'https://mycompany.com',
        storeAccessToken: true
      }
    });
  }
  return auth0
}
```

[...]

PR #255, opened 2021-01-25
[SDK-2259] Remove env var dependency at build time

adamjmcgrath:
### Description

Because of [static generation](https://nextjs.org/docs/basic-features/pages#static-generation-recommended) `withPageAuthRequired` is called at build time. We need to move the dependency `withPageAuthRequired` has on the server instance to be at runtime, so that you don't need to specify environment variables to configure the SDK at build time

### References

fixes: #154 

### Testing

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,6 +129,7 @@
     "collectCoverageFrom": [
       "src/**/*.*",
       "!src/index.browser.ts",
+      "!src/index.ts",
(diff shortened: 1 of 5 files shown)


Widcket (approved):

StianOvrevage:
Is there a chance of getting this merged into main branch?

PR #263, opened 2021-01-29
Split configuration into Next and Base config

adamjmcgrath:
### Description

- separate the config into NextConfig and BaseConfig (I'll come up with a better name when I figure out a better name for `auth0-session`) to make it easier to supply configuration that applies to the Next layer only (eg custom routes)
- Removed a bunch of docs and options that were only specific to the express SDK and got inherited by `auth0-session`

### Testing

### Checklist


src/config.ts
```diff
@@ -358,10 +276,91 @@ export interface AuthorizationParameters extends OidcAuthorizationParameters {
 }
 
 /**
- * See {@link Config}
+ * @ignore
+ */
+export type NextConfig = Pick<BaseConfig, 'routes' | 'identityClaimFilter'>;
```
adamjmcgrath, line 281: Too add custom routes

```ts
export interface NextConfig extends Pick<BaseConfig, 'identityClaimFilter'> {
  routes: {
    postLogoutRedirect: string;
    callback: string;
    login: string;
    logout: string;
  };
}
```

src/auth0-session/get-config.ts
```diff
@@ -133,8 +133,6 @@ const paramsSchema = Joi.object({
   issuerBaseURL: Joi.string().uri().required(),
   legacySameSiteCookie: Joi.boolean().optional().default(true),
   routes: Joi.object({
-    login: Joi.alternatives([Joi.string().uri({ relativeOnly: true }), Joi.boolean().valid(false)]).default('/login'),
-    logout: Joi.alternatives([Joi.string().uri({ relativeOnly: true }), Joi.boolean().valid(false)]).default('/logout'),
     callback: Joi.string().uri({ relativeOnly: true }).default('/callback'),
```
Widcket, line 136: Why have a default callback URL?

Widcket (changes requested):
Looks good, just have a question.

src/auth0-session/get-config.ts
adamjmcgrath: Good point 👍 - that was left over from the express code. Will remove that and make the field required.

src/auth0-session/get-config.ts
```diff
@@ -133,9 +133,7 @@ const paramsSchema = Joi.object({
   issuerBaseURL: Joi.string().uri().required(),
   legacySameSiteCookie: Joi.boolean().optional().default(true),
   routes: Joi.object({
-    login: Joi.alternatives([Joi.string().uri({ relativeOnly: true }), Joi.boolean().valid(false)]).default('/login'),
-    logout: Joi.alternatives([Joi.string().uri({ relativeOnly: true }), Joi.boolean().valid(false)]).default('/logout'),
-    callback: Joi.string().uri({ relativeOnly: true }).default('/callback'),
+    callback: Joi.string().uri({ relativeOnly: true }).required(),
```
Widcket, line 136: The login and logout URLs don't need to be required?

adamjmcgrath: I've removed the login and logout URLs from the config as they are not used by `auth0-session`

Widcket (approved):

PR #268, opened 2021-02-02
Simplified configuration of custom login URL [SDK-2264]

Widcket:
### Description

This PR makes it possible to set a custom login URL for the server-side version of `withPageAuthRequired` with an environment variable. This way, the developer only has to configure it once, instead of in every call site of `withPageAuthRequired`.

Also, the server-side version of `withPageAuthRequired` wasn't taking the `returnTo` parameter that the client-side takes. This was corrected and both versions offer the same public-facing configuration options.

### References

Supercedes https://github.com/auth0/nextjs-auth0/pull/253

### Testing

### Checklist

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -47,9 +47,10 @@ import {
 } from './helpers';
 import { InitAuth0, SignInWithAuth0 } from './instance';
 import version from './version';
-import { getConfig, ConfigParameters } from './config';
+import { getConfig, getEmptyNextConfig, ConfigParameters } from './config';
 
 let instance: SignInWithAuth0;
+let config = getEmptyNextConfig();
 
 function getInstance(): SignInWithAuth0 {
   if (instance) {
@@ -61,6 +62,7 @@ function getInstance(): SignInWithAuth0 {
 
 export const initAuth0: InitAuth0 = (params) => {
   const { baseConfig, nextConfig } = getConfig(params);
+  config = nextConfig;
 
   // Init base layer (with base config)
   const getClient = clientFactory(baseConfig, { name: 'nextjs-auth0', version });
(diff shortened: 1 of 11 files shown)


src/index.ts
```diff
@@ -99,7 +101,7 @@ export const getSession: GetSession = (...args) => getInstance().getSession(...a
 export const getAccessToken: GetAccessToken = (...args) => getInstance().getAccessToken(...args);
 export const withApiAuthRequired: WithApiAuthRequired = (...args) => getInstance().withApiAuthRequired(...args);
 export const withPageAuthRequired: WithPageAuthRequired = (...args: any[]): any =>
-  withPageAuthRequiredFactory(getSession)(...args);
+  withPageAuthRequiredFactory(config, getSession)(...args);
```
adamjmcgrath, line 104: The empty config you're passing here won't be populated by the time it's used because you're [querying the config](https://github.com/auth0/nextjs-auth0/blob/5a909dac5a3bc0a94120a5220c59322564452078/src/helpers/with-page-auth-required.ts#L103) before `getSession` (which calls `initAuth0`) [is called](https://github.com/auth0/nextjs-auth0/blob/5a909dac5a3bc0a94120a5220c59322564452078/src/helpers/with-page-auth-required.ts#L106)

You also can't reassign the config, you'd need to mutate it `Object.assign(config, nextConfig)` for it to work.

Either way, I think it's too brittle to initialise an empty config then expect it to be populated when you want to use it.

For this special case, the easiest thing to do is probably just pass the `loginUrl`

```js
export const withPageAuthRequired: WithPageAuthRequired = (...args: any[]): any =>
  withPageAuthRequiredFactory({ loginUrl: process.env. NEXT_PUBLIC_AUTH0_LOGIN || '/api/auth/login'}, getSession)(...args);
```

You can DRY this up by creating a `getLoginUrlFromEnv` type function in `config.ts` if you don't want to do the `process.env. NEXT_PUBLIC_AUTH0_LOGIN || '/api/auth/login'` twice

Or if you want the signature for `withPageAuthRequiredFactory` to still use the whole `NextConfig` object, you could create a `getNextConfigFromEnv` type function in `config.ts`.

Let me know if you want me to go through this with you

tests/index.test.ts
```diff
@@ -5,7 +5,7 @@ describe('index', () => {
     const secret = process.env.AUTH0_SECRET;
     delete process.env.AUTH0_SECRET;
     expect(() => withApiAuthRequired(jest.fn())).toThrow('"secret" is required');
-    expect(() => withPageAuthRequired()).not.toThrow();
+    expect(() => withPageAuthRequired(jest.fn())).not.toThrow();
```
adamjmcgrath, line 8: This is testing the server side `withPageAuthRequired`, so shouldn't take a function as it's first argument.

adamjmcgrath (changes requested):

adamjmcgrath:
Looks good 👍 Just need to figure out a different solution to that special case with the named `withPageAuthRequired` export

src/index.ts
Widcket: Thanks changed in https://github.com/auth0/nextjs-auth0/pull/268/commits/2b4b4c6e8d9dc23051e22ba777b22aa9a7bc8cda.

tests/index.test.ts
Widcket: Changed in https://github.com/auth0/nextjs-auth0/pull/268/commits/2b4b4c6e8d9dc23051e22ba777b22aa9a7bc8cda.

adamjmcgrath (approved):

Issue #296, opened 2021-02-17
Set-Cookie exceeds CHUNK_BYTE_SIZE=4000 


indiejoseph:
### Description

> We've just upgraded our project to v1.0.0, and it working fine on local environment with env variables:
```
AUTH0_SESSION_NAME=ec
AUTH0_COOKIE_DOMAIN=localhost
AUTH0_COOKIE_TRANSIENT=false
AUTH0_COOKIE_HTTP_ONLY=true
AUTH0_COOKIE_SECURE=false
AUTH0_COOKIE_SAME_SITE=lax
```
However, when the project deployed on Vercel with production environment variables, the callback not able to Set-Cookie, there show Invalid cookie on Chrome browser, and I counted the length of Set-Cookie was exceeded `CHUNK_BYTE_SIZE=4000 `:

```
[...]

PR #301, opened 2021-02-18
Chunked cookies should not exceed browser max

adamjmcgrath:
### Description

We currently allow 96 characters leeway for cookie attribute size ([browser cookie max size (4096 Bytes)](https://tools.ietf.org/html/rfc6265#section-6.1) minus [CHUNK_BYTE_SIZE](https://github.com/auth0/nextjs-auth0/blob/v1.0.0/src/auth0-session/cookie-store.ts#L12))

This is not enough for scenarios where the user provides many cookie options or longer cookie domain or path options.

Since the cookie attributes are dynamic and can be quite long, we can calculate the cookie attributes length from serializing an empty cookie with the same options and measuring it.

### References

fixes: #296 

### Testing

- Set a bunch of really long claims in your session (> 2000 chars) using the [afterCallback](https://auth0.github.io/nextjs-auth0/modules/handlers_callback.html#aftercallback) hook
- Specify a bunch of cookie options that will exceed 96 chars (eg a long cookie domain/path)
- Login (you'll need to login on the same domain/path) and check the `Size` of your chunked `appSession.{n}` cookies in Chrome Devtools > Application > Cookie
- They should not exceed 4095 (Chrome Devtools doesn't inlclude the `=`)

### Checklist

### DIFF
--- a/src/auth0-session/cookie-store.ts
+++ b/src/auth0-session/cookie-store.ts
@@ -20,6 +20,8 @@ export default class CookieStore {
 
   private currentKey: JWK.OctKey | undefined;
 
+  private chunkSize: number;
+
   constructor(public config: Config) {
     const secrets = Array.isArray(config.secret) ? config.secret : [config.secret];
     this.keystore = new JWKS.KeyStore();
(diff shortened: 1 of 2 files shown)


panva (approved):
LGTM, the same should be applied to express-openid-connect as well i assume.

Issue #294, opened 2021-02-17
Give handleProfile an option to refresh the access token.


marklawlor:
### Describe the problem

When using the `refetch` option with `handleProfile`, the `getAccessToken` function is called but without the `refresh` option.

It would be good for the handleProfile function to also refresh the user session, as this would work nicely with the `useUser` hook, allowing returning users to automatically get a refreshed session.

This can also be achieved by calling `getAccessToken` before `handleProfile`

```
await auth0.getAccessToken(request, response, { refresh: true });
return auth0.handleProfile(request, response, { refetch: true });
```

adamjmcgrath:
Hi @marklawlor 

> When using the refetch option with handleProfile, the getAccessToken function is called but without the refresh option.

This is by design. You'd only need to use the `refresh` option for `getAccessToken` if you were going to use the access token at some point in the future and you wanted to make sure that you had the full access token expiry time available on the access token.

`handleProfile` uses the access token immediately so does not need to use the `refresh` option.

The original issue you've described looks valid though:

> When enabling Refresh Token Rotation a session re-fetch will only work once.
> getAccessToken does initially update the sessionCache with the rotated secret.
> But then the profile handler removes it (it is using a stale session)
> This is why the refetch works once, but fails to work once the token as been rotated.

Which I'll take a look at

marklawlor:
@adamjmcgrath Ok makes sense. Yes the original issue is what I was experiencing, but as I was skimming the source code I got muddled on how to report it.

Do you mind just clarifying my understanding around `handleProfile` and the `useUser` hook?

I assumed that if the user revisits my website with an expired access token, but valid refresh token then the `useUser` hook would handle the refresh (when it calls the `/api/auth/me` endpoint)

But the design is actually, the `useUser` hook will return an error and I should redirect my user to the `/api/auth/login` endpoint?

If I wished for a seamless experience then the work around is to do something like this?

```
auth0.handleAuth({
  async profile(request, response) {
    await auth0.getAccessToken(request, response, { refresh: true });
    return auth0.handleProfile(request, response);
  },
});
```

adamjmcgrath:
> I assumed that if the user revisits my website with an expired access token, but valid refresh token then the useUser hook would handle the refresh (when it calls the /api/auth/me endpoint)

Yes, this is the expected behaviour - you shouldn't need to do `await auth0.getAccessToken(request, response, { refresh: true });` to get this to work. 

[...]

PR #307, opened 2021-02-18
[SDK-2330] New tokens should be applied to existing session

adamjmcgrath:
### Description

`handleProfile` will use the a stale session if `getAccessToken` updates the session.

On the whole a user should be able to continue to use a refrence to the session after `getAccessToken` is called, without having to call `getSession` again.

```js
const session = getSession(req, res);
const { accessToken } = await getAccessToken(req, res);

// If AT was updated, session.accessToken should be updated too
session.accessToken === accessToken;
```

### References

See https://github.com/auth0/nextjs-auth0/issues/294#issuecomment-781376276

### Testing

Given I have Refresh Token Rotation enabled
And I have an expired Access Token
When I visit `handleProfile` with refetch `true`
Then I should get an updated profile
And I should get an updated Access Token and Refresh Token

### Checklist

### DIFF
--- a/src/session/get-access-token.ts
+++ b/src/session/get-access-token.ts
@@ -3,7 +3,7 @@ import { NextApiRequest, NextApiResponse } from 'next';
 import { ClientFactory } from '../auth0-session';
 import { AccessTokenError } from '../utils/errors';
 import { intersect, match } from '../utils/array';
-import { SessionCache, fromTokenSet, fromJson } from '../session';
+import { SessionCache, fromTokenSet } from '../session';
 import { NextConfig } from '../config';
 
 /**
@@ -114,16 +114,11 @@ export default function accessTokenFactory(
 
       // Update the session.
       const newSession = fromTokenSet(tokenSet, config);
-      sessionCache.set(
-        req,
-        res,
-        fromJson({
-          ...session,
-          ...newSession,
-          refreshToken: newSession.refreshToken || session.refreshToken,
-          user: { ...session.user, ...newSession.user }
-        })
-      );
+      Object.assign(session, {
+        ...newSession,
+        refreshToken: newSession.refreshToken || session.refreshToken,
+        user: { ...session.user, ...newSession.user }
+      });
 
       // Return the new access token.
       return {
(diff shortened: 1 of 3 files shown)


src/session/get-access-token.ts
```diff
@@ -114,16 +114,11 @@ export default function accessTokenFactory(
...
-          ...newSession,
-          refreshToken: newSession.refreshToken || session.refreshToken,
-          user: { ...session.user, ...newSession.user }
-        })
-      );
+      Object.assign(session, {
+        ...newSession,
+        refreshToken: newSession.refreshToken || session.refreshToken,
+        user: { ...session.user, ...newSession.user }
+      });
```
Widcket, line 121: Does this get saved in the session cache?

tests/handlers/profile.test.ts
```diff
@@ -91,6 +91,28 @@ describe('profile handler', () => {
...
+      callbackOptions: {
+        afterCallback
+      },
+      userInfoToken: 'new-access-token'
+    });
+    refreshTokenRotationExchange(withApi, 'GEbRxBN...edjnXbL', {}, 'new-access-token', 'new-refresh-token');
+    const cookieJar = await login(baseUrl);
+    const profile = await get(baseUrl, '/api/auth/me', { cookieJar });
+    expect(profile).toMatchObject({ foo: 'bar' });
+    const session = await get(baseUrl, `/api/session`, { cookieJar });
```
Widcket, line 111:
```suggestion
    const session = await get(baseUrl, '/api/session', { cookieJar });
```

Widcket (changes requested):
Looks good, just have a question.

src/session/get-access-token.ts
adamjmcgrath: Yep - the session is mutable and whatever is in it at the end of the request chain [will get persisted ](https://github.com/auth0/nextjs-auth0/blob/main/src/session/cache.ts#L22)

adamjmcgrath: The second request to `/api/session` in the test should also verify this https://github.com/auth0/nextjs-auth0/pull/307/files#diff-43549841cd306e45cf537adad769f7fc683abaf46a9eab1146d958238493af2aR112-R113

Widcket (approved):

PR #336, opened 2021-03-12
chore(deps): move from @hapi/joi to joi

veksen:
### Description

The `@hapi/joi` package has moved to `joi`, and the TS definitions are now included with the package.

### Testing

N/A, but I ran the test suite locally.

### Checklist

Note: I got a warning from using package-lock v1, while the project uses v2. This would explain the huge change of `package-lock.json`. Not sure how to change that, I'm using Node 14 locally.

(by the way, the https://github.com/auth0/open-source-template/blob/main/CONTRIBUTING.md link is a 404)

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,6 @@
     "@types/clone": "^2.1.0",
     "@types/cookie": "^0.4.0",
     "@types/debug": "^4.1.5",
-    "@types/hapi__joi": "^17.1.6",
     "@types/http-errors": "^1.8.0",
     "@types/jest": "^26.0.20",
     "@types/jsonwebtoken": "^8.5.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -105,7 +104,6 @@
     "typescript": "^4.1.3"
   },
   "dependencies": {
-    "@hapi/joi": "^17.1.1",
```
adamjmcgrath, line 108: Thanks @veksen - could you add `joi` and make sure the project still compiles? We also still need the types, unless the new version of `joi` comes bundled with them?

adamjmcgrath (changes requested):

package.json
veksen: The important part 😅 Added

veksen: The types are now part of `joi`, if you want to double check that

adamjmcgrath (approved):
Thanks @veksen!

PR #343, opened 2021-03-17
[SDK-2405] Organisations support

adamjmcgrath:
### Description

Add support for Organisations (currently a Closed Beta)

- supports an `organization` config or `AUTH0_ORGANIZATION` environment variable which, when specified, passes the `organization` query param to the `/authorize` endpoint and validates it against the user's `org_id` ID Token claim.
- Allows the `organization` query param to be specified in the `loginHandler` and `callbackHandler` so that app's can support multiple Organzations.
- Document the `invitation` query parameter for the `/authorize` endpoint and add an example of creating a route that accepts Organization invites.


stevehobbsdev (approved):
🎉

PR #380, opened 2021-04-26
withPageAuthRequired adds user to ctx

hacknlove:
By submitting a PR to this repository, you agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md). Please see the [contributing guidelines](https://github.com/auth0/open-source-template/blob/master/CONTRIBUTING.md) for how to create and submit a high-quality PR for this repo.

### Description

I think it would be very convenient to make the user available to the getServerSideProps wrapped on withPageAuthRequired.

In terms of verbosity it's not a big deal, though it helps, but it's not just about typing less, it's about making sense: withPageAuthRequires adds user to the page props, and thus it makes a los of sense to add user to ctx too.

### Testing

I would be very happy to add tests for this, if the maintainers agrees on the proposed change.

I did it not yet, just to save the effort if the change is not welcome.

### Checklist

Same here. I'll be glad to add the documentation if this goes forward.

### DIFF
--- a/src/helpers/with-page-auth-required.ts
+++ b/src/helpers/with-page-auth-required.ts
@@ -115,7 +115,7 @@ export default function withPageAuthRequiredFactory(loginUrl: string, getSession
       }
       let ret: any = { props: {} };
       if (getServerSideProps) {
-        ret = await getServerSideProps(ctx);
+        ret = await getServerSideProps({ ...ctx, user: session.user });
       }
       return { ...ret, props: { ...ret.props, user: session.user } };
     };


adamjmcgrath:
Hi @hacknlove - thanks for raising this PR

My concern here is that this change will tie the current version of the SDK to the signature of `getServerSideProps` - if Vercel ever choose to add an argument to `getServerSideProps` in a nonn breaking change, we would not be able to support that without adding a breaking change or some runtime type checking to this library.

We could put `user` in the context argument, but for the sake of simplicty - I'd rather just require the user to call `getSession` in `getServerSideProps`, it's only an extra line of code and there is no overhead.

PR #399, opened 2021-05-18
Include message body in 302 responses

michielvangendt:
### Description

The 302 responses from the auth API do not contain a message body. This is in conflict with the RFCs below and causes Traefik (a reverse proxy) to invalidate the responses. In this pull request, I add a response body to the 302 responses.

### References

- https://datatracker.ietf.org/doc/html/rfc7230#section-3.3

> All 1xx (Informational), 204 (No Content), and 304 (Not Modified) responses must not include a message-body. All other responses do include a message-body, although the body may be of zero length.

- https://datatracker.ietf.org/doc/html/rfc7231#section-6.4.3

> The server's response payload usually contains a short hypertext note with a hyperlink to the different URI(s).

- https://github.com/traefik/traefik/issues/4456

### DIFF
--- a/src/auth0-session/handlers/callback.ts
+++ b/src/auth0-session/handlers/callback.ts
@@ -62,6 +62,6 @@ export default function callbackHandlerFactory(
     res.writeHead(302, {
       Location: openidState.returnTo || config.baseURL
     });
-    res.end();
+    res.end(openidState.returnTo || config.baseURL);
   };
 }
--- a/src/auth0-session/handlers/logout.ts
+++ b/src/auth0-session/handlers/logout.ts
@@ -54,6 +54,6 @@ export default function logoutHandlerFactory(
     res.writeHead(302, {
       Location: returnURL
     });
-    res.end();
+    res.end(returnURL);
   };
 }
--- a/src/auth0-session/handlers/login.ts
+++ b/src/auth0-session/handlers/login.ts
@@ -92,6 +92,6 @@ export default function loginHandlerFactory(
     res.writeHead(302, {
       Location: authorizationUrl
     });
-    res.end();
+    res.end(authorizationUrl);
   };
 }


adamjmcgrath:
Thanks @michielvangendt - thanks for raising this, looks good

Can we do what Next.js does and add `res.write('')`? eg

https://github.com/vercel/next.js/blob/canary/packages/next/next-server/server/api-utils.ts#L220

It will keep this fix consistent with how the redirect works in `withPageAuthRequired`

michielvangendt:
Hi @adamjmcgrath, it appears that their implementation (with `res.write('')`) doesn't work as well. I will open a PR/issue for Next.js tomorrow!

adamjmcgrath:
> it appears that their implementation (with `res.write('')`) doesn't work as well. I will open a PR/issue for Next.js tomorrow!

ok, np - can you reference this PR when you do? Thanks @michielvangendt

adamjmcgrath:
@michielvangendt - when vercel/next.js#25257 is merged can you update this to match and I'll go ahead and approve this.

michielvangendt:
@adamjmcgrath the PR in next.js is merged!

src/auth0-session/handlers/logout.ts
```diff
@@ -54,6 +54,6 @@ export default function logoutHandlerFactory(
     res.writeHead(302, {
       Location: returnURL
     });
-    res.end();
+    res.end(returnURL);
```
adamjmcgrath, line 57: I just noticed that you need to update lines 31 and 43

adamjmcgrath (changes requested):

src/auth0-session/handlers/callback.ts
```diff
@@ -62,6 +62,6 @@ export default function callbackHandlerFactory(
     res.writeHead(302, {
       Location: openidState.returnTo || config.baseURL
     });
-    res.end();
+    res.end(openidState.returnTo || config.baseURL);
```
adamjmcgrath, line 65: Also - I think we should escape these (and the others)

It's probably overkill because modern browsers wont serve this content, but `openidState.returnTo` can contain reflected user input from `/api/auth/login?returnTo=user-input`

I also notice that [express escapes the content in 302](https://github.com/expressjs/express/blob/master/lib/response.js#L935)

There's a util method to escape html here https://github.com/auth0/nextjs-auth0/blob/main/src/utils/errors.ts#L26-L34

FYI @lzychowski

PR #422, opened 2021-06-22
Support cross domain credentials

jamescryer:
### Description

I have two or more Next.js applications which are deployed independently. They provide different services to users, and live under different subdomains e.g. `login.mywebapp.com` and `people.mywebapp.com`.  For the user this is a single web application with a single session.

I'd like for the "login" web application to implement the nextjs-auth0 APIs, and have the other applications make use nextjs-auth0 frontend helpers without needing to implement the APIs as well. The other applications should use the APIs under the "login" sub domain.

I can almost make this work by 
- Setting the environment variables `NEXT_PUBLIC_AUTH0_LOGIN` and `NEXT_PUBLIC_AUTH0_PROFILE` in all web apps
- Setting the header `Access-Control-Allow-Origin` and `Access-Control-Allow-Credentials` in the API handler in the login web app.
- Setting the session.cookie.domain as the `.mywebapp.com` so that it is shared across sub domains.

But the cookie is not sent with the request to `https://login.mywebapp.com/api/auth/me`, returning an "unauthorised" response.

This PR makes it possible to send the auth cookie across sub domains by setting the `{ credentials: 'include' }` option in the `checkSession()` fetch call.

### Testing

- Set the environment variable `NEXT_PUBLIC_AUTH0_PROFILE` to a different URL, and add any cookie to that domain.  Upon loading the page a request will be sent to `https://login.mywebapp.com/api/auth/me` with the cookie sent on the request.

### Checklist


jamescryer:
Hey.  I'm leaving this as a draft PR for now as for some yet unknown reason I'm unable to install the examples to try the end-to-end tests.  It errors with `Could not resolve dependency: ... peer next@">=9.4.4" from @serverless-jwt/next@0.2.1`.  I don't have time to work through this right now.

But if automation can run the those test and you're all happy with the idea then perhaps we can mark as "ready for review"? Also, I've not added any documentation, let me know if you'd like something added.

adamjmcgrath:
Hi @jamescryer - thanks for raising this. I'd be happy to accept a PR that allowed you to optionally pass a custom fetcher (similar to [useSWR](https://swr.vercel.app/docs/data-fetching)) which should resolve your issue, eg

```js
<Auth0Provider
  fetcher={async (url) => {
      const response = await fetch(url, { credentials: 'include' });
      return response.ok ? response.json() : undefined;
  }}
```

jamescryer:
@adamjmcgrath Thanks for your feedback.  Do my changes work for you?

tests/frontend/use-user.test.tsx
```diff
@@ -157,4 +157,20 @@ describe('context wrapper', () => {
 
     expect(result.current.user).toEqual({ foo: 'bar' });
   });
+
+  test('should use the override fetch behaviour', async () => {
+    const fetchSpy = jest.fn();
+    (global as any).fetch = fetchSpy;
+
+    const customFetcher = jest.fn().mockReturnValue(Promise.resolve());
```
adamjmcgrath, line 165:
```js
const customFetcher = jest.fn().mockReturnValue(Promise.resolve('foo'));
// then    
expect(result.current.user).toBe('foo');
```

adamjmcgrath (changes requested):
lgtm - thanks @jamescryer 

just a small request on the test case

tests/frontend/use-user.test.tsx
jamescryer: Ah, good shout. I've made that change.

adamjmcgrath: Looks good, thanks - I just need to wait a day or two for something before I can merge. Will keep you posted

adamjmcgrath (approved):

adamjmcgrath:
Thanks @jamescryer - I'll try and get round to releasing it this week

PR #440, opened 2021-07-12
[SDK-2648] Update Vercel deploy instructions

### LINKED ISSUE #417: Hosting via Vercel not working because of AUTH0_BASE_URL.
### Description

Hi, I would like to reopen https://github.com/auth0/nextjs-auth0/issues/383, as https://github.com/auth0/nextjs-auth0/pull/404 did not fix the problem, as you can see in the last comments of https://github.com/auth0/nextjs-auth0/issues/383. 

The documentation at https://github.com/auth0/nextjs-auth0/blob/main/examples/README.md#assigning-the-auth0_base_url seems wrong to me, as the `env` key in config and the webpack `DefinePlugin` don't work for files in node_module, which are not touched by webpack.

Moreover, on a Vercel side, the VERCEL_URL reference the "by commit domain", and not the "by PR domain", so we can't use auth0 with the standard PR domain.

We have to figure out another way to do this.

Any thoughts about this?

Thanks.

### Environment

- **Version of this library used:** v1.4.0
- **Version of the platform or framework used, if applicable:** Next 10.2.3 with Vercel

adamjmcgrath:
### Description

Our original guidance to assign the `AUTH0_BASE_URL` for Vercel deployment didn't work because the env vars in the SDK weren't replaced by the `DefinePlugin` when the SDK was installed to `node_modules`

Update the instructions to use `.env.production` based on the guidance in this Next.js discussion https://github.com/vercel/next.js/discussions/15424#discussioncomment-41284

### References

Fixes #417 

### Testing

Tested on the Kitchen Sink App

### Checklist

### DIFF
--- a/.env.production
+++ b/.env.production
@@ -1,8 +1,0 @@
-# Set the baseUrl for the SDK to the Vercel "Automatic Branch URL" for Preview deploys. Use `${VERCEL_URL}` if you
-# would rather use the 'Automatic Deployment URL' (see: https://vercel.com/changelog/urls-are-becoming-consistent)
-#
-# For production deploys from the main branch or custom domains which are assigned to a specific branch (see https://vercel.com/docs/custom-domains#assigning-a-domain-to-a-git-branch)
-# you can assign the `AUTH0_BASE_URL` in the 'Environment Variables' section of your project's settings page which will override this.
-#
-# This file should be checked in so should NOT contain any secrets
-AUTH0_BASE_URL=${VERCEL_GIT_REPO_SLUG}-${VERCEL_GIT_COMMIT_REF}-${VERCEL_GIT_REPO_OWNER}.vercel.app
--- a/.env.production
+++ b/.env.production
@@ -1,0 +1,8 @@
+# Set the baseUrl for the SDK to the Vercel "Automatic Branch URL" for Preview deploys. Use `${VERCEL_URL}` if you
+# would rather use the 'Automatic Deployment URL' (see: https://vercel.com/changelog/urls-are-becoming-consistent)
+#
+# For production deploys from the main branch or custom domains which are assigned to a specific branch (see https://vercel.com/docs/custom-domains#assigning-a-domain-to-a-git-branch)
+# you can assign the `AUTH0_BASE_URL` in the 'Environment Variables' section of your project's settings page which will override this.
+#
+# This file should be checked in so should NOT contain any secrets
+AUTH0_BASE_URL=${VERCEL_GIT_REPO_SLUG}-${VERCEL_GIT_COMMIT_REF}-${VERCEL_GIT_REPO_OWNER}.vercel.app
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # @auth0/nextjs-auth0
 
-The Auth0 Next.js SDK is a library for implementing user authentication in Next.js applications
+The Auth0 Next.js SDK is a library for implementing user authentication in Next.js applications.
 
 [![CircleCI](https://img.shields.io/circleci/build/github/auth0/nextjs-auth0/main?style=flat-square)](https://circleci.com/gh/auth0/nextjs-auth0/tree/main)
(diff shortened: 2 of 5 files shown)


Widcket (approved):

mathieutu:
Unfortunately, the automatic branch url intructions doesn't actually work.

The vercel url is not composed from `[GIT_PROJECT]-git-[BRANCH_NAME]-[GIT_ORG]` but `[VERCEL_PROJECT]-git-[BRANCH_SLUG]-[VERCEL_ORG]`, and we do not have the type of info.

Ex:
I've ``console.log(`${VERCEL_GIT_REPO_SLUG}-git-${VERCEL_GIT_COMMIT_REF}-${VERCEL_GIT_REPO_OWNER}.vercel.app`)`` in a api route.

For the 
`https://mcube-git-fix-auth0-vercel-epigene.vercel.app` url, I'm loging
`data-curator-app-git-fix/auth0-epigenelabs.vercel.app`, which is quite different!

I don't know how we can solve this issue without the direct help of Vercel... 😞

mathieutu:
Linked: 
https://github.com/vercel/vercel/discussions/6656
https://github.com/vercel/vercel/discussions/6045

adamjmcgrath:
That's odd because it works for the kitchen-sink app:

The URL:
`https://nextjs-auth0-git-set-vercel-base-url-auth0.vercel.app/`

Generates the following callback URL:
`https://nextjs-auth0-git-set-vercel-base-url-auth0.vercel.app/api/auth/callback`

The Git repo is `nextjs-auth0`
The Branch is `set-vercel-base-url`
And the Git repo owner is `auth0`

TBH you could hardcode the repo and owner if the Vercel env vars don't work for you, as those names are not going to change.

I see what you mean about the `VERCEL_GIT_COMMIT_REF` though when it contains chars that aren't URL safe (like `/`). If you want to use non URL safe chars in your branch names, you'll need to do some substitution yourself, eg `${VERCEL_GIT_COMMIT_REF//\//$'-'}`

mathieutu:
Actually it works because your Vercel project is `nextjs-auth0` and your Vercel team is `auth0`.
It has no link with Git variables, so in your case it's a matter of coincidence. 😅

I've indeed hardcoded them (and I think we should edit the docs to say to do the same, avoiding taking your coincidence as a generality).

We can use substitution directly in .env? 😮

I'm digging around using the `req.headers.host` to definitely fix this issue.
I've read your thougts about integrating it directly in the sdk, but we actually could let it disabled by default and use a flag to enable it?

mathieutu:
>We can use substitution directly in .env? 😮

I've pasted your string to try it and... except if I've misunderstood something, it definitely doesn't work! 😅

adamjmcgrath:
> Actually it works because your Vercel project is nextjs-auth0 and your Vercel team is auth0.
> It has no link with Git variables, so in your case it's a matter of coincidence. 😅

Ah ok, the Vercel docs must be wrong then
image

> I've read your thougts about integrating it directly in the sdk, but we actually could let it disabled by default and use a flag to enable it?

I'm not interested in adding something that depends on `req.headers.host` for the reasons I've already described. But I'll have another look at the docs and the `VERCEL_GIT_COMMIT_REF` issue

mathieutu:
>vercel docs must be wrong then.

No, there right. The variables are ok.
It's the branch URL that is not constructed from them.
If I quote vercel/vercel#6656: 

> I've found that the "git" url follows this pattern `https://<project>-git-<branch>-<org>.vercel.app`, where :
>
> `<project>` is the project name
>    `<branch>` is a the url friendly name of the git branch,
>    `<org>` is the name of the Vercel Team or personal account.

I've updated my project to use the host instead of env var (and it works pretty well), and sent a message to Vercel about the fact that we can't retrieve this branch URL.

stephanebachelier:
@mathieutu Glad you've found my issue on Vercel project. I'm using auth0 through next-auth but the issue would be the same for any other OAuth provider.

Issue #471, opened 2021-08-22
Add ability to send user directly to signup page on login


jonkcinema:
### Describe the problem you'd like to have solved

Most webapps have some sort of "create account" button.  The auth0-react library supports this via the screen_hint parameter:

`loginWithRedirect({ screen_hint: 'signup' })`

The nextjs-auth0 library does not appear to support this.

### Describe the ideal solution

Please add a way to send the user directly to the signup page.

E.g. `<a href="/api/auth/login?screenHint=signup">Sign up</a>`

## Alternatives and current work-arounds

You can create your own signup handler (e.g. pages/api/signup.js), but that seems pretty unfortunate given how simple and common of an operation this is.

adamjmcgrath:
Hi @jonkcinema - thanks for raising this

How you do signup depends on which Universal Login Page you are using see https://community.auth0.com/t/how-do-i-redirect-users-directly-to-the-hosted-signup-page/42520 for more information

Because there are different APIs for the new Universal Login experience and the Classic one (there is actually no specific API for the classic experience), and there is no way for the SDK to know which Login experience your tenant is set to - it's not possible to provide a `signup` api in the SDK. You'll need to create one yourself as you've described in the workaround.

jonkcinema:
Then why does the auth0-react library make it so easy?  It seems like it
would be an improvement to at least make the more common / simpler approach
supported, or at a minimum, improve the documentation so I could figure
this out without having to read through GitHub issues.  It seems like
enough people are struggling with this that the status quo isn’t ideal.

adamjmcgrath:
Hi @jonkcinema 

The implimentation of `auth0-react` and this SDK are very similar and they both have the same constraints regarding the Classic and New Universal Login Pages:

```js
// auth0-react
loginWithRedirect({ screen_hint: 'signup' })

// nextjs-auth0
handleLogin(req, res, { authorizationParams: { screen_hint: 'signup' }});
```

jonkcinema:
My point is, figuring this out was a pain.  It would be nice if you guys
made it easier.  At a minimum, through better documentation, and ideally
with a slight API modification like I suggested.  The auth0-react library,
which you say is very similar, supports this with just the addition of the
screen hint param.  Why does the nextjs lib require me to write a custom
handler?  Why can’t I just pass a screen hint param somewhere?

Anyway, my point is, your service is really great, but this is the one
thing that felt much harder to do than it should be.  And based on the
[...]

Issue #400, opened 2021-05-19
Route to signup page


thgh:
### Describe the problem you'd like to have solved

I would like to link to the signup page.

### Describe the ideal solution

Add an API route `/api/auth/signup` that redirects. (similar to /api/auth/login)

## Alternatives and current work-arounds

Combine the login & signup call-to-action into 1 button.

adamjmcgrath:
Hi @thgh - Assuming you're using the new login experience,  you can create your own signup page like so:

Create `pages/api/signup.js`

```js
import auth0 from '../../lib/auth0';

export default async function signup(req, res) {
  try {
    await auth0.handleLogin(req, res, {
      authorizationParams: {
        screen_hint: 'signup',    // this prompts the signup screen
      },
    });
  } catch (error) {
    console.error(error);
    res.status(error.status || 500).end(error.message);
  }
}
```

Then link to the signup page:

```
<a href='/api/signup'>Sign Up</a>
```

If you're using the Classic login experience you might want to use a different `authorizationParams` key and there's a little more work to do on your custom login page, see: https://community.auth0.com/t/how-do-i-redirect-users-directly-to-the-hosted-signup-page/42520

orbiteleven:
TBH this seems like a normal enough use case that the lib could/should provide it out of the box.

lucasvmiguel:
Why was this closed? As @orbiteleven, this seems to be an important piece to implement in this SDK.

NathanLandsX:
Major oversight that this isn't a main part of the SDK. People need signup before login...

kkomelin:
Also need the Signup route.

epark10:
Bumping as signup is a critical part of the user identity process, should be included

Issue #379, opened 2021-04-23
Support for a /signup handler


chrispeterson3:
### Describe the problem you'd like to have solved
- Support for a signup handler for `handleAuth`

When creating a new application or service, there should be a quick and easy way for a CTA for "Sign up". It seems that today a user would have to go to `/api/auth/login` and then click "Sign up".

### Describe the ideal solution
A handler for `/api/auth/signup`

## Alternatives and current work-arounds
The suggested workarounds don't seem to work suggested in #24 . After calling `initAuth0`, the types for the options parameter on `handleLogin` do not support `authParams` it seems.

### Additional information, if any
![image](https://user-images.githubusercontent.com/2593324/115935272-ae5e9700-a447-11eb-992a-1ece2534205f.png)

chrispeterson3:
It looks like a bit of user error on my part. `authParams` was changed to `authorizationParams`. This implementation will work.

```import { initAuth0 } from '@auth0/nextjs-auth0';

const auth0 = initAuth0({
  baseURL: process.env.AUTH0_BASE_URL,
  issuerBaseURL: process.env.AUTH0_ISSUER_BASE_URL,
  clientID: process.env.AUTH0_CLIENT_ID,
  clientSecret: process.env.AUTH0_CLIENT_SECRET,
  secret: process.env.AUTH0_SECRET,
});

export default async function signup(req, res) {
  try {
    await auth0.handleLogin(req, res, {
      authorizationParams: {
        screen_hint: 'signup',
      },
    });
  } catch (error) {
    res.status(error.status || 500).end(error.message);
  }
}
```

A signup handler would be nice though!

samkahchiin:
Thanks for the sharing! This will works as well. :D

```
import { handleLogin } from '@auth0/nextjs-auth0'

export default async function signup(req, res) {
  try {
    await handleLogin(req, res, {
      authorizationParams: {
        screen_hint: 'signup',
      },
    })
  } catch (error) {
    res.status(error.status || 500).end(error.message)
  }
}
```

Issue #16, opened 2019-10-08
Handle Signup


matthieuh:
### Description

It seems that using this lib that there is no way to redirect user to signup part of the universal login.

### Describe the ideal solution

I would expect a `auth0.handleSignup`.

I'm maybe missing something but If I'm right I would enjoy open a PR.

sandrinodimattia:
Hi @matthieuh , here you go: https://github.com/auth0/nextjs-auth0/pull/24 (v0.5.0). You could implement a new API route (eg: `/api/signup`) which calls the login handler with specific options which will be used by your login page (eg: `mode=signup`).

Feel free to re-open the issue if this doesn't solve it for you.

matthieuh:
Using the login handler with a mode parameter seems not redirecting user to signup screen with the latest lib version (`v0.5.0`).
```javascript
await auth0.handleLogin(req, res, {
  authParams: {
    mode: 'signup'
  }
});
```

torresandres:
I have this working by passing the "signUp" value as `initialScreen` instead of `mode`:

```
await auth0.handleLogin(req, res, {
  authParams: {
    initialScreen: 'signUp'
  }
});
```

jvgreenaway:
I've tried all the options suggested on this issue but I cannot get this working – I land on the sign in view every time. I‘m using the universal login form. 

Has anybody managed to get this to work recently?

Nobotym:
Try this:
`{ authParams: { screen_hint: 'signup' } }`

kjkurtz:
I tried both of these in different configurations:
`      authParams:{
        initialScreen: 'signUp',
        screen_hint: 'signup'
      }`
And I still ended up with Sign In not Sign Up. So no-go on my end right now!

dayvough:
I'm trying out the approach here using the new universal login: https://community.auth0.com/t/how-do-i-link-users-directly-to-the-hosted-signup-page/42520

```js
auth0.handleLogin(req, res, {
  authParams: {
    prompt: "login",
    screen_hint: "signup",
  },
});
```

And it's still leading me to the Login and not the Sign Up tab.

dayvough:
Hmm, decided to just jump into my auth0 config and follow https://github.com/auth0/nextjs-auth0/issues/25#issuecomment-649310807 instead

```js
auth0.handleLogin(req, res, {
  authParams: {
    prompt: "signup",
  },
});
```

This then worked for me.

nodabladam:
[...]

PR #474, opened 2021-08-23
Add an example for signup

adamjmcgrath:
### Description

Add an example for creating a signup handler

### References

See #471 #16 #400 #379 #124

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -367,3 +367,9 @@ var lock = new Auth0Lock(config.clientID, config.auth0Domain, {
   initialScreen: isSignup ? 'signUp' : 'login'
 });
 ```
+
+Users can then signup using the signup handler.
+
+```html
+<a href="/api/signup">Sign up</a>
+```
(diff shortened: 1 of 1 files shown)


EXAMPLES.md
```diff
@@ -326,3 +327,49 @@ export default auth0.handleAuth({
...
+```js
+var isSignup = config.extraParams && config.extraParams.action === 'signup';
+var lock = new Auth0Lock(config.clientID, config.auth0Domain, {
+  // [...] all other Lock options
+  // use the value obtained to decide the first screen
+  initialScreen: isSignup ? 'signUp' : 'login'
+});
+```
+
+Users can then signup using the signup handler.
```
Widcket, line 371:
```suggestion
Users can then sign up using the signup handler.
```

Widcket (changes requested):
Looks good, just left a suggestion for a typo

Widcket (approved):

PR #485, opened 2021-09-06
Automate package publishing

lbalmaceda:
Makes use of our script to publish to npm/github when a new release PR is done.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,6 @@
     "lint": "eslint --fix --ext .ts ./src",
     "docs": "typedoc --options typedoc.js src",
     "prepack": "npm run build",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -21,7 +21,6 @@
     "lint": "eslint --fix --ext .ts ./src",
     "docs": "typedoc --options typedoc.js src",
     "prepack": "npm run build",
-    "prepublishOnly": "npm test && npm run lint",
```
lbalmaceda, line 24: tests are being run during the `build` job. But the lint wasn't. Added that command to the job and removed this hook, so they are not re-run in the new job.

Widcket (approved):

Issue #487, opened 2021-09-07
Set-Cookie does not work for large cookies


ludvigsen:
### Describe the problem

When trying to login with the following code.
```
      await session.handleLogin(req, res, {
        authorizationParams: {
          connection: loginConnection,
          acr_values: loginAcrValues,
          tenantId
        },
        returnTo
      });

```

I sometimes experience that Set-Cookie is ignored because the cookie is larger than 4096 bytes. 

This is what is shown in chrome console:
![image](https://user-images.githubusercontent.com/124673/132343166-2bbbe85e-d4c6-4574-aa0f-507e151b321c.png)

### What was the expected behavior?

I expect the cookie to be split into multiple parts. It is reported as fixed here:
https://github.com/auth0/nextjs-auth0/issues/91
But it does not seem to work. Are there any config options that must be set?

### Reproduction

- Step 1: Make a user with a lot of metadata
- Step 2: Try to login using handleLogin
- Step 3: Observe that the user is not logged in.

### Environment

- **Version of this library used:** 1.3.1
- **Which framework are you using, if applicable:** NextJS
- **Other modules/plugins/libraries that might be involved:** -
- **Any other relevant information you think would be useful:** This only happens in chrome I believe, not in firefox.

adamjmcgrath:
Hi @ludvigsen - thanks for raising this

That's odd - we have a pretty good regression test for this https://github.com/auth0/nextjs-auth0/blob/main/tests/auth0-session/cookie-store.test.ts#L104-L121

Can you share the value of one of the `appSession.X` `Set-Cookie` headers (and any custom cookie config you might have)? I want to take a look at the cookie metadata.

ludvigsen:
We don't use any custom cookie config. 

Set-Cookie headers:
```
[...]

PR #488, opened 2021-09-08
Add maxChunkSize to cookie options

ludvigsen:
### Description

Added a config option to the cookie config in order to be able to control the chunk size of cookies.

As discussed in this issue
https://github.com/auth0/nextjs-auth0/issues/487

We experienced our cookie being rejected due to it being too large. This turned out to be caused by us enabeling HSTS on our ingress. Which causes `; Secure` to be appended to all `Set-Cookie` headers.
https://voyagermesh.com/docs/7.0.0/guides/ingress/http/hsts/

Seems like the solution for us will be to disable secure in the cookie config and lowering the chunk size to 4088.

### References
https://github.com/auth0/nextjs-auth0/issues/487
https://voyagermesh.com/docs/7.0.0/guides/ingress/http/hsts/

### Testing
Added unit test that verifies that reducing maxChunkSize works.

### DIFF
--- a/src/auth0-session/config.ts
+++ b/src/auth0-session/config.ts
@@ -240,6 +240,11 @@ export interface CookieConfig {
    * Defaults to "Lax" but will be adjusted based on {@link AuthorizationParameters.response_type}.
    */
   sameSite: 'lax' | 'strict' | 'none';
+
+  /**
+   * Max size of the cookie chunks, use to override the default of 4096.
+   */
+  maxChunkSize: number;
 }
 
 export interface AuthorizationParameters extends OidcAuthorizationParameters {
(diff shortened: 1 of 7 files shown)


adamjmcgrath:
This looks like more of an issue with your HSTS configuration. Appending `; Secure` to the end of cookies regardless of whether they already include it seems off to me:

```sh
# Mark all cookies as secure
rsprep ^Set-Cookie:\ (.*) Set-Cookie:\ \1;\ Secure
```

https://voyagermesh.com/docs/7.0.0/guides/ingress/http/hsts/

Strictly speaking

> servers SHOULD NOT produce two attributes with the same name in the same set-cookie-string

https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1

See also https://github.com/voyagermesh/voyager/issues/1366

ludvigsen:
Yes, I agree it should not and append to `; Secure` to headers where it's already present.
Theis PR just dermed like the easiest solition to the problem.
We don't want to disable HSTS.
Also I see it as a useful feature to allow appending of headers while not breaking chunking?

adamjmcgrath:
Sorry @ludvigsen - we don't want to add support this use case. You're welcome to fork and we'll keep reviewing this decision if more information comes to light

PR #506, opened 2021-09-28
[SDK-2813] Add afterRefresh hook

### LINKED ISSUE #416: Modifications to session after login are overwritten over time
### Describe the problem

Modifications to session after login are overwritten over time

If code exists to modify the session after login like so (following the example [in the docs](https://auth0.github.io/nextjs-auth0/modules/handlers_callback.html#modify-the-session-after-login)):
```
const afterCallback = (req, res, session, state) => {
  delete session.idToken;
  return session;
};

export handleAuth({
  async callback(req, res) {
    try {
      await handleCallback(req, res, { afterCallback });
    } catch (error) {
      res.status(error.status || 500).end(error.message);
    }
  }
});
```

Then those modifications should persist indefinitely, irrespective of the session being updated at a later point in time. I'm not exactly sure what is causing the session cookie contents to be updated, it could be either:
- Because `rolling: true` in session config 
- Because `offline_access` has been added to the scope param to benefit from refresh tokens

Either way the code written in `afterCallback` obviously has no bearing on the code that updates the session after the fact.

### Reproduction

- Enable either rolling sessions or add the offline_access scope in the authorisation params
- Follow the [guide in the docs](https://auth0.github.io/nextjs-auth0/modules/handlers_callback.html#modify-the-session-after-login) to modify the session after login - e.g. remove the idToken
- after a certain amount of time (perhaps when the access token expires?) the session will be updated and now include the idToken

### Environment

- **Version of this library used:** 1.3.1
- **Which framework are you using, if applicable:** Nextjs 10.2.3

adamjmcgrath:
### Description

When the Access Token is refreshed, the session gets a new set of tokens (Access Token, ID Token, Refresh Token).

Our advice for dealing with large cookies is to delete optional data from the session (like ID Token in some cases).

In order to do this and have the tokens refreshed, we need to add another hook (this time to `getAccessToken`) that allows you to modify the session, so that changes like removing the ID Token in `afterCallback` aren't reverted in `getAccessToken` when the session is refreshed.

Ideally we'd have a `sessionUpdated` hook, that runs accross all scenarios where a session is updated, but we don't have a place to configure a non literal config globally (all config should be able to be defined as an environment variable and I don;t want to introduce features that require `initAuth`) so we'll have to make do with these 3 hooks:

- `afterCallback`: When the session is establised including tokens and user
- `afterRefetch`: When the user is potentially updated from `/userinfo`
- `afterRefresh`: When the tokens are potentially updated from the Refresh grant

### References

Fixes #416 
See [updated FAQ](https://github.com/auth0/nextjs-auth0/blob/c80ea310f0f44df75cc85e8272debbd0661a78b7/FAQ.md#2-how-can-i-reduce-the-cookie-size)

### Testing

### Checklist

### DIFF
--- a/src/session/get-access-token.ts
+++ b/src/session/get-access-token.ts
@@ -3,8 +3,10 @@ import { NextApiRequest, NextApiResponse } from 'next';
 import { ClientFactory } from '../auth0-session';
 import { AccessTokenError } from '../utils/errors';
 import { intersect, match } from '../utils/array';
-import { SessionCache, fromTokenSet } from '../session';
+import { Session, SessionCache, fromTokenSet } from '../session';
 import { NextConfig } from '../config';
+
+export type AfterRefresh = (req: any, res: any, session: Session) => Promise<Session> | Session;
 
 /**
  * Custom options to get an Access Token.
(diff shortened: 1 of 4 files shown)


src/session/get-access-token.ts
```diff
@@ -120,6 +150,12 @@ export default function accessTokenFactory(
         user: { ...session.user, ...newSession.user }
       });
 
+      if (accessTokenRequest?.afterRefresh) {
+        session = await accessTokenRequest.afterRefresh(req as NextApiRequest, res as NextApiResponse, session);
+      }
+
+      sessionCache.set(req, res, session as Session);
```
Widcket, line 157: Why does this require a cast, given that successfully awaiting `afterRefresh` produces a Session?

adamjmcgrath: `sessionCache.get` can return `undefined` (if a session hasn't been estabilished) and TS isn't smart enough to know that there is a session if `client.refresh` is successful

Widcket (dismissed):

src/session/get-access-token.ts
adamjmcgrath: Actually, my bad, I don;t need to cast it - will fix, thanks

adamjmcgrath: fixed c80ea310f0f44df75cc85e8272debbd0661a78b7

Widcket (approved):

Issue #532, opened 2021-11-07
Enable the use of returnTo params in the logout endpoint.


aiji42:
### Describe the problem you'd like to have solved
The endpoint for login (/api/auth/login) allows the use of the `?returnTo=` parameter, but the endpoint for logout is ignored even if the same is given.  
The `interface LogoutOptions` has a `returnTo` property, so we can control where to redirect after logout. However, to do that, you have to define `logout.js` separately from `[...auth0.js]`.

### Describe the ideal solution
Make it possible to receive `returnTo` as a URL parameter when logout only by setting the endpoint with `handleAuth` (`[...auth0].js`).  
If a user logs out with `/api/auth/logout?returnTo=/top`, the user will be redirected to `/top`.

## Alternatives and current work-arounds

Currently, I define a separate endpoint for logout as below.

![スクリーンショット 2021-11-07 午後0 09 56](https://user-images.githubusercontent.com/6711766/140631033-98a68136-4389-42b0-b981-7c64a89d0678.png)

```ts
import { handleLogout } from '@auth0/nextjs-auth0'
import { NextApiHandler } from 'next'

const handler: NextApiHandler = (req, res) => {
  const { returnTo } = req.query
  return handleLogout(req, res, {
    returnTo: Array.isArray(returnTo) ? returnTo[0] : returnTo
  })
}

export default handler
```

adamjmcgrath:
Thanks for raising this @aiji42 - Going to close, see my comment https://github.com/auth0/nextjs-auth0/pull/533#event-5598796393

aiji42:
@adamjmcgrath Thanks for the reaction.
I understand your idea. However, I think there is room for improvement in having to separate `[...auth0].js` and `logout.js` to specify `returnTo`.
I think it would be useful to be able to specify that with `handleAuth` arguments or environment variables. Please let me know what you think.

adamjmcgrath:
> I think it would be useful to be able to specify that with handleAuth arguments or environment variables.

Hi @aiji42 - you can already specify it as an environment variable using `AUTH0_POST_LOGOUT_REDIRECT` see https://auth0.github.io/nextjs-auth0/interfaces/config.baseconfig.html#routes

aiji42:
@adamjmcgrath
Oh, I see! I had overlooked the documentation. Thank you very much for your help.

dr-skot:
I know this is closed but I'd like to upvote this feature request.

[...]

PR #533, opened 2021-11-07
Allow logout API to accept returnTo query string.

aiji42:
### Description

Make it possible to receive returnTo as a query string with logout api endpoint.

### References

#532 

### Testing

I added some tests to `tests/handlers/logout.test.ts`.

`npm run test`

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ Executing `handleAuth()` creates the following route handlers under the hood tha
 
 - `/api/auth/callback`: Your Identity Provider redirects users to this route after they successfully log in.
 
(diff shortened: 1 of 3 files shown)


adamjmcgrath:
Hey @aiji42 - thanks for raising this!

Your implementation looks good - however, `post_logout_redirect_uri`'s need to be registered on your IdP in advance (the equivalent `returnTo` on login doesn't) so we don't want to have them populated dynamically by the route's query param like we do with `login`

Please continue to use the workaround you described in #532

PR #543, opened 2021-11-18
Fix issue where error reporting wrong instanceof

### LINKED ISSUE #535: Can no longer check if an error is instance of AccessTokenError
### Description

Using the [instanceof](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof) operator against the `AccessTokenError` constructor always returns false, even if the object that is being tested is actually an instance of `AccessTokenError`.

This is an issue introduced by `nextjs-auth0` **v1.5.0**.

### Reproduction

With `nextjs-auth0` **v1.4.2** I'm able to check if an error occurred inside a NextJS API handler is an instance of `AccessTokenError` by doing so:

```jsx
import { getAccessToken } from '@auth0/nextjs-auth0';
import { AccessTokenError } from '@auth0/nextjs-auth0/dist/utils/errors';
import { ApiError } from 'src/lib/apiClient';

try {
	const tokenInfo = await getAccessToken(req, res);

        // ...

} catch (err: any) {

         // Re-throw an API error in a more convenient format
	throw new ApiError({

                 // Decide what status to return depending on the type of exception that was thrown
		status: err instanceof AccessTokenError ? 401 : 500,
		message: `Code: ${err?.code}, Message: ${err?.message}.`,
	});
}
```

With `nextjs-auth0` **v1.5.0**, I always get a 500 error because errors thrown by `getAccessToken` are always an instance of `Error`, instead `AccessTokenError`.

At least, this is the reason why our unit test on that API endpoint handler is currently failing.

### Environment

- **Library version: v1.5.0**
- **Platform: macOS Big Sur**
- **Languages: Node v14.15.4, TS 4.4.4 (target ES6)**
- **Other modules/plugins/libraries that might be involved: Jest 26.6.3**

adamjmcgrath:
### Description

Using the `instanceof` operator against the `AccessTokenError` constructor always returns false, even if the object that is being tested is actually an instance of `AccessTokenError`.

This is an issue introduced in nextjs-auth0 v1.5.0 by https://github.com/microsoft/TypeScript/issues/13965 when we downgraded the compilation target to `es5` to support IE11 in https://github.com/auth0/nextjs-auth0/commit/c248efe2488b9884d6ab4c88c12b00aa67fe99f7#diff-3ae20d611c1c7263a9c1bce0449291ebfea1c5d578b3fcdbb596353ad885db25R19

Using the workaround here https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#extending-built-ins-like-error-array-and-map-may-no-longer-work

### References

fixes #535 

### Testing

Set the target for jest to be the same as the main binary, so I could add a test case for this regression.
Using target `es5` on jest meant I had to add a couple of `ignore next`'s to resolve an issue in istanbul (see https://github.com/gotwarlost/istanbul/issues/690)

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -149,13 +149,6 @@
       }
     },
     "preset": "ts-jest",
-    "globals": {
-      "ts-jest": {
-        "tsconfig": {
-          "target": "es6"
-        }
-      }
-    },
     "globalSetup": "./tests/global-setup.ts",
     "setupFilesAfterEnv": [
       "./tests/setup.ts"
--- a/src/utils/errors.ts
+++ b/src/utils/errors.ts
@@ -8,6 +8,7 @@ import { HttpError } from 'http-errors';
 export class AccessTokenError extends Error {
   public code: string;
 
+  /* istanbul ignore next */
   constructor(code: string, message: string) {
     super(message);
 
@@ -19,6 +20,7 @@ export class AccessTokenError extends Error {
 
     // Machine readable code.
     this.code = code;
+    Object.setPrototypeOf(this, AccessTokenError.prototype);
   }
 }
 
@@ -48,6 +50,7 @@ export class HandlerError extends Error {
   public status: number | undefined;
   public code: string | undefined;
 
+  /* istanbul ignore next */
   constructor(error: Error | AccessTokenError | HttpError) {
     super(htmlSafe(error.message));
 
@@ -60,5 +63,6 @@ export class HandlerError extends Error {
     if ('status' in error) {
       this.status = error.status;
     }
+    Object.setPrototypeOf(this, HandlerError.prototype);
   }
 }
--- a/tests/utils/errors.test.ts
+++ b/tests/utils/errors.test.ts
@@ -1,0 +1,8 @@
+import { AccessTokenError, HandlerError } from '../../src/utils/errors';
+
+describe('errors', () => {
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -149,13 +149,6 @@
       }
     },
     "preset": "ts-jest",
-    "globals": {
```
adamjmcgrath, line 152: Using same target as project (`es5`) so I can have a regression test for this issue

src/utils/errors.ts
```diff
@@ -8,6 +8,7 @@ import { HttpError } from 'http-errors';
 export class AccessTokenError extends Error {
   public code: string;
 
+  /* istanbul ignore next */
```
adamjmcgrath, line 11: See https://github.com/gotwarlost/istanbul/issues/690

tests/utils/errors.test.ts
```diff
@@ -0,0 +1,8 @@
+import { AccessTokenError, HandlerError } from '../../src/utils/errors';
+
+describe('errors', () => {
+  test('should be instance of themselves', () => {
```
adamjmcgrath, line 4: Believe it or not, this test will fail in TS with `target: es5` without the `Object.setPrototypeOf(this, ThisError.prototype);` workaround

Widcket (approved):

PR #555, opened 2021-12-08
Fix tab character open redirect

cfbender:
### Description

Fixes an open redirect vulnerability via the `returnTo` parameter, via checking for whitespace characters.

Chrome based browsers (in my testing) will trim off the tab character for a redirect, as [defined in the URL parser spec](https://url.spec.whatwg.org/#concept-basic-url-parser). This leads to an open redirect through `returnTo` by setting it to something like `/%09/google.com`.

You can test this in chrome by executing the following in the console:

```javascript
window.location.href = '/\x09/google.com'
```

### Testing

Include `returnTo=/%09/google.com` in the login, it should fail with this change.

### Checklist

### DIFF
--- a/src/utils/url-helpers.ts
+++ b/src/utils/url-helpers.ts
@@ -8,7 +8,7 @@ export default function isSafeRedirect(url: string): boolean {
   }
 
   // Prevent open redirects using the //foo.com format (double forward slash).
-  if (/\/\//.test(url)) {
+  if (/\/\s*\//.test(url)) {
     return false;
   }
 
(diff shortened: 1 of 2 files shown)


adamjmcgrath:
Hi @cfbender  - thanks for raising this, much appreciated!

We're working on a slightly different fix for this right now, and will raise another PR shortly

In future, we kindly ask that you use the [Responsible Disclosure Program](https://auth0.com/responsible-disclosure-policy) to report security issues

cfbender:
@adamjmcgrath my mistake, thank you so much for pointing me toward that! I had no idea that was a thing. I will absolutely keep that in mind for this and other projects.

Thanks for your diligent response as well!

PR #557, opened 2021-12-14
Enforce configured host on user supplied returnTo

### LINKED ISSUE #545: When somewhere in the returnTo Url a absolute url found -> Exeption

adamjmcgrath:
### Description

This PR changes how the user supplied `returnTo` param from `/api/auth/login` is sanitized.

Instead of using a regexp like a block list to prevent URLs that aren't relative from being accepted, it converts the URL to absolute using the configured `redirect_uri` as a base. It then enforces that the absolute URL has the same origin as the configured `redirect_uri`. 

This will ensure that only valid URLs with the same origin as the configured `redirect_uri` (or `baseURL`) are accepted as `returnTo` URLs.

### References

Fixes #545 
See also #555 

### Testing

Tested using https://github.com/adamjmcgrath/open-redirect-finder


tests/utils/url-helpers.test.ts
```diff
@@ -1,25 +1,522 @@
...
+      ];
+
+      for (const payload of payloads) {
+        expect(toSafeRedirect(payload, safeBaseUrl) || safeBaseUrl.toString()).toMatch(/^http:\/\/www.example.com\//);
+      }
     });
 
-    test('should throw when non string provided', () => {
-      expect(isSafeRedirect.bind(null)).toThrow('Invalid url: undefined');
+    test('should throw for empty redirect', () => {
```
Widcket, line 518: Test says `should throw` but the assert is `.not.toThrow()`.

tests/utils/url-helpers.test.ts
```diff
@@ -1,25 +1,522 @@
...
+      expect(toSafeRedirect('/foo?someUrl=https://www.google.com', safeBaseUrl)).toEqual(
+        'http://www.example.com/foo?someUrl=https://www.google.com'
+      );
+      expect(toSafeRedirect('/foo', new URL('http://www.example.com:8888'))).toEqual('http://www.example.com:8888/foo');
     });
 
     test('should prevent open redirects', () => {
-      expect(isSafeRedirect('//google.com')).toEqual(false);
-      expect(isSafeRedirect('///google.com')).toEqual(false);
+      const payloads = [
```
Widcket, line 24: This is a rather large list of payloads. Could you extract it into a json file (wrapped in an object)?

tests/utils/url-helpers.test.ts
```diff
@@ -1,25 +1,522 @@
...
+        '///whitelisted.com@google.com/%2f%2e%2e',
+        '//whitelisted.com@google.com/%2f%2e%2e',
+        '//whitelisted.com+&@google.com#+@whitelisted.com/',
+        '//whitelisted.com@https:///Ⓛ𝐨𝗰�𝕝ⅆ𝓸ⓜₐℹⓃ｡Ｐⓦ/%2e%2e',
+        '//whitelisted.com@https:///google.com/%2e%2e',
+        '//whitelisted.com+&@Ⓛ𝐨𝗰�𝕝ⅆ𝓸ⓜₐℹⓃ｡Ｐⓦ#+@whitelisted.com/'
+      ];
+
+      for (const payload of payloads) {
+        expect(toSafeRedirect(payload, safeBaseUrl) || safeBaseUrl.toString()).toMatch(/^http:\/\/www.example.com\//);
```
Widcket, line 514: Shouldn't this be asserting that the outcome of calling `toSafeRedirect` is `undefined`?

Widcket (changes requested):
The implementation looks good, just there are some improvement opportunities in the tests.

tests/utils/url-helpers.test.ts
adamjmcgrath: Good catch

Widcket (approved):

PR #564, opened 2021-12-21
Include message body in 302 responses

michielvangendt:
### Description
The 302 responses from the auth API do not contain a message body. This is in conflict with the RFCs below and causes Traefik (a reverse proxy) to invalidate the responses. In this pull request, I add a response body to the 302 responses.

### References
- My previous PR, sorry for not following up on this! https://github.com/auth0/nextjs-auth0/pull/399
- https://datatracker.ietf.org/doc/html/rfc7230#section-3.3

> All 1xx (Informational), 204 (No Content), and 304 (Not Modified) responses must not include a message-body. All other responses do include a message-body, although the body may be of zero length.

- https://datatracker.ietf.org/doc/html/rfc7231#section-6.4.3

> The server's response payload usually contains a short hypertext note with a hyperlink to the different URI(s).

- Similar fix in Next.js: https://github.com/vercel/next.js/pull/25257
- Similar fix in Next.js: https://github.com/vercel/next.js/pull/31886

### DIFF
--- a/src/auth0-session/handlers/callback.ts
+++ b/src/auth0-session/handlers/callback.ts
@@ -6,6 +6,7 @@ import { ClientFactory } from '../client';
 import TransientStore from '../transient-store';
 import { decodeState } from '../hooks/get-login-state';
 import { SessionCache } from '../session-cache';
+import { htmlSafe } from '../../utils/errors';
 
 function getRedirectUri(config: Config): string {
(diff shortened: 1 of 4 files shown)


adamjmcgrath (approved):
Thanks @michielvangendt

adamjmcgrath:
Will put up a release after the holidays

PR #570, opened 2022-01-04
Cookies with samesite=none must have the secure attr set

### LINKED ISSUE #569: Clearcookie Doesn't Respect AUTH0_COOKIE_SAME_SITE
### Describe the problem
It seems that the `clearCookie` function used to remove the `appSession` cookie doesn't respect the `AUTH0_COOKIE_SAME_SITE` environment variable.

This means that if you set `AUTH0_COOKIE_SAME_SITE=none` to support cross-origin requests, users can log in, but when they try to logout, the `appSession` cookie isn't cleared.

The `clearCookie` function should respect `AUTH0_COOKIE_SAME_SITE` when being used to clear `appSession`.

### Environment

- **Version of this library used:** 1.6.1

adamjmcgrath:
### Description

Fix issue where transient and session cookies weren't cleared when `SameSite=None`

> If `SameSite=None` is set, the cookie Secure attribute must also be set (or the cookie will be blocked)
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#none

### References

fixes #569 

### Testing

### Checklist

### DIFF
--- a/src/auth0-session/utils/cookies.ts
+++ b/src/auth0-session/utils/cookies.ts
@@ -22,5 +22,17 @@ export const set = (res: ServerResponse, name: string, value: string, options: C
 };
 
 export const clear = (res: ServerResponse, name: string, options: CookieSerializeOptions = {}): void => {
-  set(res, name, '', { ...options, maxAge: 0 });
+  const { domain, path, secure, sameSite } = options;
+  const clearOptions: CookieSerializeOptions = {
+    domain,
+    path,
+    maxAge: 0
+  };
+  // If SameSite=None is set, the cookie Secure attribute must also be set (or the cookie will be blocked)
+  // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#none
+  if (sameSite === 'none') {
+    clearOptions.secure = secure;
+  }
+
+  set(res, name, '', clearOptions);
 };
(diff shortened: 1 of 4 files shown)


src/auth0-session/utils/cookies.ts
```diff
@@ -22,5 +22,17 @@ export const set = (res: ServerResponse, name: string, value: string, options: C
 };
 
 export const clear = (res: ServerResponse, name: string, options: CookieSerializeOptions = {}): void => {
-  set(res, name, '', { ...options, maxAge: 0 });
+  const { domain, path, secure, sameSite } = options;
```
Widcket, line 25: sameSite is not being passed to `set`, because it's never added to `clearOptions`.

Widcket: Ooops didn't see you had already closed the PR.

adamjmcgrath: That's intentional - you only need to pass `path` and `domain` (and sometime `secure`) when clearing the cookie, I'm just closing this because I can't reproduce #569 on https://localhost and I'm doing a bit more investigation

adamjmcgrath:
Update: reproduced #569 in an iframe and verified the fix in https://codepen.io/adamjmcgrath/pen/bGoKBMW

Widcket (approved):

PR #571, opened 2022-01-06
Honor configured sameSite in transient cookies so you can login to iframe using 'none'

### LINKED ISSUE #541: Checks.state argument is missing when trying to sign in inside an IFrame
### Description

> I am facing the error `Checks.state argument is missing` when i try sign in inside an IFrame. I checked some similar errors where env variables were misconfigured, but i think this is not my case since the authentication flow works fine in outside IFrame.

> The problem happens in the first login attempt, if the user already is logged the redirect works as expected.

### Reproduction

> - Create an application and install @auth0/nextjs-auth0
> - Setup environment variables 
> - Instantiate application inside an IFrame
> - Sign in

### Environment

> Please provide the following:

- **Version of this library used:** 1.5.0
- **Next JS version:** 11.1.0

adamjmcgrath:
### Description

Honor the `SameSite` configuration (which defaults to 'lax') when setting the transient nonce/state cookies.

This will allow users to login to iframes over https when using the [AUTH0_COOKIE_SAME_SITE](https://auth0.github.io/nextjs-auth0/interfaces/config.cookieconfig.html#samesite) config set to `'none'`

### References

Same as https://github.com/auth0/express-openid-connect/pull/188
fixes #541 

### Testing

Tested by logging in to auth0 from an iframe in https://codepen.io/adamjmcgrath/pen/bGoKBMW

### Checklist

### DIFF
--- a/src/auth0-session/handlers/login.ts
+++ b/src/auth0-session/handlers/login.ts
@@ -40,7 +40,7 @@ export default function loginHandlerFactory(
     };
 
     const transientOpts: StoreOptions = {
-      sameSite: opts.authorizationParams.response_mode === 'form_post' ? 'none' : 'lax'
+      sameSite: opts.authorizationParams.response_mode === 'form_post' ? 'none' : config.session.cookie.sameSite
     };
 
     const stateValue = await opts.getLoginState(req as any, opts);
--- a/tests/auth0-session/handlers/login.test.ts
+++ b/tests/auth0-session/handlers/login.test.ts
@@ -1,7 +1,7 @@
 import { parse } from 'url';
 import { CookieJar } from 'tough-cookie';
 import { setup, teardown } from '../fixtures/server';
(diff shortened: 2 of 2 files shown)


Widcket (approved):

richardscarrott:
I'm guessing this doesn't solve the issue for browsers which disable now 3rd party cookies, e.g. Chrome in incognito and Safari 🤔?

adamjmcgrath:
@richardscarrott - for login to work with Safari's ITP (or Chrome Incognito) you'd need to use [Custom Domains](https://auth0.com/docs/customize/custom-domains) (In these cases the default `Lax` SameSite would suffice)

richardscarrott:
@adamjmcgrath thanks for the reply, we're actually already using custom domains, however we want users to be able to login within our Shopify app, whereby our site is rendered in an iframe within the Shopify admin site so our domain is still considered 3rd party.

I don't suppose you know if Auth0 has any cookie-less solutions to login?

PR #581, opened 2022-01-22
Fix silent error in token refresh 

kanalo-shrek:
This fixes the Scenario where if `refresh` is passed as `true` and no refresh token is present there is a silent failure

If `AccessTokenRequest` is requested with `refresh: true` and no refresh_token is present in the session, there is no error thrown esp if there is an existing session, this causes a silent error.

### DIFF
--- a/src/session/get-access-token.ts
+++ b/src/session/get-access-token.ts
@@ -129,23 +129,24 @@ export default function accessTokenFactory(
       }
     }
 
-    // Check if the token has expired.
-    // There is an edge case where we might have some clock skew where our code assumes the token is still valid.
-    // Adding a skew of 1 minute to compensate.
-    if (!session.refreshToken && session.accessTokenExpiresAt * 1000 - 60000 < Date.now()) {
-      throw new AccessTokenError(
-        'access_token_expired',
-        'The access token expired and a refresh token is not available. The user will need to sign in again.'
-      );
-    }
 
     // Check if the token has expired.
     // There is an edge case where we might have some clock skew where our code assumes the token is still valid.
     // Adding a skew of 1 minute to compensate.
     if (
-      (session.refreshToken && session.accessTokenExpiresAt * 1000 - 60000 < Date.now()) ||
-      (session.refreshToken && accessTokenRequest && accessTokenRequest.refresh)
+      (session.accessTokenExpiresAt * 1000 - 60000 < Date.now()) ||
+      (accessTokenRequest && accessTokenRequest.refresh)
     ) {
+      // Check if the token has expired.
+      // There is an edge case where we might have some clock skew where our code assumes the token is still valid.
+      // Adding a skew of 1 minute to compensate.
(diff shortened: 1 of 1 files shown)


src/session/get-access-token.ts
```diff
@@ -129,23 +129,24 @@ export default function accessTokenFactory(
       }
     }
 
-    // Check if the token has expired.
-    // There is an edge case where we might have some clock skew where our code assumes the token is still valid.
-    // Adding a skew of 1 minute to compensate.
-    if (!session.refreshToken && session.accessTokenExpiresAt * 1000 - 60000 < Date.now()) {
-      throw new AccessTokenError(
```
adamjmcgrath, line 136: Hi @kanalo-shrek - can you [fix the case](https://app.circleci.com/pipelines/github/auth0/nextjs-auth0/636/workflows/1320a98c-f2fa-4d55-8e7c-6aac69533ab0/jobs/696/parallel-runs/0/steps/0-107) where `The access token expired and a refresh token is not available`

and also add a test case for the scenario that you've introduced

src/session/get-access-token.ts
```diff
@@ -129,23 +129,24 @@ export default function accessTokenFactory(
...
     if (
-      (session.refreshToken && session.accessTokenExpiresAt * 1000 - 60000 < Date.now()) ||
-      (session.refreshToken && accessTokenRequest && accessTokenRequest.refresh)
+      (session.accessTokenExpiresAt * 1000 - 60000 < Date.now()) ||
+      (accessTokenRequest && accessTokenRequest.refresh)
     ) {
+      // Check if the token has expired.
+      // There is an edge case where we might have some clock skew where our code assumes the token is still valid.
+      // Adding a skew of 1 minute to compensate.
+      if (!session.refreshToken) {
```
adamjmcgrath, line 143: Also, for readability, rather than nesting these if branches, can you create a top level one, along the lines of 

```js
if (!session.refreshToken && accessTokenRequest?.refresh) {
  throw new AccessTokenError(...)
}
```

src/session/get-access-token.ts
kanalo-shrek: Ah I was relying on both the cases, but your comment on detecting the errors early makes sense, just for the sake of readability though do you feel like the SDK should declare variables as `const hasAccessTokenExpired = ...` etc, instead of repeating the same logic everywhere?

adamjmcgrath: Yes, thats true, although you don't need to touch that part of the code for your change (just add a couple of lines for the `!refreshToken && refresh` branch and a single test case for it)

adamjmcgrath:
Hi @kanalo-shrek - let me know if you'd like to continue this review, if not - we'll add something to our backlog to fix

kanalo-shrek:
Hey @adamjmcgrath haven't had a chance to update your comments, might take this out this week-ish.

Issue #597, opened 2022-02-18
Unable to use persisted layouts following the official Next documentation


HMilbradt:
### Description

Following the [documentation ](https://nextjs.org/docs/basic-features/layouts#with-typescript)for adding persistent layouts to Next, I end up with the following Typescript error when using authenticated pages.

`Property 'getLayout' does not exist on type 'FC<WithPageAuthRequiredProps>'`

I realize that this is sort of a hack by Next themselves to get these layouts to work, but it would be really handy if the `withPageAuthRequired` function would support this.  It's a recommended pattern to enable layout persistence between pages.

### Reproduction

The following is enough to reproduce the error.  Note that the `getLayout` function here is taken almost exactly from the Next documentation, just modified to work with Auth0.

Adding a `// @ts-ignore` resolves the error message and allows the layout to work as expected.  

```
import type { ReactElement } from 'react'
import { withPageAuthRequired } from '@auth0/nextjs-auth0'

function Page(): ReactElement {
  return <p>Hello</p>
}

const AuthWrapper = withPageAuthRequired(Page)

AuthWrapper.getLayout = function getLayout(page: ReactElement) {
  return <div>{page}</div>
}

export default AuthWrapper
```

### Environment

- **Version of this library used: 1.7.0**
- **Next: 12.0.10**

adamjmcgrath:
Hi @HMilbradt - thanks for raising this

`getLayout` doesn't exist on a React component, so this should fail regardless of whether you are using `withPageAuthRequired` or not.

Have a look at https://dev.to/ofilipowicz/next-js-per-page-layouts-and-typescript-lh5 for a suggested fix for this

HMilbradt:
Thanks for the link @adamjmcgrath!  To be perfectly honest, the reason I raised this issue is that React and Next _do_ support this.  The example from the docs I linked works out of the box.

It would be really nice to not need to extend types to get this functionality back.   It should also be supported out of the box IMO.

HMilbradt:
@ccccooooddddyyyy. and for anyone else stumbling on this:

https://stackoverflow.com/a/71433194/8727490

I had mistakenly wrapped the `page` with `withPageAuthRequired`, but it turns out you're actually supposed to wrap the `getServerSideProps` with it.  Oddly this "worked", but looking back there was definitely a delayed page load as a result of this error, and obviously the typings I noted earlier did not work.

Kerumen:
[...]

PR #620, opened 2022-03-18
Ensure `withPageAuthRequired` is hoisting non React static properties

Kerumen:
### Description

Fixes https://github.com/auth0/nextjs-auth0/issues/597

`withPageAuthRequired` is an HOC and as every good HOC-citizen, it should hoist the non React static properties. This way we can use, for instance, the [`getLayout` trick](https://nextjs.org/docs/basic-features/layouts#per-page-layouts) to display a layout component in Next.js.

### References

React docs: https://reactjs.org/docs/higher-order-components.html#static-methods-must-be-copied-over
Material UI `withStyles` HOC: https://github.com/mui/material-ui/blob/f2d6337305b75df97d49e58f19288429b7f767e5/packages/mui-styles/src/withStyles/withStyles.js#L88

### Testing

### Checklist


src/frontend/with-page-auth-required.tsx
```diff
@@ -81,7 +82,7 @@ export type WithPageAuthRequired = <P extends WithPageAuthRequiredProps>(
  * @ignore
  */
 const withPageAuthRequired: WithPageAuthRequired = (Component, options = {}) => {
-  return function withPageAuthRequired(props): JSX.Element {
+  function WithPageAuthRequired(props: WithPageAuthRequiredProps): JSX.Element {
```
Kerumen, line 85: I think the correct type would be `props: P` but I don't know how to forward the generic from the type `WithPageAuthRequired` without duplicating the types. 🤔

adamjmcgrath:
Hi @Kerumen - thanks for your PR

Unfortunately, I don't want to take on a dependency to do this - but rather encourage you to add it in your own app if you want to add this behaviour to your components

PR #554, opened 2021-12-08
Improve types in server-side withPageAuthRequired

misoton665:
By submitting a PR to this repository, you agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md). Please see the [contributing guidelines](https://github.com/auth0/open-source-template/blob/master/CONTRIBUTING.md) for how to create and submit a high-quality PR for this repo.

### Description

This PR is about server-side withPageAuthRequired.
Because `withPageAuthRequired` does not accept Query as a type parameter, it's returned GetServerSideProps with a default Query type of ParsedUrlQuery.

```ts
withPageAuthRequired({
  getServerSideProps,
  returnTo: "/foo/bar"
});
```
The type of this will be `GetServerSideProps<P, ParsedUrlQuery>` no matter what type of getServerSideProps is passed.

In this PR, withPageAuthRequired has been changed so that Query can also be specified as a type parameter.

### Testing
Since this is a modification of the type definition, no tests have been added.

- [ ] This change adds test coverage for new/changed/fixed functionality

### Checklist

- [ ] I have added documentation for new/changed functionality in this PR or in auth0.com/docs
- [ ] All active GitHub checks for tests, formatting, and security are passing
- [ ] The correct base branch is being used, if not `main`

PR #632, opened 2022-04-01
Add default to PageRoute type parameter

grantbdev:
### Description

Add default to PageRoute type parameter to match the default for the WithPageAuthRequiredOptions type and make it easier for the generic type to be used.

The generic type arguments for URL queries were added here: https://github.com/auth0/nextjs-auth0/pull/554 and these changes were released in v1.7.0.

For codebases that were already importing the PageRoute type, the introduction of the new required type argument failed type-checking with this error:
> Generic type 'PageRoute' requires 2 type argument(s). ts(2314)

It is easily fixed by providing a value for the new type argument, but by adding the default into this package it makes it easier for users of the package since they do not have to update their codebases.

### Testing

Since this is a modification of the type definition, no tests have been added.

### Checklist

### DIFF
--- a/src/helpers/with-page-auth-required.ts
+++ b/src/helpers/with-page-auth-required.ts
@@ -34,7 +34,7 @@ export type GetServerSidePropsResultWithSession<P = any> = GetServerSidePropsRes
  *
  * @category Server
  */
-export type PageRoute<P, Q extends ParsedUrlQuery> = (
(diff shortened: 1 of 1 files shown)


adamjmcgrath (approved):

adamjmcgrath:
Thanks @grantbdev!

PR #664, opened 2022-05-13
[SDK-3332] Constrain session lifecycle to `withPageAuthrequired` to avoid Next warning

### LINKED ISSUE #524: warn  - You should not access 'res' after getServerSideProps resolves.
### Description

next12 warns

> warn  - You should not access 'res' after getServerSideProps resolves.
> Read more: https://nextjs.org/docs/messages/gssp-no-mutating-res

### Reproduction
Using withPageAuthRequired or getSession is enough to get the warning

### Environment

nextjs@12

adamjmcgrath:
### Description

Next.js >=12 have started to issue a warning (`You should not access 'res' after getServerSideProps resolves.`), and in some cases an error, when you access the response object after `getServerSideProps` has run.

Because this SDK provides a rolling session by default, it writes to the response at the end of every request. It does this at the end of the request using [on-headers](https://github.com/jshttp/on-headers), which causes the warning because `on-headers` executes after `getServerSideProps` resolves.

We can fix this in `withAuthenticationRequired` by not using `on-headers` and writing to the headers just before `withAuthenticationRequired` completes, eg.

```js
async function withAuthenticationRequired({ getServerSideProps }) {
  session.get(disableOnHeaders=true) // read cookie and disable on-headers
  //  ... existing code ...
  session.save() // write cookie (because on-headers is disabled)
}
``` 

If you want to use `getSession` (and `getAccessToken`) without requiring auth on your route, you'll still need to use the `withAuthenticationRequired` wrapper, but you can do this now with the `authRequired: false` option. eg

```js
export const getServerSideProps = withAuthenticationRequired({
  authRequired: false,
  async getServerSideProps(ctx) {
    const session = getSession(ctx.req, ctx.res);
    if (session) {
      // user is logged in
    } else {
      // user is not logged in
    }
  }
});
```

I also fixed an issue in `withAuthenticationRequired` when `props` is a Promise, eg

```js
function getServerSideProps() {
  return { props: Promise.resolve({}) }
}
```

### References

https://nextjs.org/docs/messages/gssp-no-mutating-res
fixes #524 

### Testing

Run the basic example app and notice that there is no warning when visiting `/protected`

### Checklist

### DIFF
--- a/src/helpers/with-page-auth-required.ts
+++ b/src/helpers/with-page-auth-required.ts
@@ -1,5 +1,5 @@
 import { GetServerSideProps, GetServerSidePropsContext, GetServerSidePropsResult } from 'next';
-import { Claims, GetSession } from '../session';
+import { Claims, SessionCache } from '../session';
 import { assertCtx } from '../utils/assert';
 import React, { ComponentType } from 'react';
 import {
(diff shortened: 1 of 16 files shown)


src/helpers/with-page-auth-required.ts
```diff
@@ -62,11 +62,38 @@ export type PageRoute<P, Q extends ParsedUrlQuery = ParsedUrlQuery> = (
...
+ *   }
+ * });
+ * ```
+ *
  * @category Server
  */
 export type WithPageAuthRequiredOptions<P = any, Q extends ParsedUrlQuery = ParsedUrlQuery> = {
   getServerSideProps?: GetServerSideProps<P, Q>;
   returnTo?: string;
+  authRequired?: boolean;
```
Widcket, line 96: Is an `authRequired` option on `withAuthRequired` the best way to go? It kind of defeats the entire purpose of  `withAuthRequired`. Should we go with a different wrapper instead?

src/index.ts
```diff
@@ -50,17 +50,17 @@ import { InitAuth0, SignInWithAuth0 } from './instance';
 import version from './version';
 import { getConfig, getLoginUrl, ConfigParameters } from './config';
 
-let instance: SignInWithAuth0;
+let instance: SignInWithAuth0 & { sessionCache: SessionCache };
```
Widcket, line 53: Can we define a type alias for this?

src/helpers/with-page-auth-required.ts
adamjmcgrath: I thought about that (eg `withPageAuth` or `withSessionInitialised` etc) - but decided this would be the easiest to grok

src/index.ts
adamjmcgrath: It's only used internally in this file, creating an alias seems overkill

Widcket (dismissed):

src/helpers/with-page-auth-required.ts
Widcket: I'm not sure that should determine the design of this feature (documenting it well along with a good name can make it easy to grok as well). I'll leave it up to you.

adamjmcgrath: I went for `getServerSidePropsWrapper` in e523099 (wait until after your day off to take a look :P)

Widcket (approved):

Issue #668, opened 2022-05-18
Token exchanges provide no ip forwarding headers, triggering issues at Auth0 under scale


ShawnFumo:
Our next.js application is using a serverless lambda to proxy API calls, to add the access token to the request (so the browser never has access to the unencrypted access token). This has been working, but as volume increased, we started to get reports of people being logged out before the inactivity timeout. We looked in the logs and saw intermittent "Failed Exchange of Refresh Token for Access Token". After talking to Auth0 support, they noted that there was a lot of different user requests coming from the same IPs, and that may be the cause.

You can see from the attached image that the user's real IP gets to Auth0 for account creation and login. Then that person's exchange from auth code to access token happens from an Amazon IP. Then a different user's refresh token to access token exchange happens, coming from the same IP. Then that same user has another exchange soon after, but from a different IP. I believe lambdas don't have static IP addresses, so that makes sense.

![nextjs ips](https://user-images.githubusercontent.com/1048925/169157906-f97a3217-6373-4079-b9d7-c7c5e7404802.PNG)

Support suggested adding the `auth0-forwarded-for` header and I'm currently trying to see how to hook into this library to inject that header. But it seems like this would be an issue for anyone using the library at a decent scale and that the header should be added by default? This is the code we're using, so you can see it is the default functionality of the library:

```
import { withApiAuthRequired, getAccessToken } from '@auth0/nextjs-auth0';

export default withApiAuthRequired(async function shows(req, res) {
    let accessToken = ""
    try {
      const {accessToken: at} = await getAccessToken(req, res);
      accessToken = at
    } catch (error) {
      res.status(401).send({error: error})
      return
    }

  // Omitting the code below which adds the token to the graphql request
}
```

Here's an example of a failing request with most of it redacted, but showing our library and node versions:
```
{
  "content": {
    "message": "Failed Exchange of Refresh Token for Access Token",
    "attributes": {
      "data": {
        "auth0_client": {
          "version": "1.7.0",
          "name": "nextjs-auth0",
          "env": {
            "node": "v14.19.1"
          }
        },
      },
      "evt": {
        "name": "failed_exchange_refresh_token_for_access_token"
      },
      "http": {
        "useragent": "nextjs-auth0/1.7.0"
      },
      "service": "auth0"
    }
   }
}
```

We are not using refresh token rotation. The access token expiration was set at 2 minutes and refresh token inactivity timeout set to 12 minutes. We recently tried increasing both values but are still seeing the errors.

[...]

Issue #616, opened 2022-03-15
getSession triggers sending data to the client


dariosky:
I'm trying to create a thin proxy to use the session to get the apiToken and proxy to an internal API.
I'm using next-http-proxy-middleware to retrieve the session an add the Authorization header.
The issue is that whenever I call getSession (or getAccessToken) something is sent to the client, and setting a proxy-request header I get:

```
Cannot set headers after they are sent to the client
```

I also tried mocking the res parameter sent to getSession but without luck - how can I get the session readonly without interfering with the response?
I was also trying to read the cookie manually and using `jose` to decrypt it - but it seems an hack.

PS. I'm using Next.JS 12.1.0

Here's the code of my proxy

```js
/* pages/api/proxy/[...all].js */

import { getSession, withApiAuthRequired } from "@auth0/nextjs-auth0"
import httpProxyMiddleware from "next-http-proxy-middleware"
import { API_BASE } from "/utils/config"

export const config = {
  api: {
    bodyParser: false,
    externalResolver: true,
  },
}

const handleProxyInit = proxy => {
  proxy.once("proxyReq", async (proxyReq, req, res) => {
    // console.debug(`New request ${req.url}`)
    // console.log(req.cookies.appSession)
    let session = null
    session = await getSession(req, res)
    if (session?.user) {
      console.debug(`a request from ${session.user.name}-${session.user.sub}`)
    }
    const token = session?.accessToken
    if (token) {
      proxyReq.setHeader("Authorization", `Bearer ${token}`)
    }
  })
}

export default withApiAuthRequired(async (req, res) =>
  httpProxyMiddleware(req, res, {
    target: API_BASE,
    pathRewrite: [{ patternStr: "^/api/proxy", replaceStr: "" }],
    onProxyInit: handleProxyInit,
  }),
)
```

adamjmcgrath:
Hi @dariosky - this is currently by design. The session by default is a rolling session, so a new cookie expiry must be written to the response every time you get the session. 

We do have plans to change this behaviour to optionally only save on demand when you disable the rolling session option or when the session changes, but I can't give you a timeline on when this might happen.

Closing, feel free to ping me if you have any other questions

dariosky:
Thank you for the answer @adamjmcgrath 

If it can help - I ended up decoding the `req.cookies.appSession` manually (so to avoid the getSession).

```js
import hkdf from "futoin-hkdf"
import { JWK, JWE } from "jose"
[...]

Issue #425, opened 2021-06-23
Is it possible to not send set-cookie header each time? 


divyenduz:
### Description

`getSession` calls sessionFactory(sessionCache) which calls cookie utils where setting `Set-Cookie` header happens for each request (probably because of rotation policy?)

https://github.com/auth0/nextjs-auth0/blob/main/src/auth0-session/utils/cookies.ts#L16

Is it possible to disable this each time we try to get the current session? Only login response should sent the `set-cookie` header and `getSession` helper should just return the session. 

### Context

I want to do this because I want to cache some user specific responses but GraphCDN doesn't cache when origin sends `Set-Cookie` https://graphcdn.io/docs/how-to/debug-cache-misses

### Environment

> Please provide the following:

- **Version of this library used:**: 1.4.0
- **Version of the platform or framework used, if applicable:** next.js 11

adamjmcgrath:
Hi @divyenduz - thanks for raising this.

By default this wouldn't be possible because we use a rolling session (eg [log out after 24hrs of innactivity](https://auth0.github.io/nextjs-auth0/interfaces/config.sessionconfig.html#rolling)), which means every time you visit the site the session needs to be updated to reset the innactivity clock.

divyenduz:
Thank you for your answer, is there a way to configure this? If no, is that a feature that you would consider? (happy to send a PR for it)

adamjmcgrath:
You can disable the rolling session feature, but we currently still [store a new 'updated at' value](https://github.com/auth0/nextjs-auth0/blob/main/src/auth0-session/cookie-store.ts#L178) for every request and I don't see us changing that.

However there's nothing stopping you adding your own [`onHeaders` listener ](https://github.com/auth0/nextjs-auth0/blob/main/src/session/cache.ts#L22) to the response and removing the `Set-Cookie` if your confident that you don't need the updated session.

Closing this - feel free to ping me if you'd like to reopen to discuss more

divyenduz:
For folks who reach here, this works and successfully removed the `Set-Cookie` header. I am only doing this in the GraphQL part of my API and it resolves my use case. I didn't have to set `rolling` to false. 

![image](https://user-images.githubusercontent.com/746482/123510771-9ab8f200-d67d-11eb-9b94-49e48ed07e69.png)

PR #787, opened 2022-08-24
[SDK-3557] Refactor session lifecycle

adamjmcgrath:

### 📋 Changes

Changed the cookie writing behaviour from lazy (at the end of the request) to eager (when it's first accessed and the session is rolling and every-time it's updated).

This has enabled us to:
- remove `on-headers` dependency (which doesn't play well with Next.js server and doesn't work at all with Next.js middleware)
- remove wrappers (`getServerSideProps` wrapper and removed the need for an API route wrapper)
- removed the magic about implicit session updates by adding an explicit `updateUser` method
- allowed users to prevent the cookie from being written on every request when the session is absolute (not rolling) - which can be an issue for specific implementations and platforms

At a cost of potentially higher overhead when updating the session multiple times in the request, because multiple calls to `updateUser` will cause multiple cookie serialisation (encryption) attempts (but multiple reads will stay the same).

In order to do this we've needed to:
- Make a stateful Cookie class to manage/aggregate writes to the `set-cookie` header
- Add an `updateUser` method
- Update the session cache to eagerly write to the cookie, rather than use `on-headers`

### 📎 References

see #668 #425 #616 #524

### 🎯 Testing

https://github.com/auth0/nextjs-auth0/blob/54cd48f1e1cb78eb37b700f39f4da826a5683c1a/tests/session/update-user.test.ts


package.json
```diff
@@ -93,6 +93,7 @@
     "next": "^12.1.6",
     "nock": "^13.0.5",
     "oidc-provider": "^7.6.0",
+    "on-headers": "^1.0.2",
```
Widcket, line 96: Shouldn't this be removed?

Widcket (changes requested):
Looks good! Nice solution. Just left a question regarding the `on-headers` dependency, which –according to the PR description– was removed, yet is still in the `package.json`.

package.json
adamjmcgrath: I'm still using it in some test fixtures, so I moved it to `devDependencies`

Widcket (approved):

PR #798, opened 2022-08-31
[SDK-3570] Upgrade to Jose@4

adamjmcgrath:

### 📋 Changes

Have upgraded jose from 2 -> 4 - this is a requirement for running the SDK on [Vercel Edge Middleware](https://vercel.com/docs/concepts/functions/edge-middleware) (and other edge runtimes). cookie read/save is now async so there are a bunch of changes to make accessing/updating the session async.

Main changes are:
- Updating the [cookie-store.ts](https://github.com/auth0/nextjs-auth0/compare/vNext...jose4?expand=1#diff-bcef69150bdfde777d944c2f77a397df1d14db5775bd7e3963fc534e1f939a04)
- Updating the [transient-store.ts](https://github.com/auth0/nextjs-auth0/compare/vNext...jose4?expand=1#diff-1cc5cd9cbee1a19dec99b13acdc078b207a24ad20a7d522e42456c856ac334a3)
- Updating [utils/hkdf.ts](https://github.com/auth0/nextjs-auth0/compare/vNext...jose4?expand=1#diff-699b38e75430506a5637d969c4136aabcf9a38379644d8aa14de5d3022e640de)

Then adding a bunch of `async` / `await`s

### 📎 References

https://github.com/panva/jose/releases/tag/v3.0.0
https://github.com/panva/jose/releases/tag/v4.0.0


src/auth0-session/cookie-store.ts
```diff
@@ -49,20 +38,23 @@ export default class CookieStore {
...
 
-  private decrypt(jwe: string): JWE.completeDecrypt {
-    return JWE.decrypt(jwe, this.keystore, {
-      complete: true,
-      contentEncryptionAlgorithms: [enc],
-      keyManagementAlgorithms: [alg]
-    });
+  private async decrypt(jwe: string): Promise<jose.CompactDecryptResult> {
+    const keys = await this.getKeys();
+    return pAny(keys.map((key) => jose.compactDecrypt(jwe, key))) as Promise<jose.CompactDecryptResult>;
```
panva, line 57: I believe it would be more efficient to iterate over the individual keys until a success occurs than to fire away all decrypt operations at once. This can be accompanied by a doc recommendation to put the most recent key at the front of the array of keys (when array is used).

Also, any reason not to use `jwtDecrypt` instead of `compactDecrypt`?

src/auth0-session/transient-store.ts
```diff
@@ -10,70 +11,52 @@ export interface StoreOptions {
...
+    await pAny(
+      keys.map((key) =>
+        jose.flattenedVerify(flattenedJWS, key, {
+          algorithms: ['HS256'],
+          crit: {
+            b64: false
+          }
+        })
+      )
+    );
```
panva, line 34: same comment about iterate vs firing all at once.

src/auth0-session/utils/p-any.ts
```diff
@@ -0,0 +1,7 @@
+const flip = (promise: Promise<any>) => new Promise((a, b) => promise.then(b, a));
+
+// Lightweight Promise.any-like implementation
+// Promise.all returns the first rejected promise or all resolved promises
+// Promise.any returns the first resolved promise or all rejected promises
+// If we flip all the promises of Promise.all then flip back the result, we get the behaviour of Promise.any
+export default async (promises: Promise<any>[]) => flip(Promise.all(promises.map(flip)));
```
panva, line 7: 🪄

src/auth0-session/cookie-store.ts
adamjmcgrath: 👍 005c8e4

src/auth0-session/transient-store.ts
adamjmcgrath: 👍 005c8e4

src/auth0-session/transient-store.ts
```diff
@@ -10,70 +10,49 @@ export interface StoreOptions {
...
+    payload: `${k}=${value}`,
+    signature
+  };
+  for (let key of keys) {
+    try {
+      await jose.flattenedVerify(flattenedJWS, key, {
+        algorithms: ['HS256'],
+        crit: {
+          b64: false
+        }
```
panva, line 29: I believe passing this option to the verify API is not necessary.

https://github.com/panva/jose/blob/v4.9.2/docs/interfaces/types.VerifyOptions.md

> The JWS extension Header Parameter `b64` is always recognized and processed properly. No other registered Header Parameters that need this kind of default built-in treatment are currently available.

adamjmcgrath: 👍 5aa49c3

package.json
```diff
@@ -107,13 +107,12 @@
...
   "dependencies": {
-    "base64url": "^3.0.1",
+    "@panva/hkdf": "^1.0.2",
     "cookie": "^0.5.0",
     "debug": "^4.3.4",
-    "futoin-hkdf": "^1.5.0",
     "http-errors": "^1.8.1",
     "joi": "^17.6.0",
-    "jose": "^2.0.5",
+    "jose": "^4.9.0",
```
panva, line 115:
```suggestion
    "jose": "^4.9.2",
```

panva (approved):

src/session/update-user.ts
```diff
@@ -26,18 +26,18 @@ export type UpdateUser = (
...
-    sessionCache.init(req, res, false);
-    const session = sessionCache.get(req, res);
+  return async (req, res, user) => {
+    await sessionCache.init(req, res, false);
+    const session = await sessionCache.get(req, res);
     if (!session || !user) {
       return;
     }
-    sessionCache.set(req, res, { ...session, user });
+    await sessionCache.set(req, res, { ...session, user });
```
Widcket, line 41: Is this last `await` necessary, or could it be just `return sessionCache.set(req, res, { ...session, user });`?

src/session/cache.ts
```diff
@@ -16,52 +16,52 @@ export default class SessionCache implements ISessionCache {
...
     return session?.idToken;
   }
 
-  set(req: NextApiOrPageRequest, res: NextApiOrPageResponse, session: Session | null): void {
-    this.init(req, res, false);
+  async set(req: NextApiOrPageRequest, res: NextApiOrPageResponse, session: Session | null): Promise<void> {
+    await this.init(req, res, false);
     this.cache.set(req, session);
-    this.save(req, res);
+    await this.save(req, res);
```
Widcket, line 60: Is this last `await` necessary, or could it be just `return this.save(req, res);`?

src/session/cache.ts
```diff
@@ -16,52 +16,52 @@ export default class SessionCache implements ISessionCache {
...
   }
 
-  delete(req: NextApiOrPageRequest, res: NextApiOrPageResponse): void {
-    this.init(req, res, false);
+  async delete(req: NextApiOrPageRequest, res: NextApiOrPageResponse): Promise<void> {
+    await this.init(req, res, false);
     this.cache.set(req, null);
-    this.save(req, res);
+    await this.save(req, res);
```
Widcket, line 42: Same as before.

src/session/cache.ts
```diff
@@ -16,52 +16,52 @@ export default class SessionCache implements ISessionCache {
...
-    this.cookieStore.save(req, res, this.cache.get(req), this.iatCache.get(req));
+  async save(req: NextApiOrPageRequest, res: NextApiOrPageResponse): Promise<void> {
+    await this.cookieStore.save(req, res, this.cache.get(req), this.iatCache.get(req));
   }
 
-  create(req: NextApiOrPageRequest, res: NextApiOrPageResponse, session: Session): void {
+  async create(req: NextApiOrPageRequest, res: NextApiOrPageResponse, session: Session): Promise<void> {
     this.cache.set(req, session);
-    this.save(req, res);
+    await this.save(req, res);
```
Widcket, line 36: Same as before.

src/session/cache.ts
```diff
@@ -16,52 +16,52 @@ export default class SessionCache implements ISessionCache {
...
-        this.save(req, res);
+        await this.save(req, res);
       }
     }
   }
 
-  save(req: NextApiOrPageRequest, res: NextApiOrPageResponse): void {
-    this.cookieStore.save(req, res, this.cache.get(req), this.iatCache.get(req));
+  async save(req: NextApiOrPageRequest, res: NextApiOrPageResponse): Promise<void> {
+    await this.cookieStore.save(req, res, this.cache.get(req), this.iatCache.get(req));
```
Widcket, line 31: Same as before.

Widcket (approved):

src/session/update-user.ts
adamjmcgrath: You could do either. It only really makes a difference when you're using try/catch statements

```js
try {
  return promise;
} catch () {}

// is very different from
try {
  return await promise;
} catch () {}
```

Returning from this would change the function return type from `Promise<void>` to Promise<undefined>`, I think for a function that shouldn't return anything `Promise<void>` is probably more suitable

Issue #525, opened 2021-10-26
Edge function support


andrewscofield:
### Describe the problem you'd like to have solved

Authenticate users at the edge (https://vercel.com/features/edge-functions) 

### Describe the ideal solution

Middleware function for auth0 to get user object at page render.

aiji42:
I'm looking forward to working with the Edge function!

Incidentally, I bypassed the cookie from the middleware to `/api/auth/me` using `fetch` to check the authentication status on the middleware.
**Requesting the api route from the Edge function is not the right way**, as it hurts performance. I am using this approach as an interim one until formal support is in place.

https://github.com/aiji42/next-fortress/blob/main/src/auth0.ts
```ts
const verifyAuth0Session = async (
  req: NextRequest,
  apiEndpoint: string
): Promise<boolean> => {
  const res = await fetch(req.nextUrl.origin + '/api/auth/me', {
    headers: { cookie: req.headers.get('cookie') ?? '' }
  })

  return res.ok
}
```
(I am developing a plugin that uses Next.js middleware to check the authentication status of IDSaaS such as Auth0 and determine whether content is accessible or not. Hope this is helpful to someone. https://github.com/aiji42/next-fortress#control-by-auth0)

theMadness:
Would recommend checking https://github.com/vercel/examples/tree/main/edge-functions, particularly https://github.com/vercel/examples/tree/main/edge-functions/jwt-authentication

aiji42:
Thanks for the recommendation. Yes, I did read it. However, the JWT stored in the cookie handled by nextjs-auth0 does not have a payload and there is only a small amount of information that can be retrieved from the header.
```
{
  "alg": "RS256",
  "enc": "A256GCM",
  "iat": 1636505290,
  "uat": 1636505291,
  "exp": 1636591691
}
```
https://community.auth0.com/t/token-in-appsession-cookie-cannot-be-decoded/60591/3

GMaiolo:
Any news about this?

adamjmcgrath:
Hi @GMaiolo - thanks for checking. No news, we'll update this thread when we do

paulobarcelos:
Hi @adamjmcgrath, could you clarify if this is being currently developed, or if it hasn't started yet? I understand if there aren't any meaningful updates at this point, but it would be nice to have some indication of what is the priority this has internally. Being able to authenticate on the edge would be a game-changer for us, so I am really curious about this :)

[...]

PR #815, opened 2022-09-09
[SDK-3554] Next.js Middlware support

adamjmcgrath:

### 📋 Changes

Added a new `withMiddlwareAuthRequired` helper, to protect routes using middleware.

To protect all your routes:

```js
// middleware.js
import { withMiddlewareAuthRequired } from '@auth0/nextjs-auth0/middleware';

export default withMiddlewareAuthRequired();
```

To protect specific routes:
```js
// middleware.js
import { withMiddlewareAuthRequired } from '@auth0/nextjs-auth0/middleware';

export default withMiddlewareAuthRequired();

export const config = {
  matcher: '/about/:path*',
}
```
For more info see: https://nextjs.org/docs/advanced-features/middleware#matching-paths

To run custom middleware for authenticated users:
```js
// middleware.js
import { withMiddlewareAuthRequired, getSession } from '@auth0/nextjs-auth0/middleware';

export default withMiddlewareAuthRequired(async function middleware(req) {
  const res = NextResponse.next();
  const user = await getSession(req, res);
  res.cookies.set('hl', user.language);
  return res;
});
```

### 📎 References

See https://github.com/auth0/nextjs-auth0/issues/525
https://nextjs.org/docs/advanced-features/middleware

### 🎯 Testing

Run kitchen-sink example app
Visit http://localhost:3000/profile-mw


package.json
```diff
@@ -132,7 +134,12 @@
       "<rootDir>/src/**/*.*",
       "!<rootDir>/src/index.browser.ts",
       "!<rootDir>/src/index.ts",
-      "!<rootDir>/src/handlers/auth.ts"
+      "!<rootDir>/src/middleware.ts",
+      "!<rootDir>/src/instance.ts",
+      "!<rootDir>/src/handlers/auth.ts",
+      "!<rootDir>/src/auth0-session/config.ts",
+      "!<rootDir>/src/auth0-session/index.ts",
+      "!<rootDir>/src/auth0-session/session-cache.ts"
```
adamjmcgrath, line 142: New coverage instrumenter c8 doesn't like files with only TS definitions. (was needed because istanbul instrumented code doesn't run on edge-runtime)

src/handlers/callback.ts
```diff
@@ -108,26 +108,23 @@ export type HandleCallback = (req: NextApiRequest, res: NextApiResponse, options
...
-      `Organization Id (org_id) claim value mismatch in the ID token; ` +
-        `expected "${organization}", found "${session.user.org_id}"`
-    );
-  }
-  if (afterCallback) {
-    return afterCallback(req, res, session, state);
-  }
-  return session;
-};
+const idTokenValidator =
```
adamjmcgrath, line 111: Whitespace changes from `eslint --fix`

src/helpers/with-middleware-auth-required.ts
```diff
@@ -0,0 +1,88 @@
+import { NextMiddleware, NextRequest, NextResponse } from 'next/server';
+import { SessionCache } from '../session';
+
+/**
+ * Protect your pages with Next.js Middleware. eg
```
Widcket, line 5:
```suggestion
 * Protect your pages with Next.js Middleware. For example:
```
The docs guidelines suggest not to use latin abbreviations.

src/helpers/with-middleware-auth-required.ts
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * ```js
+ * // middleware.js
+ * import { withMiddlewareAuthRequired } from '@auth0/nextjs-auth0/middleware';
+ *
+ * export default withMiddlewareAuthRequired();
+ *
+ * export const config = {
+ *   matcher: '/about/:path*',
+ * }
```
Widcket, line 26:
```suggestion
 * };
```

src/auth0-session/session-cache.ts
```diff
@@ -1,10 +1,9 @@
-import { IncomingMessage, ServerResponse } from 'http';
-import { TokenSet } from 'openid-client';
+import type { TokenSet } from 'openid-client';
 
 export interface SessionCache {
-  create(req: IncomingMessage, res: ServerResponse, session: { [key: string]: any }): Promise<void>;
-  delete(req: IncomingMessage, res: ServerResponse): Promise<void>;
-  isAuthenticated(req: IncomingMessage, res: ServerResponse): Promise<boolean>;
-  getIdToken(req: IncomingMessage, res: ServerResponse): Promise<string | undefined>;
+  create(req: unknown, res: unknown, session: { [key: string]: any }): Promise<void>;
```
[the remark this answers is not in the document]
adamjmcgrath: Good point, I can use generics here 👍

Widcket (approved):

package.json
```diff
@@ -11,6 +11,7 @@
   "files": [
     "dist",
     "src",
+    "middleware.js",
```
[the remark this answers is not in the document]
adamjmcgrath: hmm, good point - I didn't realise you had to do that. Will fix

Widcket (approved):

package.json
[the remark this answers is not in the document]
Widcket: You also have to include the typings file here –otherwise it will not be present in the package because it's not in side src.

Widcket (changes requested):
Sorry, just noticed that the typings file also needs to be included in the `files` array in the `package.json` –otherwise it will not be present in the package because it's not inside `src`.

package.json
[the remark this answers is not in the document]
adamjmcgrath: d'oh, yep - good point 👍

Widcket (approved):

Jared-Dev:
Not sure if this is the correct place to put this, but I believe I'm seeing some typing issues with the latest version of Next on the `vNext` branch.

When using a copy/paste example from the Next docs I'm seeing the below.
image
image

adamjmcgrath:
Hi @Jared-Dev - thanks for testing this out!

Looks like you're using `npm link` and the version of next installed in `Sites/packages/nextjs-auth0` is incompatible with the version installed in `Sites/ox-portal/` - if you use `npm pack` to package the dependency (or install from npm when the beta is available - v soon) you will only have one `next` dependency so shouldn't run into this problem.

src/helpers/with-middleware-auth-required.ts
```diff
@@ -0,0 +1,88 @@
...
+      const res = await (middleware && middleware(...args));
+
+      if (res) {
+        const headers = new Headers(res.headers);
+        const cookies = headers.get('set-cookie')?.split(', ') || [];
+        const authCookies = authRes.headers.get('set-cookie')?.split(', ') || [];
+        if (cookies.length || authCookies.length) {
+          headers.set('set-cookie', [...authCookies, ...cookies].join(', '));
+        }
+        return NextResponse.next({ ...res, headers });
```
divmgl, line 82: This is not working as intended for redirects. We'll need a test around this as well. See https://github.com/auth0/nextjs-auth0/issues/874.

```suggestion
        return NextResponse.next({ ...res, headers, status: res.status });
```

PR #826, opened 2022-09-22
Add support for configuring the built-in handlers [SDK-3566]

Widcket:
- [X] All new/changed/fixed functionality is covered by tests (or N/A)
- [X] I have added documentation for all new/changed functionality (or N/A)

### 📋 Changes

Currently it's not possible to dynamically configure the built-in handlers. For example, to pass a `connection` parameter to the login handler, it's necessary to override it. 

#### Before

```js
export default handleAuth({
  async login(req, res) {
    try {
      await handleLogin(req, res, {
        authorizationParams: {
          connection: 'github'
        },
      });
    } catch (error) {
      // ...
    }
  }
});
```

This PR allows to pass option objects to configure the built-in handlers, without having to override them.

#### After

```js
export default handleAuth({
  login: {
    authorizationParams: {
      connection: 'github'
    }
  }
});
```

It's still possible to override the built-in handlers by passing custom handler functions.

### 🎯 Testing

Besides adding unit tests, this change was tested manually in the kitchen sink example app.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,16 @@
+{
+  "version": "1.0.0",
+  "configurations": [
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Jest: current file",
+      //"env": { "NODE_ENV": "test" },
+      "program": "${workspaceFolder}/node_modules/.bin/jest",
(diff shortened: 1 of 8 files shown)


adamjmcgrath:
Closing, as we're going in a different direction with this

PR #838, opened 2022-10-04
[SDK-3567] Allow response customization in afterCallback

adamjmcgrath:
### 📋 Changes

Allow customisation of the response in `afterCallback` to enabled a couple of use cases

1. Terminating the request if a condition in the claim is not met.

```js
const afterCallback = (req, res, session, state) => {
  if (session.user.isAdmin) {
    return session;
  } else {
    res.status(401).end('User is not admin');
  }
}; // Terminates the request with 401 if user is not admin
```

2. Customising the return to url based on the claims

```js
const afterCallback = (req, res, session, state) => {
  if (!session.user.isAdmin) {
    res.setHeader('Location', '/admin');
  }
  return session;
}; // Redirects to `/admin` if user is admin
```

### DIFF
--- a/src/auth0-session/handlers/callback.ts
+++ b/src/auth0-session/handlers/callback.ts
@@ -19,7 +19,12 @@ function getRedirectUri(config: Config): string {
   return urlJoin(config.baseURL, config.routes.callback);
 }
 
-export type AfterCallback = (req: any, res: any, session: any, state?: Record<string, any>) => Promise<any> | any;
+export type AfterCallback = (
+  req: any,
+  res: any,
+  session: any,
+  state?: Record<string, any>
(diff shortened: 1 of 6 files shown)


src/auth0-session/handlers/callback.ts
```diff
@@ -19,7 +19,12 @@ function getRedirectUri(config: Config): string {
   return urlJoin(config.baseURL, config.routes.callback);
 }
 
-export type AfterCallback = (req: any, res: any, session: any, state?: Record<string, any>) => Promise<any> | any;
+export type AfterCallback = (
+  req: any,
```
Widcket, line 23: Any way we can type the parameters?

Widcket (approved):

PR #862, opened 2022-10-14
Shouldn't serve routes for object builtins

adamjmcgrath:
### 📋 Changes

Noticed while testing that we're serving routes for js builtins (like `__proto__`)

`/api/auth/__proto__` should return 404 (currently returning 500)


Widcket (approved):
Good catch!

PR #864, opened 2022-10-17
[ASEC-891] Escape idp error

adamjmcgrath:
### 📋 Changes

Escape the `cause.error` and `cause.error_description` properties in addition to the `message` per security review.

### DIFF
--- a/src/utils/errors.ts
+++ b/src/utils/errors.ts
@@ -48,8 +48,11 @@ export abstract class AuthError extends Error {
   /**
    * The underlying error, if any.
    *
-   * **IMPORTANT** This underlying error is **not** escaped in any way,
-   * so do **not** render it without escaping it first!
+   * **IMPORTANT** When this error is from the Identity Provider ({@Link IdentityProviderError}) it can contain user
+   * input and is only escaped using basic escaping for putting untrusted data directly into the HTML body.
+   *
+   * You should **not** render this error without using a templating engine that will properly escape it for other
+   * HTML contexts first.
    */
   public readonly cause?: Error;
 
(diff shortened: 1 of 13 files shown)


src/utils/errors.ts
```diff
@@ -1,23 +1,12 @@
...
-}
-
 /**
  * @ignore
  */
 export function appendCause(errorMessage: string, cause?: Error): string {
   if (!cause) return errorMessage;
   const separator = errorMessage.endsWith('.') ? '' : '.';
-  return `${errorMessage}${separator} CAUSE: ${htmlSafe(cause.message)}`;
+  return `${errorMessage}${separator} CAUSE: ${cause.message}`;
```
Widcket, line 9: Shouldn't this keep escaping the cause message?
If not, we should amend the documentation in line 51.

src/utils/errors.ts
```diff
@@ -1,23 +1,12 @@
 import { HttpError } from 'http-errors';
```
Widcket, line 2: We should amend the documentation in line 51:

> **IMPORTANT** This underlying error is **not** escaped in any way

adamjmcgrath: Good point [8e4ab04](https://github.com/auth0/nextjs-auth0/pull/864/commits/8e4ab04ca54b750c4cb1ca988855afec6cd9ffcf)

adamjmcgrath:
Couple of things to check, will reopen

adamjmcgrath:
6e7aa2a Also escaping `ApplicationError` (From `RPError`) and basically any error property that comes from a third party.

PR #878, opened 2022-10-21
At error cause to AT error when it's from a failed grant

### LINKED ISSUE #872: [TypeScript Error Interface ] Can't find correct error interface for cast error to this type (getAccessToken)
### Checklist

- [X] The issue can be reproduced in the [sample app](https://github.com/auth0-samples/auth0-nextjs-samples/tree/main/Sample-01) (or N/A).
- [X] I have looked into the [README](https://github.com/auth0/nextjs-auth0#readme) and have not found a suitable solution or answer.
- [X] I have looked into the [examples](https://github.com/auth0/nextjs-auth0/blob/main/EXAMPLES.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://auth0.github.io/nextjs-auth0/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/nextjs-auth0/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com/c/sdks/5) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

When I try to use the `refresh token` to get a new `access token` via `getAccessToken` function my refresh token can be expired. Because that handling such cases I wrap  `getAccessToken` function call with try-catch. 

```
  try {
    const { accessToken } = await getAccessToken(req, res);
    console.log(accessToken)
  } catch (error) {
    console.log(JSON.stringify(error)); // {"error":"invalid_grant","error_description":"Unknown or invalid refresh token."}
  }
```

But the problem is I can't find any interface for cast error to this type. 

I know SDK has `AccessTokenError` but this error structure is not a similar error that I get 

https://auth0.github.io/nextjs-auth0/classes/utils_errors.accesstokenerror.html

### Reproduction

Request getAccessToken(req, res); with expired refresh token 

### SDK version

1.9.1

### Next.js version

10.2.3

### Node.js version

v16.13.0

adamjmcgrath:
### 📋 Changes

When the access token error comes from a failed grant request we should create an `AccessTokenError` with an `IdentityProvider` error as a cause

### 📎 References

fixes #872

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -1,7 +1,8 @@
 import { NextMiddleware, NextRequest, NextResponse } from 'next/server';
 import { default as CookieStore } from './auth0-session/cookie-store';
 import MiddlewareCookies from './utils/middleware-cookies';
-import { SessionCache, Session } from './session/';
+import Session from './session/session';
+import SessionCache from './session/cache';
 import {
   WithMiddlewareAuthRequired,
   default as withMiddlewareAuthRequiredFactory
--- a/src/session/get-access-token.ts
+++ b/src/session/get-access-token.ts
@@ -1,6 +1,7 @@
 import { IncomingMessage, ServerResponse } from 'http';
 import { NextApiRequest, NextApiResponse } from 'next';
-import { ClientFactory } from '../auth0-session';
+import type { errors } from 'openid-client';
+import { ClientFactory, IdentityProviderError } from '../auth0-session';
 import { AccessTokenError, AccessTokenErrorCode } from '../utils/errors';
 import { intersect, match } from '../utils/array';
 import { Session, SessionCache, fromTokenSet } from '../session';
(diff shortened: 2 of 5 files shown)


src/middleware.ts
```diff
@@ -1,7 +1,8 @@
 import { NextMiddleware, NextRequest, NextResponse } from 'next/server';
 import { default as CookieStore } from './auth0-session/cookie-store';
 import MiddlewareCookies from './utils/middleware-cookies';
-import { SessionCache, Session } from './session/';
+import Session from './session/session';
```
adamjmcgrath, line 4: MW was importing `get-access-token.ts` in it's binary, which it doesn't need

Widcket (approved):

PR #906, opened 2022-11-11
drop support for node 12 or lower

ahkhanjani:
Drops support for node 12 and lower.


adamjmcgrath:
Hi @ahkhanjani - thanks for raising this

What's your motivation for wanting to remove Node 10? (other than it being EOL)

ahkhanjani:
> Hi @ahkhanjani - thanks for raising this
> 
> What's your motivation for wanting to remove Node 10? (other than it being EOL)

Hi.
No other reason. Just for the sake of moving forward.

adamjmcgrath:
ok, thanks @ahkhanjani - we're dropping Node 10 in the next major release (see https://github.com/auth0/nextjs-auth0/blob/vNext/package.json#L20) so I'll close this in favour of that

PR #909, opened 2022-11-15
WithMiddlewareAuthRequired should return 401 for /api routes

### LINKED ISSUE #903: Protect API routes with withMiddlewareAuthRequired
### Checklist

- [X] I have looked into the [README](https://github.com/auth0/nextjs-auth0#readme) and have not found a suitable solution or answer.
- [X] I have looked into the [examples](https://github.com/auth0/nextjs-auth0/blob/main/EXAMPLES.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://auth0.github.io/nextjs-auth0/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/nextjs-auth0/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com/c/sdks/5) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

I have a NextJS server that has both page routes and API endpoints under `/api`. I want to use `withMiddlewareAuthRequired` to validate session token for all page routes and API endpoints that are not ignored by `config.matcher`. Once `@auth0/nextjs-auth0` did its magic, I want to execute custom code to check for the presence of specific roles and custom claims in the token. If these elements are not present, depending on the entry point (page route or API endpoint) I'd either redirect the user to Auth0 login or return a 401.

I can already craft my custom logic for the middleware by providing `withMiddlewareAuthRequired` a function. However right now when using `withMiddlewareAuthRequired` if I attempt to access a protected API route without any authentication information, the client gets redirected to the Auth0 login. 

IMO redirection is the desired behaviour when accessing page routes however when you access any protected endpoint under `/api` it should return a 401 without redirecting the client.

### Alternatives and current workarounds

I don't have any since the way the redirect is handled is outside of the user control. One way would be to ignore all API routes but it defeats the purpose of middleware IMO. I'd have to craft the token validation myself and apply it manually to all protected API endpoints which is error prone.

### Additional context

next@12.3.2
@auth0/nextjs-auth0@2.0.0-beta.1
node@16.15.1

adamjmcgrath:
### 📋 Changes

`/api` routes protected by `withMiddlewareAuthRequired` should respond with a 401 for unauthenticated routes (currently it just expects UI pages so only responds with temporary redirects to the login page)

Since middleware can't respond with a body - the recommended way to do this in Next.js middleware is to create a stub api route and rewrite to that

### 📎 References

https://nextjs.org/docs/messages/returning-response-body-in-middleware
fixes #903

### 🎯 Testing

```bash
$ npm run start:kitchen-sink
$ curl -I http://localhost:3000/api/hello-world-mw
HTTP/1.1 401 Unauthorized
```

### DIFF
--- a/examples/kitchen-sink-example/middleware.ts
+++ b/examples/kitchen-sink-example/middleware.ts
@@ -3,5 +3,5 @@ import { withMiddlewareAuthRequired } from '@auth0/nextjs-auth0/middleware';
 export default withMiddlewareAuthRequired();
 
 export const config = {
-  matcher: '/profile-mw'
+  matcher: ['/profile-mw', '/api/hello-world-mw']
 };
--- a/examples/kitchen-sink-example/pages/api/hello-world-mw.ts
+++ b/examples/kitchen-sink-example/pages/api/hello-world-mw.ts
@@ -1,0 +1,3 @@
+export default async function helloWorld(req, res) {
+  res.status(200).json({ text: 'Hello World!' });
+}
--- a/src/config.ts
+++ b/src/config.ts
@@ -308,6 +308,7 @@ export interface NextConfig extends Pick<BaseConfig, 'identityClaimFilter'> {
   routes: {
     callback: string;
     login: string;
+    unauthorized: string;
   };
 }
 
@@ -522,7 +523,8 @@ export const getConfig = (params: ConfigParameters = {}): { baseConfig: BaseConf
   const nextConfig = {
     routes: {
       ...baseConfig.routes,
-      login: baseParams.routes?.login || getLoginUrl()
+      login: baseParams.routes?.login || getLoginUrl(),
+      unauthorized: baseParams.routes?.unauthorized || '/api/auth/401'
     },
     identityClaimFilter: baseConfig.identityClaimFilter,
     organization: organization || AUTH0_ORGANIZATION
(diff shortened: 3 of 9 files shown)


Nargonath:
Thanks for the PR @adamjmcgrath. Shouldn't we update the documentation (README.md and EXAMPLES.MD) to mention that feature?

adamjmcgrath:
Thanks @Nargonath - I think the [existing docs](https://github.com/auth0/nextjs-auth0/blob/vNext/src/helpers/with-middleware-auth-required.ts#L5) explain it pretty well - my expectation would be that "pages" in that context would include `pages/api` as well. I am however going to take another pass through the README before we release the Beta

Nargonath:
The reason I mentioned the documentation is that from what I saw in the code it seems you give the ability to the user to override the path used to redirect API requests to the 401 API endpoint. I believe this is undocumented but perhaps I missed it.

Nargonath:
I refer to this: https://github.com/auth0/nextjs-auth0/pull/909/files#diff-c3095d5010e65c52737a98a5d618ea24049ebe90c8470752426081d70ed6e012R527

adamjmcgrath:
👍 yep, good point. I'm going to add some examples for your use case and will add docs for customising that path

Nargonath:
Alright, thank you.

PR #913, opened 2022-11-18
Rearrange exports for RSC and add experimental RSC route to example

adamjmcgrath:
### 📋 Changes

- moved `/frontend` -> `/client` to [align with Next.js](https://nextjs.org/learn/foundations/how-nextjs-works/client-and-server)
- moved client code out of the main export to `/client` (eg `import { useUser } from '@auth0/nextjs-auth0/client'`) because server components can't import client code
- moved `/middleware` -> `/edge`  (eg `import { withMiddlewareAuthRequired } from '@auth0/nextjs-auth0/edge'`) because this export will include code that can run on the Edge runtime in Server Components and API Routes in addition to Middleware.


src/helpers/testing.ts
```diff
@@ -1,6 +1,24 @@
-import { Config as BaseConfig, CookieStore, NodeCookies as Cookies } from '../auth0-session';
+import { Config as BaseConfig, CookieConfig, CookieStore, NodeCookies as Cookies } from '../auth0-session';
 import { Session } from '../session';
-import { GenerateSessionCookieConfig } from '../../testing';
+
+/**
+ * Configuration parameters used by ({@link generateSessionCookie}.
```
Widcket, line 5:
```suggestion
 * Configuration parameters used by {@link generateSessionCookie}.
```

src/client/index.ts
```diff
@@ -0,0 +1,19 @@
...
+  RequestError,
+  useUser
+} from './use-user';
+import {
+  default as withPageAuthRequired,
+  WithPageAuthRequired,
+  WithPageAuthRequiredProps,
+  WithPageAuthRequiredOptions
+} from './with-page-auth-required';
+export { ConfigProvider, ConfigProviderProps, useConfig };
```
Widcket, line 17: Do we want to export `useConfig` given that it's only used internally?

examples/kitchen-sink-example/app/profile-experimental-rsc/page.tsx
```diff
@@ -0,0 +1,32 @@
...
+};
+
+export function getSession() {
+  const { req, res } = reqRes();
+  return auth0GetSession(req, res);
+}
+
+export default async function ExperimentalRscPage() {
+  const session = await getSession();
+  // const session = {};
```
Widcket, line 24:
```suggestion
  const session = await getSession();
```

examples/kitchen-sink-example/components/header.tsx
```diff
@@ -1,12 +1,13 @@
...
 import Link from 'next/link';
-import { useUser } from '@auth0/nextjs-auth0';
+import { useUser } from '@auth0/nextjs-auth0/client';
 
 const Header = (): React.ReactElement => {
   const { user } = useUser();
 
   return (
-    <header>
+    <div id="header">
```
Widcket, line 10: Nit: should this be a `<nav>`?

examples/kitchen-sink-example/app/global.css
```diff
@@ -0,0 +1,9 @@
+body {
+    margin: 0;
+    color: #333;
+    font-family: -apple-system, 'Segoe UI';
```
Widcket, line 4:
```suggestion
    font-family: -apple-system, 'Segoe UI', sans-serif;
```

V2_MIGRATION_GUIDE.md
```diff
@@ -4,6 +4,7 @@ Guide to migrating from `1.x` to `2.x`
 
 - [Node 10 is no longer supported](#node-10-is-no-longer-supported)
 - [`getSession` now returns a `Promise`](#getsession-now-returns-a-promise)
+- [Client methods and components are now exported under /client](#client-methods-and-components-are-now-exported-under-client)
```
Widcket, line 7: Should we add an entry for the middleware (now under `/edge`) as well?

examples/kitchen-sink-example/components/header.tsx
adamjmcgrath: It should be a `<header>` - but the Server Component renderer was throwing errors on `<header>`, so I changed it to a `<div>`

src/client/index.ts
adamjmcgrath: Yeah, that's a fair point, will fix

adamjmcgrath: 7c6e917

PR #945, opened 2022-12-06
fix/rerenders: useMemo to avoid unnecessary rerenders

stavros-liaskos:
- All new/changed/fixed functionality is covered by tests **N/A**
- I have added documentation for all new/changed functionality **N/A**

### 📋 Changes
`useContext` is triggering unnecessary re-renders. With `useMemo` this can be avoided as described [here](https://blog.agney.dev/useMemo-inside-context/).

tl;tr
I noticed this when I integrated [why-did-you-render](https://github.com/welldone-software/why-did-you-render) to [release-raccoon](https://github.com/stavros-liaskos/release-raccoon/commit/b066d95068f210d505167afb39bf0a8ef85abc7e).  Atm it is using 2 contexts, I implemented the above fix to [ThemeProvider](https://github.com/stavros-liaskos/release-raccoon/commit/b066d95068f210d505167afb39bf0a8ef85abc7e#diff-e141c0adf082355a42cf2966d4290b966ffeecc03040917c3b421a1c3e1d243bR14) and this PR takes care of fixing this in `useUser` too.

### 📎 References
- Related issue from other repository: https://github.com/stavros-liaskos/release-raccoon/pull/20/files#r1041055599

### 🎯 Testing
Clicking on the `Home` btn calls the hook again but returns same results _different objects that are equal by value._
withoutFix
withUseMemo


adamjmcgrath:
Thanks for your contribution @stavros-liaskos!

Could you remove `why-did-you-render` and add a regression test similar to https://github.com/auth0/auth0-react/pull/318/files#diff-50f0ae9241a4ab6bc7b4c79924a45f52f654e4a03ad982942b68e0142a287877

src/client/use-user.tsx
```diff
@@ -214,10 +214,11 @@ export default ({
   }, [state.user]);
 
   const { user, error, isLoading } = state;
+  const value = useMemo(() => ({ user, error, isLoading, checkSession }), [user, error, isLoading]);
```
adamjmcgrath, line 217: Can you add `checkSession` to the list of dependent keys

adamjmcgrath (dismissed):
Thanks @stavros-liaskos

stavros-liaskos:
> Thanks @stavros-liaskos

Thank you @adamjmcgrath

Issue #977, opened 2022-12-23
withMiddlewareAuthRequired ignorePaths should not contain '/_next'


tisonkun:
### Checklist

- [X] The issue can be reproduced in the [sample app](https://github.com/auth0-samples/auth0-nextjs-samples/tree/main/Sample-01) (or N/A).
- [X] I have looked into the [README](https://github.com/auth0/nextjs-auth0#readme) and have not found a suitable solution or answer.
- [X] I have looked into the [examples](https://github.com/auth0/nextjs-auth0/blob/main/EXAMPLES.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://auth0.github.io/nextjs-auth0/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/nextjs-auth0/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com/c/sdks/5) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

```ts
      const ignorePaths = [login, callback, unauthorized, '/_next', '/favicon.ico'];
      if (ignorePaths.some((p) => pathname.startsWith(p))) {
        return;
      }
```

Currently, request to `/_next` always bypass the auth process. This can leak content when the hack figure out the path of the generated js file.

### Reproduction

https://everlasting-lgzkwcx3s-korandoru.vercel.app/_next/static/chunks/pages/index-045929068d457202.js

This page uses the basic middleware config:

```ts
import {withMiddlewareAuthRequired} from "@auth0/nextjs-auth0/edge";

export default withMiddlewareAuthRequired();
```

You can find the page content leaks by searching "Get started by editing" which is part of the content.

### SDK version

2.0.1

### Next.js version

13.1.0

### Node.js version

18

tisonkun:
It seems I may implement this with "For using middleware with your own instance of the SDK" example. I'll give it a try.

adamjmcgrath:
Hi @tisonkun - thanks for raising this.

This is by design, this middleware is not designed to protect static files like css and images. You tend to run into issues like https://github.com/nextauthjs/next-auth/issues/4848 if you start doing that

tisonkun:
[...]

PR #978, opened 2022-12-23
fix: remove _next from withMiddlewareAuthRequired ignorePaths

tisonkun:
This closes https://github.com/auth0/nextjs-auth0/issues/977.

Users can still use `export const config` to protect specific routes and bypass _next.

Manually tested.


tisonkun:
Maybe we can add a parameter to config `ignorePaths` completely. But this patch is general to avoid bypassing almost all context from protected.

tisonkun:
After first successful auth, these files should be able to access. Also users specific their own config often don't include `_next`.

adamjmcgrath:
Closing, see #977 

Happy to look at a PR that allows you to configure `ignorePaths` if you want to submit that (with tests)

PR #980, opened 2022-12-25
feat: support customized shouldBypassReq logics

tisonkun:
### 📋 Changes

Support intercept hardcode bypass logic in `withMiddlewareAuthRequired`.

### 📎 References

#977

### 🎯 Testing

New tests in `tests/helpers/with-middleware-auth-required.test.ts`

Open for review.

### DIFF
--- a/src/helpers/with-middleware-auth-required.ts
+++ b/src/helpers/with-middleware-auth-required.ts
@@ -43,7 +43,7 @@ import { SessionCache } from '../session';
  *
  * @category Server
  */
-export type WithMiddlewareAuthRequired = (middleware?: NextMiddleware) => NextMiddleware;
+export type WithMiddlewareAuthRequired = (middleware?: NextMiddleware, shouldBypassReq?: (req: NextRequest) => boolean) => NextMiddleware;
 
 /**
  * @ignore
@@ -52,13 +52,19 @@ export default function withMiddlewareAuthRequiredFactory(
   { login, callback, unauthorized }: { login: string; callback: string; unauthorized: string },
   getSessionCache: () => SessionCache<NextRequest, NextResponse>
 ): WithMiddlewareAuthRequired {
-  return function withMiddlewareAuthRequired(middleware?): NextMiddleware {
+  return function withMiddlewareAuthRequired(middleware?, shouldBypassReq?): NextMiddleware {
     return async function wrappedMiddleware(...args) {
       const [req] = args;
       const { pathname, origin } = req.nextUrl;
-      const ignorePaths = [login, callback, unauthorized, '/_next', '/favicon.ico'];
-      if (ignorePaths.some((p) => pathname.startsWith(p))) {
-        return;
+
+      const resolveShouldBypassReq = shouldBypassReq ? shouldBypassReq : (req: NextRequest) => {
+        const { pathname } = req.nextUrl;
+        const ignorePaths = [login, callback, unauthorized, '/_next', '/favicon.ico'];
+        return ignorePaths.some((p) => pathname.startsWith(p));
+      }
+
+      if (resolveShouldBypassReq(req)) {
+        return
       }
 
       const sessionCache = getSessionCache();
--- a/tests/helpers/with-middleware-auth-required.test.ts
+++ b/tests/helpers/with-middleware-auth-required.test.ts
@@ -34,8 +34,8 @@ const encrypted = async (claims: Partial<IdTokenClaims> = { sub: '__test_sub__' 
     .encrypt(key);
 };
 
-const setup = async ({ url = 'http://example.com', config = withoutApi, user, middleware }: any = {}) => {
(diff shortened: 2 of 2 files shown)


tisonkun:
@adamjmcgrath now I have one more reason to add this hook:

I'd like to bypass `img/favicon.ico`. With this patch it's more simple to [Match paths based on conditional statements](https://vercel.com/docs/concepts/functions/edge-middleware#match-paths-based-on-conditional-statements), instead of a complex regex like `/((?!img/favicon.ico).*)`.

adamjmcgrath:
Hi @tisonkun - thanks for raising this

I still don't see this as a valid use case and it should be fairly trivial to add your own middleware with this behaviour using `getSession` (see https://github.com/auth0/nextjs-auth0#edge-for-middleware-and-the-edge-runtime)

But I'll continue to assess this based on other use cases that may come to light

tisonkun:
@adamjmcgrath No. Before calling `getSession`, the middleware logic returned:

```js
const ignorePaths = [login, callback, unauthorized, '/_next', '/favicon.ico'];
      if (ignorePaths.some((p) => pathname.startsWith(p))) {
        return; // <-- NOTICE THIS LINE
```

So my own middleware hook has no change to be called.

tisonkun:
To be clear, the control flow gets returned at L61:

https://github.com/auth0/nextjs-auth0/blob/5c15a1e17dc0f56bdcdc832c69a7b85aa260a73b/src/helpers/with-middleware-auth-required.ts#L60-L62

... while customized middleware gets called at L76:

https://github.com/auth0/nextjs-auth0/blob/f8d2dac8ad74030fb46d9b863b0324f418bacffa/src/helpers/with-middleware-auth-required.ts#L76

adamjmcgrath:
@tisonkun - you don't need to use `withMiddlewareAuthRequired` to use `getSession`

```js
import { getSession } from '@auth0/nextjs-auth0/edge';
import { NextResponse } from 'next/server';

export default async function middleware(req) {
  // Ignore the paths you want
  const res = NextResponse.next();
  const user = await getSession(req, res);
  if (!user) {
    // redirect to login/401
  } else {
    return res;
  }
}

```

tisonkun:
@adamjmcgrath Aha. That's it.

Previously, I try to copy the middleware logics from `withMiddlewareAuthRequiredFactory` and find some methods are not exported. But maybe I don't need them at all.

BTW, here is the remaining part I don't understand and wonder whether I should keep it when customized:

```ts
      if (res) {
        const headers = new Headers(res.headers);
        const cookies = headers.get('set-cookie')?.split(', ') || [];
        const authCookies = authRes.headers.get('set-cookie')?.split(', ') || [];
        if (cookies.length || authCookies.length) {
          headers.set('set-cookie', [...authCookies, ...cookies].join(', '));
        }
        return NextResponse.next({ ...res, status: res.status, headers });
      } else {
        return authRes;
      }
```

tisonkun:
Finally I write something like:

```ts
export default async function middleware(req: NextRequest) {
    const {pathname, origin} = req.nextUrl

    const ignorePaths = ['/api/auth/login', '/api/auth/callback', '/api/auth/401', '/img/favicon.ico']
    if (ignorePaths.some((p) => pathname.startsWith(p))) return

    const res = NextResponse.next()
    const session = await getSession(req, res)

    if (!session?.user) {
        const unauthorized = NextResponse.rewrite(new URL('/api/auth/401', origin), {status: 401})
        const encodedReturnTo = encodeURIComponent(req.nextUrl.toString())
        const login = NextResponse.redirect(new URL(`/api/auth/login?returnTo=${encodedReturnTo}`, origin))
        return pathname.startsWith('/api') ? unauthorized : login
    }

    const user = session.user
    if (isWhitelistUser(user)) {
        if (path.extname(pathname)) return res
        if (pathname.startsWith('/api')) return res

        // WHAT: specially handle / or /docs case.
        //
        // WHY: Next.js will intercept these urls and result in a 404 page.
        // We should append an index.html segment to succeed in the first Docusaurus pages loading.
        // Then the Docusaurus SPA will take over the control flow.
        return NextResponse.redirect(new URL(path.resolve(pathname, 'index.html'), origin))
    }

    if (pathname.startsWith('/api/auth/logout')) return res

    const forbidden = new URL(`/api/auth/403?user=${encodeURIComponent(JSON.stringify(user))}`, origin)
    return NextResponse.rewrite(forbidden)
}
```

adamjmcgrath:
> BTW, here is the remaining part I don't understand and wonder whether I should keep it when customized:

This ensures that any updates to the session during the lifecycle of the middleware are saved back in the cookie, if you're not updating the session (and don't mind that the cookie expires isn't updated to bump the [rolling session](https://auth0.github.io/nextjs-auth0/interfaces/config.sessionconfig.html#rolling)) you don't need it

PR #987, opened 2023-01-03
Correct API file name

tisonkun:
I don't know the rationale concretely, but `/pages/api/auth/[...auth0].js` failed to start and `/pages/api/auth/[auth0].js` can work.

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -26,7 +26,7 @@ AUTH0_CLIENT_ID='CLIENT_ID'
 AUTH0_CLIENT_SECRET='CLIENT_SECRET'
 ```
 
-Create a [dynamic API route handler](https://nextjs.org/docs/api-routes/dynamic-api-routes) at `/pages/api/auth/[...auth0].js`.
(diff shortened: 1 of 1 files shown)


adamjmcgrath:
The API filename can be either a single dynamic route or a [catch all dynamic route](https://nextjs.org/docs/routing/dynamic-routes#catch-all-routes)

Either will work because we handle the case of the query being an array or a string here https://github.com/auth0/nextjs-auth0/blob/main/src/handlers/auth.ts#L168

We prefer to use the catch all dynamic route [in our docs](https://github.com/auth0/nextjs-auth0#add-the-dynamic-api-route) because this ultimately gives more flexibility in case we want to introduce nested folders under `/auth` in the future.

> but /pages/api/auth/[...auth0].js failed to start 

I can't reproduce the issue you're seeing on the example app - if you have an example you can share, or you can share the error you received along with steps to reproduce and care to raise an issue, I'd be happy to investigate.

tisonkun:
@adamjmcgrath Thanks for your reply! I meet this issue in a private repo. If I later find it during some open works, I'm glad to share an example :)

PR #993, opened 2023-01-09
SDK-3807 Add custom session stores

### LINKED ISSUE #279: Custom store implementation

A cookie implementation isn't sufficient when the amount of claims starts to rise and the browser will emit a `431 Request Header Fields Too Large`.
For those who'll need a different store strategy allowing for a custom store implementation will unblock without putting the work and maintenance on Auth0, as each user can bring their own store, a bit like `express-session`.

Allow to pass an instance of a Store in the params to overwrite the CookieStore.

No current workaround.
Currently patching the library to allow this.

adamjmcgrath:
Adding the ability to provide a key value store like Redis to back sessions and make them stateful.

```ts
import { SessionStore, SessionStorePayload, initAuth0 } from '@auth0/nextjs-auth0';

class Store implements SessionStore {
  private store: KeyValueStoreLikeRedis<SessionStorePayload>;
  constructor() {
    // If you set the expiry accross the whole store use the session config,
    // e.g. `min(config.session.rollingDuration, config.session.absoluteDuration)`
    // the default is 24hrs
    this.store = new KeyValueStoreLikeRedis();
  }
  async get(id) {
    const val = await this.store.get(id);
    return val;
  }
  async set(id, val) {
    // To set the expiry per item, use `val.header.exp` (in secs)
    const expiryMs = val.header.exp * 1000;
    await this.store.set(id, val);
  }
  async delete(id) {
    await this.store.delete(id);
  }
}

export default initAuth0({
  session: {
    store: new Store()
  }
});
```

4e57b7eb32b5137e031bdac14ce9b1c573d5b18a - Move the existing stateless cookie store
7a7e5d97e0e94d202f1eb0a4a6e52c274c02ed1d - Add stateful session store (**main commit**)
6468115714f89a8589f9feeee60031e4479f056c, c5b87d8753fe2b40a607de0d7176f87705411416, 3e97155cf01d719425ef4fc731247d626f51efac, efc518239ce8e6ed397cf6834ecd21ce020f9f6b - Other commits are merges and adding/reverting deployment for the beta

fixes #279

## Testing

Example app using a Redis session https://nextjs-auth0-redis-session.vercel.app/ (source code https://github.com/adamjmcgrath/nextjs-auth0-redis-session)


EXAMPLES.md
```diff
@@ -81,6 +82,79 @@ export default () => {
...
+  withPageAuthRequired,
+  getSession,
+  getAccessToken
+} from '@auth0/nextjs-auth0';
+```
+
+However, there are various reasons why you might want to create and manage an instance of the SDK yourself:
+
+- You may want to create your own instance for testing
+- You may not want to use environment variables for the configuration of secrets (eg using CredStash or AWS's Key Management Service)
```
Widcket, line 108:
```suggestion
- You may not want to use environment variables for the configuration of secrets (for example, to use CredStash or AWS's Key Management Service)
```

EXAMPLES.md
```diff
@@ -81,6 +82,79 @@ export default () => {
...
+> Note: You should not use the instance methods in combination with the named exports,
+> otherwise you will be creating multiple instances of the SDK. For example:
+
+```js
+// DON'T Mix instance methods and named exports
+import auth0 from '../../../utils/auth0';
+import { handleLogin } from '@auth0/nextjs-auth0';
+
+export default auth0.handleAuth({
+  // <= instance method
```
Widcket, line 145:
```suggestion
export default auth0.handleAuth({ // <= instance method
```

EXAMPLES.md
```diff
@@ -467,3 +469,56 @@ export const getServerSideProps = (ctx) => {
...
+
+```typescript
+// lib/auth0.ts
+import { SessionStore, SessionStorePayload, initAuth0 } from '@auth0/nextjs-auth0';
+
+class Store implements SessionStore {
+  private store: KeyValueStoreLikeRedis<SessionStorePayload>;
+  constructor() {
+    // If you set the expiry accross the whole store use the session config,
+    // e.g. `min(config.session.rollingDuration, config.session.absoluteDuration)`
```
Widcket, line 485:
```suggestion
    // for example `min(config.session.rollingDuration, config.session.absoluteDuration)`
```

EXAMPLES.md
Widcket: The doc team's guidelines instruct not to use latin abbreviations.

EXAMPLES.md
Widcket: The doc team's guidelines instruct not to use latin abbreviations.

EXAMPLES.md
```diff
@@ -467,3 +469,56 @@ export const getServerSideProps = (ctx) => {
...
+    // the default is 24hrs
```
Widcket, line 486:
```suggestion
    // the default is 24 hrs
```

EXAMPLES.md
```diff
@@ -467,3 +469,56 @@ export const getServerSideProps = (ctx) => {
...
+    this.store = new KeyValueStoreLikeRedis();
+  }
+  async get(id) {
+    const val = await this.store.get(id);
+    return val;
+  }
+  async set(id, val) {
+    // To set the expiry per item, use `val.header.exp` (in secs)
+    const expiryMs = val.header.exp * 1000;
```
Widcket, line 495: Where is `expiryMs` used?

src/auth0-session/session/abstract-session.ts
```diff
@@ -0,0 +1,125 @@
...
+import { CookieSerializeOptions } from 'cookie';
+import { Config } from '../config';
+import { Cookies } from '../utils/cookies';
+
+const debug = createDebug('session');
+
+export interface SessionPayload<Session> {
+  header: {
+    /**
+     * timestamp (in secs) when the session was created.
```
Widcket, line 11:
```suggestion
     * Timestamp (in secs) when the session was created.
```

src/auth0-session/session/abstract-session.ts
```diff
@@ -0,0 +1,125 @@
...
+     */
+    iat: number;
+    /**
+     * timestamp (in secs) when the session was last touched.
```
Widcket, line 15:
```suggestion
     * Timestamp (in secs) when the session was last touched.
```

src/auth0-session/session/abstract-session.ts
```diff
@@ -0,0 +1,125 @@
...
+     */
+    uat: number;
+    /**
+     * timestamp (in secs) when the session expires.
```
Widcket, line 19:
```suggestion
     * Timestamp (in secs) when the session expires.
```

EXAMPLES.md
adamjmcgrath: Fair, point - it depends on the store, will add an example for redis

```suggestion
    const expiryMs = val.header.exp * 1000;
    // Example for Redis: redis.set(id, val, { pxat: expiryMs });
```

src/auth0-session/session/stateful-session.ts
```diff
@@ -0,0 +1,115 @@
...
+import { Cookies } from '../utils/cookies';
+import { AbstractSession, SessionPayload } from './abstract-session';
+import { generateCookieValue, getCookieValue } from '../utils/signed-cookies';
+import { signing } from '../utils/hkdf';
+
+const debug = createDebug('stateful-session');
+
+export interface SessionStore<Session> {
+  /**
+   * Gets the session from the store given a session ID and passes it to `callback`.
```
Widcket, line 13: Where is `callback`?

src/auth0-session/session/stateful-session.ts
```diff
@@ -0,0 +1,115 @@
...
+   */
+  get(sid: string): Promise<SessionPayload<Session> | null | undefined>;
+
+  /**
+   * Upsert a session in the store given a session ID and `SessionData`
```
Widcket, line 18:
```suggestion
   * Upsert a session in the store given a session ID and `SessionData`.
```

src/auth0-session/session/stateful-session.ts
adamjmcgrath: Good point

```suggestion
   * Gets the session from the store given a session ID.
```

src/auth0-session/session/stateful-session.ts
```diff
@@ -0,0 +1,115 @@
...
+  Req,
+  Res,
+  Session extends { [key: string]: any } = { [key: string]: any }
+> extends AbstractSession<Req, Res, Session> {
+  private keys?: Uint8Array[];
+  private store: SessionStore<Session>;
+
+  constructor(protected config: Config, protected Cookies: new () => Cookies) {
+    super(config, Cookies);
+    this.store = config.session.store as SessionStore<Session>;
```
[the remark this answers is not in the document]
adamjmcgrath: I think I had a good reason, let me check...

src/auth0-session/session/stateful-session.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    if (!sessionId) {
+      sessionId = await genid!(req);
+      debug('generated new session id %o', sessionId);
+      const cookieValue = await generateCookieValue(sessionName, sessionId, keys[0]);
+      cookieSetter.set(sessionName, cookieValue, cookieOptions);
+      cookieSetter.commit(res);
+    }
+    debug('set session %o', sessionId);
+    await this.store.set(sessionId, {
```
Widcket, line 95: Just checking, if the session storage operation fails and a new cookie was generated previously (L92), will this cookie still be set?

src/auth0-session/session/abstract-session.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  abstract getSession(req: Req): Promise<SessionPayload<Session> | undefined | null>;
+
+  abstract setSession(
+    req: Req,
+    res: Res,
+    session: Session,
+    uat: number,
+    iat: number,
+    exp: number,
```
Widcket, line 49: If in `getSession()` we're returning a `SessionPayload<Session>`, should `setSession` take a `SessionPayload<Session>` as well?

src/auth0-session/session/stateless-session.ts
```diff
@@ -0,0 +1,159 @@
...
+const notNull = <T>(value: T | null): value is T => value !== null;
+
+export class StatelessSession<
+  Req,
+  Res,
+  Session extends { [key: string]: any } = { [key: string]: any }
+> extends AbstractSession<Req, Res, Session> {
+  private keys?: Uint8Array[];
+
+  private chunkSize: number;
```
Widcket, line 24:
```suggestion
  private keys?: Uint8Array[];
  private chunkSize: number;
```

src/auth0-session/session/stateless-session.ts
```diff
@@ -0,0 +1,159 @@
...
+    };
+    if (!transient) {
+      cookieOptions.expires = new Date();
+    }
+
+    const emptyCookie = serialize(`${sessionName}.0`, '', cookieOptions);
+    this.chunkSize = MAX_COOKIE_SIZE - emptyCookie.length;
+  }
+
+  private async getKeys(): Promise<Uint8Array[]> {
```
Widcket, line 43: Should this method be in `AbstractSession`?

src/auth0-session/config.ts
```diff
@@ -174,6 +175,19 @@ export interface SessionConfig {
...
+   * you can provide a store with `get`, `set` and `destroy` methods to store the session on the server side.
+   */
+  store?: SessionStore<any>;
+
+  /**
+   * A function for generating a session id when using a custom session store.
+   *
+   * **IMPORTANT** You must use a suitably unique value to prevent collisions.
+   */
+  genid?: <Req = any>(req: Req) => string | Promise<string>;
```
Widcket, line 189: Nit: given that we're using camelCase, and `genid` stands for `gen id`, should it be cased `genId`?

src/auth0-session/session/stateful-session.ts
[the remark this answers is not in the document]
adamjmcgrath: I get errors dong this because `Session` has required fields

image

src/auth0-session/session/abstract-session.ts
adamjmcgrath: That would makes sense 😄 

But the way the 2 session classes create the header data (one in a JWT header, one in a nested object) - means that it makes more sense for them to create the full payload themselves

src/auth0-session/session/stateless-session.ts
adamjmcgrath: The 2 session classes get different types of keys (signing for stateful and encrypting for stateless) so you can't share the code. And it's reasonable to expect that an implementation of abstract session might not need keys for either.

src/auth0-session/config.ts
adamjmcgrath: Yep, I used the same values as express-openid-connect (that is influenced by express-session) - but you're right, this should be camel case - will fix

src/auth0-session/session/stateful-session.ts
adamjmcgrath: Yes, you'd be left with a cookie pointing to a non existing entry in the store - but the SDK handles this scenario by ignoring it and cleaning it up on the next request.

src/config.ts
```diff
@@ -186,6 +188,20 @@ export interface SessionConfig {
...
+   * By default, the session is stateless and stored in an encrypted cookie. But if you want a stateful session
+   * you can provide a store with `get`, `set` and `destroy` methods to store the session on the server.
+   */
+  store?: SessionStore<Session>;
+
+  /**
+   * A Function for generating a session id when using a custom session store.
+   *
+   * **IMPORTANT** If you override this, you must use a suitable value from your platform to
+   * prevent collisions. e.g. for Node: `require('crypto').randomBytes(16).toString('hex')`
```
Widcket, line 201:
```suggestion
   * prevent collisions. For example, for Node: `require('crypto').randomBytes(16).toString('hex')`.
```

PR #1023, opened 2023-01-23
Allow client to pass screen_hint and login_hint via params or url params

### LINKED ISSUE #180: Pass login_hint in from the request to /api/login

I'd like to pass `login_hint` to the login page via a link from my page. eg I'd like this to work `/api/login?login_hint=email@foo.com`

It would just pick it up from the request, similar to `redirect_to`

To get this behavior I have to do the following

```js
export default async function login (req, res) {
  try {
    await auth0.handleLogin(req, res, {
      authParams: {
        login_hint: req.query.login_hint,
      },
    })
  } catch (error) {
    console.error(error)
    res.status(error.status || 400).end(error.message)
  }
}
```

### Additional context

Is this a security risk? I can't see it being any more dangerous than `redirect_to`, but ya'll the experts.

Thanks!

rhysawilliams2010:

### 📋 Changes

Added the ability to pass `screenHint` and `loginHint` to the login url so that they can be passed in `authorizationParams` to auth0 `handleLogin`

```
    const ProtectedPage = withPageAuthRequired(MyPage, {
      screenHint: 'signup',
      loginHint: 'test@test.com'
    });

```

### 📎 References
Fixes #180

### DIFF
--- a/src/client/with-page-auth-required.tsx
+++ b/src/client/with-page-auth-required.tsx
@@ -3,6 +3,7 @@ import React, { ComponentType, useEffect } from 'react';
 
 import { useConfig } from './use-config';
 import { useUser, UserProfile } from './use-user';
+import querystring, { stringify } from 'querystring';
 
 /**
  * @ignore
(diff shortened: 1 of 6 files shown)


adamjmcgrath:
Hi @rhysawilliams2010 - thanks for your contribution, always appreciated

Just passing `screenHint` or `loginHint` to the login url wont send these on to the authorization server, you'd need to create a custom handler to accept these arguments and pass them on (which you cold do [quite simply](https://github.com/auth0/nextjs-auth0/blob/main/EXAMPLES.md#add-a-signup-handler))

Also, if you are using the Classic Universal Login experience, just passing `screen_hint: 'signup'` wont show the signup dialog, you'd need to write some custom code in Lock.js (see https://github.com/auth0/nextjs-auth0/blob/v1.9.2/EXAMPLES.md#add-a-signup-handler)

Also, there are a large number of possible parameters you might want to pass on the the auth0 authorization server (or the social connection's authorization server) - too many to fully support by explicitly adding this type of logic to every helper.

For the above reasons I'm going to close your PR.

However, it is a reasonable request to want to overwrite the login url at runtime - to make adding support for things like `login_hint` simpler - so I'd be happy to look at a PR that let you override this in the helpers options. eg

```js
withMiddlewareAuthRequired({ loginUrl: '/api/auth/login?foo=bar' });
withPageAuthRequired({ loginUrl: '/api/auth/login?foo=bar' });
```

kneza23:
is there a way @adamjmcgrath to create a custom handler just for login page to pass login_hint to auth server from query params?

Issue #1049, opened 2023-02-03
Support for Access Token Authentication


Wolfleader101:
### Checklist

- [X] I have looked into the [README](https://github.com/auth0/nextjs-auth0#readme) and have not found a suitable solution or answer.
- [X] I have looked into the [examples](https://github.com/auth0/nextjs-auth0/blob/main/EXAMPLES.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://auth0.github.io/nextjs-auth0/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/nextjs-auth0/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com/c/sdks/5) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

I would like to be able to use nextjs api routes from my mobile app, which uses auth0 for authentication. Currently there isnt any way for me to validate this access token in my api routes. 

### Describe the ideal solution

Ideally I would like to be able to generate a valid session between my react native app, and the nextjs api routes using the access token, thus allowing me to access the api route.

adamjmcgrath:
See https://github.com/auth0/nextjs-auth0/pull/1052#issuecomment-1418801403

Wolfleader101:
> See [#1052 (comment)](https://github.com/auth0/nextjs-auth0/pull/1052#issuecomment-1418801403)

@adamjmcgrath Thanks, responded in the PR :)

PR #1052, opened 2023-02-04
Jwt api auth

Wolfleader101:

### 📋 Changes

I have gone and added in API Auth protection for JWT Access Tokens. This means you can use withAPIAuthRequired, and still allow external API calls, as long as you receive either a valid session cookie, or valid JWT access token. It uses the `AUTH0_ISSUER_BASE_URL` to get the JWK's to correctly verify the token.

### 📎 References

GitHub Issue I made earlier: https://github.com/auth0/nextjs-auth0/issues/1049

### 🎯 Testing

This can be tested by reviewers by setting up a simple API endpoint with the `withAPIAuthRequired` guard. Using a valid access token (can get from https://auth0.com/docs/secure/tokens/access-tokens/get-access-tokens)

*Please note still in testing*


Wolfleader101:
These changes are ready to be reviewed :)

Wolfleader101:
These changes are ready to be reviewed :)

adamjmcgrath:
Hi @Wolfleader101 - thanks for your contribution

This SDK is for protecting routes with a cookie session, since the API routes are on the same domain as the application there should be no reason to use a bearer access token - therefore, I'm afraid we have no plans to add this functionality to the SDK. Many thanks again for your contribution and I'm sorry we couldn't merge it.

adamjmcgrath:
Thanks for the feedback - and I'm sure it is a useful feature for people wanting to use the api from outside of their next app.

But this feature is outside of the scope of this library, which is essentially an openid-connect client for doing web sign on with auth0.

If we did implement this feature for next js it would be in another SDK, and would probably live in this repo https://github.com/auth0/node-oauth2-jwt-bearer - which is for protecting apis with jwt bearer access tokens

PR #1055, opened 2023-02-06
Fixed the Claims Type

Wolfleader101:

### 📋 Changes

`Claims` is now properly typed, this allows the user to get type safety when accessing `session.user`.
`export type Claims = IdTokenClaims | { [key: string]: unknown };`

This could be extended to include the auth0 meta data key using generics.

### DIFF
--- a/src/session/session.ts
+++ b/src/session/session.ts
@@ -1,4 +1,4 @@
-import type { TokenSet } from 'openid-client';
+import type { IdTokenClaims, TokenSet } from 'openid-client';
 import { Config } from '../auth0-session';
 import { NextConfig } from '../config';
 
@@ -7,9 +7,7 @@ import { NextConfig } from '../config';
  *
  * @category Server
  */
-export interface Claims {
-  [key: string]: any;
-}
+export type Claims = IdTokenClaims | { [key: string]: unknown };
 
 /**
  * The user's session.


adamjmcgrath:
Hi @Wolfleader101 - thanks for raising this

`Claims` is the wrong name for that type (which is my bad) - since most of the standard oidc claims (`aud`, `exp`, `iss` etc) are [stripped off](https://github.com/auth0/nextjs-auth0/blob/main/src/session/session.ts#L66-L68) when the session is created. Only the user profile claims remain. 

We already have a type for `UserProfile` which we should reuse from here https://github.com/auth0/nextjs-auth0/blob/main/src/client/use-user.tsx#L11

My concerns with using this are that we will create compilation errors from everyone using the `user` object in their TypeScript apps (they will need to start checking if `user.name` exists and start casting unknown properties before using them). I think we should change this, but we'll have to wait until our next breaking change so it's not too disruptive.

In the meantime, for type safety in your methods, I would recommend defining user as something stricter when you access it, for example:

```ts
type MyUser = {
  sub: string;
};

const session = await getSession(req, res);
const user = session.user as MyUser;
```

Wolfleader101:
Hi @adamjmcgrath ,

Sorry missed the type already there. I don't believe changing it should cause any errors for developers using users, or be disruptive, as you can still keep it with  `{ [key: string]: unknown };` thus not actually breaking any previous tests, or apps, however still getting nice type safety when it comes to the known fields.

adamjmcgrath:
Changing from `{ [key: string]: any };` to `{ [key: string]: unknown };` will break some use cases, see the following example

image

Thanks for your contribution, but we're going to wait for a new major to introduce this improvement

PR #1074, opened 2023-02-21
fix: no caching client side /auth/me

benevbright:

### 📋 Changes

A while ago, this PR fixed the profile fetch's cache: https://github.com/auth0/nextjs-auth0/pull/233

But the client's `/auth/me` fetch is still not using any cache option,
which leads to an issue that the user can't be redirected to the re-login page when the token has been expired even if they reload a page. 

We found on our app that the library is calling `/auth/me` successfully with `200` **when the token has been actually expired**.

### DIFF
--- a/src/client/use-user.tsx
+++ b/src/client/use-user.tsx
@@ -178,7 +178,9 @@ type UserProviderState = {
 const userFetcher: UserFetcher = async (url) => {
   let response;
   try {
-    response = await fetch(url);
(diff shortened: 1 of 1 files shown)


benevbright:
Screenshot 2023-02-21 at 16 21 22

It's weird... it already has the `no-store` cache option. 

I don't know why this is sometimes not failing.

PR #1104, opened 2023-03-10
[SDK-3887] Always honor auth0Logout config

adamjmcgrath:
See https://github.com/auth0/express-openid-connect/pull/447

### Description

The OIDC RP Initiated Logout endpoint is incompatible with Auth0's proprietary logout. Make sure this SDK does not use it if `auth0Logout` is configured and an `end_session_endpoint` is Discovered in the OIDC Discovery document.

### Testing

If `auth0Logout` is true -> use v2/logout regardless of discovery
If `auth0Logout` is false -> use discovered endpoint or nothing
If `auth0Logout` is not set -> use v2/logout regardless of discovery (this is different from express as this is an auth0 SDK)

### Checklist

- [ ] The correct base branch is being used, if not the default branch

### DIFF
--- a/tests/auth0-session/client.test.ts
+++ b/tests/auth0-session/client.test.ts
@@ -105,6 +105,66 @@ describe('clientFactory', function () {
     expect(client.id_token_signed_response_alg).toEqual('RS256');
   });
 
+  it('should use discovered logout endpoint by default', async function () {
+    const client = await getClient({ ...defaultConfig, idpLogout: true });
+    expect(client.endSessionUrl({})).toEqual('https://op.example.com/session/end?client_id=__test_client_id__');
+  });
+
+  it('should use auth0 logout endpoint if configured', async function () {
+    const client = await getClient({ ...defaultConfig, idpLogout: true, auth0Logout: true });
+    expect(client.endSessionUrl({})).toEqual('https://op.example.com/v2/logout?client_id=__test_client_id__');
+  });
+
+  it('should use auth0 logout endpoint if domain is auth0.com', async function () {
+    nock('https://foo.auth0.com')
+      .get('/.well-known/openid-configuration')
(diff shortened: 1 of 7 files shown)


tests/auth0-session/client.test.ts
```diff
@@ -105,6 +105,66 @@ describe('clientFactory', function () {
...
+    const client = await getClient({
+      ...defaultConfig,
+      issuerBaseURL: 'https://foo.auth0.com',
+      idpLogout: true,
+      auth0Logout: true
+    });
+    expect(client.endSessionUrl({})).toEqual('https://foo.auth0.com/v2/logout?client_id=__test_client_id__');
+  });
+
+  it('should not use discovered logout endpoint if domain is auth0.com but configured with auth0logout false', async function () {
```
adamjmcgrath, line 139:
```suggestion
  it('should use discovered logout endpoint if domain is auth0.com but configured with auth0logout false', async function () {
```

blakeplumb:
@adamjmcgrath is there a way to set `auth0Logout` to false via an environment variable.  We don't use the base config, and removing the default as false broke our logouts

adamjmcgrath:
Hi @blakeplumb - how did `auth0Logout` default break your logout, are you using this with a different identity provider than auth0?

KentDi801:
@adamjmcgrath - I work with Blake. We are currently using @auth0/nextjs-auth0 because it gives us great support for doing OAuth2 auth in NextJS but we are using Okta as our Identity provider. When we go to logout it passes `oauth2/default/v2/logout?returnTo=http` in the url and we get a 404 error because our tenants to not support `v2` for logout.

adamjmcgrath:
Hi @KentDi801 - we should have an environment variable for auth0logout, so you can set it to false. Will add a PR for that shortly.

KentDi801:
@adamjmcgrath thank you for the quick response on this.

PR #1105, opened 2023-03-10
[SDK-3887] Always honor auth0Logout config for v1

adamjmcgrath:
Same as https://github.com/auth0/nextjs-auth0/pull/1104 for v1 (note the base branch)

### DIFF
--- a/tests/auth0-session/client.test.ts
+++ b/tests/auth0-session/client.test.ts
@@ -131,7 +131,7 @@ describe('clientFactory', function () {
     expect(client.endSessionUrl({})).toEqual('https://foo.auth0.com/v2/logout?returnTo=&client_id=__test_client_id__');
   });
 
-  it('should not use discovered logout endpoint if domain is auth0.com but configured with auth0logout false', async function () {
+  it('should use discovered logout endpoint if domain is auth0.com but configured with auth0logout false', async function () {
     nock('https://foo.auth0.com')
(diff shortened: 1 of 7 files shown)


tests/auth0-session/client.test.ts
```diff
@@ -100,6 +100,66 @@ describe('clientFactory', function () {
...
+    const client = await getClient({
+      ...defaultConfig,
+      issuerBaseURL: 'https://foo.auth0.com',
+      idpLogout: true,
+      auth0Logout: true
+    });
+    expect(client.endSessionUrl({})).toEqual('https://foo.auth0.com/v2/logout?returnTo=&client_id=__test_client_id__');
+  });
+
+  it('should not use discovered logout endpoint if domain is auth0.com but configured with auth0logout false', async function () {
```
Widcket, line 134: Should this test be described as "should not *auth0* logout endpoint if domain is auth0.com but configured with auth0logout false"?

Widcket: Or, 'should use discovered logout endpoint if domain is auth0.com but configured with auth0logout false'.

adamjmcgrath: Yep that one ☝️ - let me change it

```suggestion
  it('should use discovered logout endpoint if domain is auth0.com but configured with auth0logout false', async function () {
```

Widcket (approved):

PR #1113, opened 2023-03-16
Add AUTH0_LOGOUT env var

adamjmcgrath:
### 📋 Changes

Add the `AUTH0_LOGOUT` environment variable so you can configure `auth0Logout` for the named exports as well as own instance configuration of the SDK.

### 📎 References

See https://github.com/auth0/nextjs-auth0/pull/1104#issuecomment-1470496486

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -25,7 +25,9 @@ export interface BaseConfig {
 
   /**
    * Boolean value to enable Auth0's proprietary logout feature.
-   * Since this SDK is for Auth0, it's set to `true`by default.
+   * Since this SDK is for Auth0, it's set to `true` by default.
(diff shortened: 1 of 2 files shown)


src/config.ts
```diff
@@ -25,7 +25,9 @@ export interface BaseConfig {
 
   /**
    * Boolean value to enable Auth0's proprietary logout feature.
-   * Since this SDK is for Auth0, it's set to `true`by default.
+   * Since this SDK is for Auth0, it's set to `true` by default.
+   * Set it to `false` if you don't want to use https://auth0.com/docs/api/authentication#logout
```
Widcket, line 29:
```suggestion
   * Set it to `false` if you don't want to use https://auth0.com/docs/api/authentication#logout.
```

Widcket (approved):

PR #1237, opened 2023-06-08
VERCEL_BRANCH_URL is no longer set

samuelcole:
You have to use `VERCEL_URL`, since `VERCEL_BRANCH_URL` is deprecated.


adamjmcgrath:
> `VERCEL_BRANCH_URL` is deprecated.

@smaeda-ks is this true?

adamjmcgrath:
Thanks for raising this @samuelcole 

I've been told by Vercel that `VERCEL_BRANCH_URL` is now generally available to all customers so this PR should no longer be needed

Issue #912, opened 2022-11-17
"The getAccessToken method can only be used from the server side" when used in middleware.js


robertwbradford:
### Checklist

- [X] The issue can be reproduced in the [sample app](https://github.com/auth0-samples/auth0-nextjs-samples/tree/main/Sample-01) (or N/A).
- [X] I have looked into the [README](https://github.com/auth0/nextjs-auth0#readme) and have not found a suitable solution or answer.
- [X] I have looked into the [examples](https://github.com/auth0/nextjs-auth0/blob/main/EXAMPLES.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://auth0.github.io/nextjs-auth0/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/nextjs-auth0/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com/c/sdks/5) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

Hello, I see there is a [`withMiddlewareAuthRequired`](https://github.com/auth0/nextjs-auth0/blob/v2.0.0-beta.3/EXAMPLES.md#protecting-pages-with-middleware) helper coming in v2.

We have a use case where we'd like to use `getAccessToken` in a Next.js middleware file _but not require authentication_. That is, auth is optional, but we will do additional things if a user is authenticated.

However, in a `middleware.js` file, the call to `getAccessToken()` is reporting: `Error: The getAccessToken method can only be used from the server side.`Seems that its check for the server environment is not accounting for the Next.js middleware feature.

Any ideas on how to accomplish this?

### Reproduction

I have the following in a `middleware.js` file:
```js
import { NextResponse } from "next/server";
import { getAccessToken } from "@auth0/nextjs-auth0";

export async function middleware(req) {
  const res = NextResponse.next();
  const { accessToken } = await getAccessToken(req, res);
  console.log("Middleware", accessToken);
}
```

### SDK version

^1.9.2

### Next.js version

^12.3.0

### Node.js version

v16.16.0

adamjmcgrath:
Hi @robertwbradford - thanks for raising

[...]

Issue #1257, opened 2023-06-17
Support edge runtime in app based routes


gerlv:
### Checklist

- [X] I have looked into the [Readme](https://github.com/auth0/nextjs-auth0#readme), [Examples](https://github.com/auth0/nextjs-auth0/blob/main/EXAMPLES.md), and [FAQ](https://github.com/auth0/nextjs-auth0/blob/main/FAQ.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://auth0.github.io/nextjs-auth0/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/nextjs-auth0/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

NextJS Auth0 can work with edge runtime.

At the moment, only the middleware supports the edge runtime. You can't use edge based getSession in app routes.

### Describe the ideal solution

Edge runtime supported for app based routes

### Alternatives and current workarounds

Use nodejs runtime

adamjmcgrath:
Hi @gerlv 

> You can't use edge based getSession in app routes

You can use the edge based `getSession`  in app based routes in the new Beta (see #1235)

eg.

```ts
import { getSession } from '@auth0/nextjs-auth0/edge';
import { NextRequest, NextResponse } from 'next/server';

const GET = async (req: NextRequest) => {
  const res = new NextResponse();
  const session = await getSession(req, res);

  return NextResponse.json(session?.user, res);
};

export { GET };
```

gerlv:
@adamjmcgrath thank you for a code snippet, I can see it works, but as far as I can see this wouldn't work with app based routes, and you can't override `handleAuth` functionality as all imports are from `@auth0/nextjs-auth0`.

I.e. there is a way to get the session, but the large part of the functionality can't be used with edge runtime.

adamjmcgrath:
> as I can see this wouldn't work with app based routes

Can you elaborate? The code snippet I shared works with app based routes (on the edge runtime) in the new beta, have tested it on the example app

> but the large part of the functionality can't be used with edge runtime

[...]

Issue #1150, opened 2023-04-08
withMiddlewareAuthRequired breaks Readablestream response


codewitch:
### Checklist

- [X] The issue can be reproduced in the [sample app](https://github.com/auth0-samples/auth0-nextjs-samples/tree/main/Sample-01) (or N/A).
- [X] I have looked into the [README](https://github.com/auth0/nextjs-auth0#readme) and have not found a suitable solution or answer.
- [X] I have looked into the [examples](https://github.com/auth0/nextjs-auth0/blob/main/EXAMPLES.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://auth0.github.io/nextjs-auth0/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/nextjs-auth0/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com/c/sdks/5) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

I have an nextjs edge api that returns a ReadableStream. It works great, but when i wrap it in a withMiddlewareAuthRequired to gate the api for users only the stream does not end up being returned. 

### Reproduction

the api code looks something like this. I have abstracted out the stream logic in to a `getStream()` function for simplicity:
```
export const config = {  runtime: "edge" };

export default withMiddlewareAuthRequired(async function middleware(req) {
  const completionsStream = await getStream();
  return new Response(completionsStream);
});
```

### SDK version

2.4.0

### Next.js version

13.3.0

### Node.js version

18

adamjmcgrath:
Hi @codewitch - thanks for raising this

Support for [producing a response](https://nextjs.org/docs/advanced-features/middleware#producing-a-response) in Next.js middleware was added after we added `withMiddlewareAuthRequired` - so it does not currently support this use case. Let me schedule so work to fix this.

In the meantime, you should be able to use `getSession` as a workaround.

```js
import { NextResponse } from "next/server";
import { getSession } from "@auth0/nextjs-auth0/edge";

export async function middleware(req) {
  const res = NextResponse.next();
  const session = await getSession(req, res);
  if (session) {
    // logged in
  } else {
    // redirect to login
  }
}
```

adamjmcgrath:
[...]

PR #1269, opened 2023-06-26
[SDK-4319] Add support for Edge runtime

adamjmcgrath:
### 📋 Changes

Have added support for the Edge runtime (in the app directory) to the auth handlers and session helpers:

**Auth Handlers**
`handleAuth`
`handleLogin`
`handleCallback`
`handleLogout`
`handleProfile`

**Session Helpers**
`withApiAuthRequired`
`withPageAuthRequired`
`updateSession`
`getAccessToken`
(`getSession` is already supported)

We now use 2 OIDC clients:
- For Node: https://github.com/panva/node-openid-client
- For Edge: https://github.com/panva/oauth4webapi

Using Jest "projects" to run the handler tests in both runtimes
Have updated the example app to use edge as well as node
Also support added for middleware responses (see https://github.com/auth0/nextjs-auth0/issues/1150)

### 📎 References

Fixes https://github.com/auth0/nextjs-auth0/issues/912 https://github.com/auth0/nextjs-auth0/issues/1150 https://github.com/auth0/nextjs-auth0/issues/1257

### 🎯 Testing

There is an example app running on the edge runtime on Vercel https://app-dir-auth0.vercel.app/ ([source](https://github.com/adamjmcgrath/app-dir-auth0/blob/main/app/api/auth/%5Bauth0%5D/route.ts#L9))


example-app/app/nav.tsx
```diff
@@ -41,28 +41,52 @@ export default function Nav() {
                 <a>Profile</a>
               </Link>
             </li>
+            <li>
+              <Link href="/edge-profile" legacyBehavior>
```
Widcket, line 45: Is `legacyBehavior` still needed here (and in other links)?

src/auth0-session/client/edge-client.ts
```diff
@@ -0,0 +1,241 @@
...
+  private async getClient(): Promise<[oauth.AuthorizationServer, oauth.Client]> {
+    if (this.as) {
+      return [this.as, this.client as oauth.Client];
+    }
+
+    const issuer = new URL(this.config.issuerBaseURL);
+    try {
+      this.as = await oauth
+        .discoveryRequest(issuer, this.httpOptions())
+        .then((response) => oauth.processDiscoveryResponse(issuer, response));
```
Widcket, line 73: Is the discovery doc being cached?

src/auth0-session/client/edge-client.ts
```diff
@@ -0,0 +1,241 @@
...
+      Object.entries(extraParams).forEach(([key, value]: [string, string]) => {
+        if (value === null || value === undefined) {
+          return;
+        }
+        auth0LogoutUrl.searchParams.set(key, value);
+      });
+      return auth0LogoutUrl.toString();
+    }
+    if (!as.end_session_endpoint) {
+      throw new Error('no rp inititated logout');
```
Widcket, line 188: Should this error have a more specific type?

src/auth0-session/client/edge-client.ts
```diff
@@ -0,0 +1,241 @@
...
+    const [as, client] = await this.getClient();
+
+    const { clientAssertionSigningKey, clientAssertionSigningAlg } = this.config;
+
+    let clientPrivateKey = clientAssertionSigningKey as CryptoKey | undefined;
+    /* c8 ignore next 3 */
+    if (clientPrivateKey && !(clientPrivateKey instanceof CryptoKey)) {
+      clientPrivateKey = await jose.importPKCS8<CryptoKey>(clientPrivateKey, clientAssertionSigningAlg || 'RS256');
+    }
+    const response = await oauth.authorizationCodeGrantRequest(
```
Widcket, line 137: Can this call throw an exception? If so, should it be caught, wrapped, and rethrown?

src/auth0-session/client/edge-client.ts
```diff
@@ -0,0 +1,241 @@
...
+      redirectUri,
+      checks.code_verifier as string,
+      {
+        additionalParameters: extras.exchangeBody,
+        ...(clientPrivateKey && { clientPrivateKey }),
+        ...this.httpOptions()
+      }
+    );
+
+    const result = await oauth.processAuthorizationCodeOpenIDResponse(
```
Widcket, line 150: Same as above.

src/auth0-session/client/edge-client.ts
```diff
@@ -0,0 +1,241 @@
...
+      oidcLogoutUrl.searchParams.set(key, value);
+    });
+
+    oidcLogoutUrl.searchParams.set('client_id', this.config.clientID);
+    return oidcLogoutUrl.toString();
+  }
+
+  async userinfo(accessToken: string): Promise<Record<string, unknown>> {
+    const [as, client] = await this.getClient();
+    const response = await oauth.userInfoRequest(as, client, accessToken, this.httpOptions());
```
Widcket, line 204: Can this call throw an exception? If so, should it be caught, wrapped, and rethrown?

src/auth0-session/client/edge-client.ts
```diff
@@ -0,0 +1,241 @@
...
+
+    return oauth.processUserInfoResponse(as, client, oauth.skipSubjectCheck, response);
+  }
+
+  async refresh(refreshToken: string, extras: { exchangeBody: Record<string, any> }): Promise<TokenEndpointResponse> {
+    const [as, client] = await this.getClient();
+    const res = await oauth.refreshTokenGrantRequest(as, client, refreshToken, {
```
Widcket, line 211: Can this call throw an exception? If so, should it be caught, wrapped, and rethrown?

src/auth0-session/client/edge-client.ts
```diff
@@ -0,0 +1,241 @@
...
+      additionalParameters: extras.exchangeBody,
+      ...this.httpOptions()
+    });
+    const result = await oauth.processRefreshTokenResponse(as, client, res);
```
Widcket, line 215: Same as above.

src/auth0-session/handlers/login.ts
```diff
@@ -65,27 +62,31 @@ export default function loginHandlerFactory(
...
-    assert(
-      validResponseTypes.includes(authParams.response_type as string),
-      `response_type should be one of ${validResponseTypes.join(', ')}`
-    );
-    assert(/\bopenid\b/.test(authParams.scope as string), 'scope should contain "openid"');
+    if (!validResponseTypes.includes(authParams.response_type as string)) {
+      throw new Error(`response_type should be one of ${validResponseTypes.join(', ')}`);
+    }
+    if (!/\bopenid\b/.test(authParams.scope as string)) {
+      throw new Error('scope should contain "openid"');
```
Widcket, line 88: Should these errors have a more specific type?

src/edge.ts
```diff
@@ -28,39 +29,57 @@ const genId = () => {
     .join('');
 };
 
-function getInstance(params?: ConfigParameters): Auth0Edge {
+const telemetry = { name: 'nextjs-auth0', version };
```
Widcket, line 32: Should this be moved to some common helper, as it's also used in https://github.com/auth0/nextjs-auth0/blob/141ec7bff57684747e99baddf02840706adda33b/src/index.ts#L26?

src/utils/errors.ts
```diff
@@ -110,7 +108,7 @@ export class AccessTokenError extends AuthError {
 /**
  * @ignore
  */
-export type HandlerErrorCause = Error | AuthError | HttpError;
+export type HandlerErrorCause = Error | AuthError | (Error & { status: number; statusCode: number });
```
Widcket, line 111: Should `Error & { status: number; statusCode: number }` be its own type?

example-app/app/nav.tsx
adamjmcgrath: I was just moving this over from the old example app, I could change them - but I'd rather do that in another PR

src/auth0-session/client/edge-client.ts
adamjmcgrath: None of our configuration errors have a type, they just need to tell the developer they've configured the SDK wrong

src/auth0-session/client/edge-client.ts
adamjmcgrath: This gets caught here https://github.com/auth0/nextjs-auth0/blob/edge-auth-handlers/src/auth0-session/handlers/callback.ts#L61-L79

src/auth0-session/client/edge-client.ts
adamjmcgrath: This get's caught here https://github.com/auth0/nextjs-auth0/blob/edge-auth-handlers/src/handlers/profile.ts#L237

src/auth0-session/client/edge-client.ts
adamjmcgrath: This should throw an `AccessTokenError` will fix

src/auth0-session/client/edge-client.ts
adamjmcgrath: This doesn't throw, if there's an error `oauth.isOAuth2Error(result)` is `true` (which is caught below)

src/auth0-session/client/edge-client.ts
Widcket: In the node handler we're wrapping the errors in our own, more specific error types: https://github.com/auth0/nextjs-auth0/blob/141ec7bff57684747e99baddf02840706adda33b/src/auth0-session/client/node-client.ts#L188

src/auth0-session/handlers/login.ts
adamjmcgrath: This is a configuration error too - we don't have a type for these. We could look at changing that, but it would be a lot of changes for not much/any benefit.

src/edge.ts
adamjmcgrath: Yeah, I can move this to `shared.ts`

src/auth0-session/client/edge-client.ts
adamjmcgrath: Not for `userinfo` - but yeah - fair enough, I'll create a `UserInfoError` and add it to the node and edge clients

src/utils/errors.ts
adamjmcgrath: Sure - I'll make an `HttpError` type

src/auth0-session/client/edge-client.ts
Widcket: Should this error in particular be a little more descriptive then? If I were a developer not versed in identity and got that error, I think I'd be confused.

src/auth0-session/client/edge-client.ts
Widcket: In the node handler we're wrapping the errors in our own, more specific error types:

https://github.com/auth0/nextjs-auth0/blob/141ec7bff57684747e99baddf02840706adda33b/src/auth0-session/client/node-client.ts#L188

tests/auth0-session/handlers/callback.test.ts
```diff
@@ -436,60 +436,6 @@ describe('callback', () => {
     expect(header.alg).toEqual('RS256');
   });
 
-  it('should use client secret jwt on token endpoint', async () => {
```
Widcket, line 439: Apologies if this is something obvious, why are these tests being removed?

tests/fixtures/app-router-helpers.ts
```diff
@@ -1,23 +1,52 @@
...
+const isEdgeRuntime =
+  // @ts-ignore
+  typeof EdgeRuntime !== 'undefined';
+
+export const initAuth0 = (config: ConfigParameters) => {
+  if (isEdgeRuntime) {
+    return edgeInitAuth0(config);
+  } else {
+    return nodeInitAuth0(config);
+  }
```
Widcket, line 32:
```suggestion
  }
return nodeInitAuth0(config);
```

tests/auth0-session/handlers/callback.test.ts
adamjmcgrath: Auth0 doesn't support client secret jwt and the edge client doesn't so I've removed support

src/auth0-session/client/edge-client.ts
```diff
@@ -0,0 +1,241 @@
...
+        signal: AbortSignal.timeout(this.config.httpTimeout),
+        headers
+      };
+    };
+  }
+
+  private async getClient(): Promise<[oauth.AuthorizationServer, oauth.Client]> {
+    if (this.as) {
+      return [this.as, this.client as oauth.Client];
+    }
```
Widcket, line 67: Not sure how long-lived instances in the Edge runtime can be (apparently [there's still some notion](https://news.ycombinator.com/item?id=31912794) of "cold-start"). If these were short-lived, wouldn't we be calling the discovery endpoint a lot?

adamjmcgrath: We can't do any better than keeping them in memory 🤷 (unless Next.js has some built in distributed cache I haven't heard of)

src/auth0-session/client/edge-client.ts
adamjmcgrath: Sure - I can improve the message

src/auth0-session/client/edge-client.ts
Widcket: Could we have an option to not use the discovery endpoint?

tests/handlers/login-page-router.test.ts
```diff
@@ -0,0 +1,331 @@
...
+import { withoutApi, withApi } from '../fixtures/default-settings';
+import { decodeState } from '../../src/auth0-session/utils/encoding';
+import { setup, teardown } from '../fixtures/setup';
+import { get, getCookie } from '../auth0-session/fixtures/helpers';
+import { Cookie, CookieJar } from 'tough-cookie';
+
+describe('login handler (page router)', () => {
+  afterEach(teardown);
+
+  test('should create a state', async () => {
```
Widcket, line 11:
```suggestion
  test('should create a state, nonce, and code verifier', async () => {
```

src/auth0-session/client/edge-client.ts
adamjmcgrath: The question of in memory discovery responses not being persisted is not a new thing we've introduced now that it's running on the Edge runtime. Serverless functions that serve the Node runtime also have the same behaviour, so this SDK has had this behaviour since 0.1.

We could ask users to provide enough configuration to skip the Discovery step, sure (we could also do this in express) - but I don't think this is a feature we need to add just because the client can now be served from Edge Serverless functions as well as Node Serverless functions.

Widcket: Asking because serverless !== edge. As Vercel's architect mentions, "Edge Functions bring another set of trade-offs into the mix".

Widcket: What was enough/serviceable with serverless might not be with edge functions.

Widcket: In any case, this probably warrants more investigation and testing with the edge runtime to find out if something more needs to be done.

Widcket: Since edge functions apparently use Cloudflare Workers, I guess it comes down to the differences between Cloudflare Workers and AWS Lambda.

adamjmcgrath: This SDK is agnostic to a particular edge vendor (Vercel/AWS/Azure/CloudFlare/GCP) - but sure, we can look into how their memory persistence differs between the 2 runtimes. I guess at the very least, Edge functions will be more populous given that their are more Edge locations - so I guess the probability of hitting the same box is less. But on the other hand, the Discovery is served from a CDN, so it's not a huge cost.

src/auth0-session/client/edge-client.ts
adamjmcgrath: `authorizationCodeGrantRequest` doesn't throw an `IdentityProvider` error (one with an `error` / `error_description` that comes from the IDP) it just returns a fetch response (or throws config errors like, "you forgot the `response_type`")

`processAuthorizationCodeOpenIDResponse` returns an IDP error (not throws) which is thrown on L158

src/auth0-session/client/edge-client.ts
adamjmcgrath: See https://github.com/auth0/nextjs-auth0/pull/1269/files/141ec7bff57684747e99baddf02840706adda33b#r1245019748

src/auth0-session/client/edge-client.ts
Widcket: This seems to be missing in https://github.com/auth0/nextjs-auth0/pull/1269/commits/0bb1e4ae418dc5841727bd778ef4af2d47bff586

adamjmcgrath: Ah yeah, was going to say - this is the same as https://github.com/auth0/nextjs-auth0/pull/1269#discussion_r1245019748 the error from the idp get's thrown from `processRefreshTokenResponse`

adamjmcgrath:
Thanks for the review @Widcket 🙏

PR #1427, opened 2023-09-08
Fix customizable handler

jbednar97:
- [x] All new/changed/fixed functionality is covered by tests (or N/A) - It was tested by a human not a written test.
- [x] I have added documentation for all new/changed functionality (or N/A) - It is self explanatory -> there was a missing function call, and incorrectly set up handler

### 📋 Changes

I've added an function call to async login function, which would not work without it.
I've refactored the invite route, because it was using an out of scope property "req". It was refactored to a similar handler that "login" handler uses, because it requires an async function with req, and res as a property passed inside of it.

### 🎯 Testing

To reproduce this issue, simply create a sample Next.js auth0 project with CLI. Run the project with the issued routes, and try to navigate to them in your browser window, the login handler will return an error with stalled request, because the function login isn't called. To reproduce the issue with invite handler, you can simply try to navigate over there in your browser window, and it will return an error because of the out of scope variable, which is basically not defined anywhere. After applying my changes, just navigate to following routes in your browser window, and everything should work as expected.

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -106,13 +106,15 @@ export default handleAuth({
         custom_param: 'custom'
       },
       returnTo: '/custom-page'
-    });
+    })(req,res);
   },
-  invite: handleLogin({
-    authorizationParams: {
-      invitation: req.query.invitation
-    }
-  }),
+  async invite(req,res) {
+    await handleLogin(req,res, {
+        authorizationParams: {
+        invitation: req.query.invitation
+      }
+    })(req,res)
+  },
   'login-with-google': handleLogin({ authorizationParams: { connection: 'google' } }),
   'refresh-profile': handleProfile({ refetch: true }),
   onError(req, res, error) {


EXAMPLES.md
```diff
@@ -106,13 +106,15 @@ export default handleAuth({
         custom_param: 'custom'
       },
       returnTo: '/custom-page'
-    });
+    })(req,res);
```
adamjmcgrath, line 109: This isn't necessary

EXAMPLES.md
```diff
@@ -106,13 +106,15 @@ export default handleAuth({
...
+  },
+  async invite(req,res) {
```
adamjmcgrath, line 111: Thanks for spotting this, could you change to:

```js
invite: handleLogin((req) => ({
  authorizationParams: {
    invitation: req.query.invitation
  }
}))
```

adamjmcgrath (changes requested):

EXAMPLES.md
jbednar97: It didn't run on my Next.js project without calling the function with these parameters, are you 100% sure this isn't necessary ? I've got stalled requests (the function didn't run) when using this without the `(req,res)`

jbednar97: > This isn't necessary

It didn't run on my Next.js project without calling the function with these parameters, are you 100% sure this isn't necessary ? I've got stalled requests (the function didn't run) when using this without the `(req,res)`

adamjmcgrath: @jbednar97 

> are you 100% sure this isn't necessary ?
Yes, I've just confirmed this in the sample app https://github.com/auth0/nextjs-auth0/compare/main...test-handle-login

Could you share a repo that demonstrates your issue?

jbednar97: > @jbednar97
> 
> > are you 100% sure this isn't necessary ?
> 
> Yes, I've just confirmed this in the sample app [main...test-handle-login](https://github.com/auth0/nextjs-auth0/compare/main...test-handle-login)
> 
> Could you share a repo that demonstrates your issue?

It was in a private repo, I'll need to re-create it later tonight if you're willing to wait, I cannot really share the current repo. In the example, you aren't using `pageRouterAuth`, which might indeed fire the function for you, but it doesn't if you use the example code, and navigate to it with a link with href `/api/auth/invite` or something similar

adamjmcgrath: @jbednar97 - `pageRouterAuth.handleLogin` is the same as `handleLogin` - happy to wait for you to share your example

jbednar97: @adamjmcgrath alrighty then, will get back to you, thank you

adamjmcgrath:
Closing due to inactivity - feel free to update and I'll reopen

PR #1489, opened 2023-10-12
Mock Next.js in page router tests

adamjmcgrath:
### 📋 Changes

Mock out the next test app for page router tests - since standing up a next app for every test is too expensive for the CI (didn't work in GH actions and was flakey on CircleCI)

Before:
image

After:
image

### 🎯 Testing

Tested on GH actions https://github.com/auth0/nextjs-auth0/actions/runs/6494889625/job/17638727581


tests/helpers/with-page-auth-required.test.ts
```diff
@@ -170,15 +170,6 @@ describe('with-page-auth-required ssr', () => {
       delete process.env.NEXT_PUBLIC_AUTH0_LOGIN;
     });
 
-    test('is a no-op when invoked as a client-side protection from the server', async () => {
```
adamjmcgrath, line 173: We don't need this test anymore since the same `withPageAuthRequired` import isn't used for SSR and CSR since v3

tests/auth0-session/fixtures/server.ts
```diff
@@ -94,9 +93,10 @@ const createHandlers = (params: ConfigParameters): Handlers => {
   };
 };
 
-const jsonParse = bodyParser.json();
-const parseJson = (req: IncomingMessage, res: ServerResponse): Promise<IncomingMessage> =>
-  new Promise((resolve, reject) => {
+export const parseJson = async (req: IncomingMessage, res: ServerResponse): Promise<IncomingMessage> => {
+  const { default: bodyParser } = await import('body-parser');
```
Widcket, line 97: Just curious, what is the reason we need a dynamic import here?

Widcket (approved):

tests/auth0-session/fixtures/server.ts
adamjmcgrath: I added  a dependency on this file in `tests/fixtures/server.ts` which some test files that run edge tests depend on and `body-parser` can't be imported on the edge runtime (it must use `eval` or something)

PR #1546, opened 2023-11-08
Enhance signup handler example to add code for custom login page

ADTC:
Also fix the heading level.

### 📋 Changes

Enhance signup handler example to add code for custom login page. Also fix the heading level.

### 📎 References

It's similar to what's already here but I chose to use `screen_hint` instead of `action` as the name is compatible with both New and Classic: https://community.auth0.com/t/how-do-i-redirect-users-directly-to-the-hosted-signup-page/42520

### 🎯 Testing

I have tested the implementation in my own project. The instructions should be correct.


ADTC:
Thanks! You have the cleanest list of open issues and PRs in any frequently used project I've seen. Kudos!

PR #1777, opened 2024-10-04
Update auth.ts to await `params` in NextJS 15

ammubhave:
- [N/A] All new/changed/fixed functionality is covered by tests (or N/A)
- [N/A] I have added documentation for all new/changed functionality (or N/A)

### 📋 Changes

In NextJS 15, `params` from routes is now a Promise which must be awaited. Direct access is deprecated and will be removed in future versions.

This change is backwards compatible because awaiting on a non-promise returns back the value itself.

Without this change we get the following warning:
```
In route /api/auth/[auth0] a param property was accessed directly with `params.auth0`. `params` is now a Promise and should be awaited before accessing properties of the underlying params object. In this version of Next.js direct access to param properties is still supported to facilitate migration but in a future version you will be required to await `params`. If this use is inside an async function await it. If this use is inside a synchronous function then convert the function to async or await it from outside this function and pass the result in.
```

### 📎 References

See more details in https://github.com/vercel/next.js/pull/68812

### 🎯 Testing

Add the SDK in NextJS 15 stable and see that the warning is getting generated whenever the auth0 endpoints are hit.

### DIFF
--- a/src/handlers/auth.ts
+++ b/src/handlers/auth.ts
@@ -207,7 +207,7 @@ export default function handlerFactory({
 const appRouteHandlerFactory: (customHandlers: ApiHandlers, onError?: AppRouterOnError) => AppRouteHandlerFn =
   (customHandlers, onError) => async (req: NextRequest, ctx) => {
     const { params } = ctx;
-    let route = params.auth0;
+    let route = (await params).auth0;
 
     if (Array.isArray(route)) {
       let otherRoutes;


Seva98 (commented):
As v15 is officialy out now, why this is still hanging?

dufferzz:
Tried this, now getting

``` Error: Route "/api/auth/[auth0]" used `params.auth0`. `params` should be awaited before using its properties.```

Seemingly from 'appRouteHandlerFactory'

Would be great to use Auth0 with Next 15! Exciting

ammubhave:
was implemented in https://github.com/MineStudio-Host/nextjs-auth0/commit/5f58942f37c4a2918f6fc035a01f742ebe1d7015

PR #2076, opened 2025-04-30
Update middleware combination example to prevent unintended backend execution

tusharpandey13:
## Overview
The existing documentation example in EXAMPLES.md for combining middleware advised copying all headers from the auth0.middleware response (authResponse) to the final response. This includes the x-middleware-next header, which Next.js uses as a signal to always forward the request to the backend application, regardless of the middleware's response status (e.g., 401 Unauthorized).

This behavior leads to a vulnerability where even if custom middleware logic correctly identifies an unauthenticated user and attempts to block the request with a 401 response, the presence of the copied x-middleware-next header causes Next.js to still execute the backend route handler. This could expose page data or allow unintended execution of backend logic for unauthenticated users.

## Changes
This PR updates the EXAMPLES.md file to mitigate this issue:
Added a warning block and code sample.


frederikprijck (dismissed):
Can we add some links to the explaination of this behavior in Next.js, both in the PR and EXAMPLES.md?

nandan-bhat (dismissed):

nandan-bhat (approved):

Issue #1884, opened 2025-02-06
v4: Bring back the ability to pass refresh: true to getAccessToken


dkokotov:
### Checklist

- [x] I have looked into the [Readme](https://github.com/auth0/nextjs-auth0#readme), [Examples](https://github.com/auth0/nextjs-auth0/blob/main/EXAMPLES.md), and [FAQ](https://github.com/auth0/nextjs-auth0/blob/main/FAQ.md) and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://auth0.github.io/nextjs-auth0/) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/nextjs-auth0/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

In v3, when calling `getAccessToken`, it was possible to pass `{ refresh: true }` to force the access token to be refreshed even if it had not yet expired. 

This is useful in contexts when you know that something about the user has been updated in auth0 that would cause the access token to be different, and you would like to retrieve an updated token that reflects the changes. An example would be:
* you use the access token to call an API that checks for certain claims or permissions
* you've updated the user in auth0 via management api to give the new permissions
* now you need to refresh the access token so it reflects those new permissions before calling the API. 

In v4, I am not seeing this option, or any way of replicating this functionality.

### Describe the ideal solution

Re-introduce the optional `{ refresh?: boolean }` parameter to `getAccessToken()` - if set to true (and refresh token exists) it will always request a new token before returning it, even if current token exists and is not expired.

### Alternatives and current workarounds

I haven't found a workaround yet, the relevant code seems to be in `get-access-token.ts` and there is no way to inject the desired behavior.

dkokotov:
As a followup, I found a workaround for the issue as described above. basically the idea is to update the session to make it look like the token has expired, then call `getAccessToken()`. this is tacky but it works:

```
class MyAuth0Client extends Auth0Client {
    async refreshAccessToken(): Promise<{ token: string; expiresAt: number }> {
        const existingSession = await this.getSession();
        if (!existingSession) {
            throw new Error('The user is not authenticated.');
        }

        const sessionToForceTokenRefresh = {
[...]

PR #2189, opened 2025-06-23
feat: update id_token when a new Access Token is fetched

tusharpandey13:
Fixes: #1884 

This change makes the access token methods, `getAccessToken` and `handleAccessToken` also update the ID token claims present in `session.user` when a new Access Token is retrieved. `beofreSessionSaved` hook is called before session data is updated.

### Changes
- `getTokenSet` now also returns updated id_token claims information in the case of a token update using the refresh token.
- both `getAccessToken` and `handleAccessToken` now handle the updated `user` object.
- Added `finalizeSession` method that calls `beforeSessionSaved` hook if present (for custom filtering of id_token claims) else filters the id_token claims using default filtering rules.
- Added flow tests for the same, mocking http layer with msw.
- Added relevant docstrings.

### Tests
Unit tests are _PASS_.
Manual testing was done on a sample app and existing issue was not reproducible.


src/server/client.ts
```diff
@@ -420,20 +419,17 @@ export class Auth0Client {
...
     }
-
+    const { tokenSet, user } = getTokenSetResponse;
     // update the session with the new token set, if necessary
-    if (
-      tokenSet.accessToken !== session.tokenSet.accessToken ||
-      tokenSet.expiresAt !== session.tokenSet.expiresAt ||
-      tokenSet.refreshToken !== session.tokenSet.refreshToken
-    ) {
+    if (user) {
```
frederikprijck: Why is this check changed?

src/server/auth-client.ts
```diff
@@ -771,6 +774,9 @@ export class AuthClient {
           ];
         }
 
+        const idTokenClaims = oauth.getValidatedIdTokenClaims(oauthRes)!;
+        const filteredClaims = filterDefaultIdTokenClaims(idTokenClaims);
```
frederikprijck: We should call the hook.

src/server/auth-client.ts
```diff
@@ -648,17 +651,17 @@ export class AuthClient {
...
       scope: updatedTokenSet.scope,
       expires_at: updatedTokenSet.expiresAt
     });
 
-    if (
-      updatedTokenSet.accessToken !== session.tokenSet.accessToken ||
-      updatedTokenSet.refreshToken !== session.tokenSet.refreshToken ||
-      updatedTokenSet.expiresAt !== session.tokenSet.expiresAt
-    ) {
+    if (user) {
```
frederikprijck: Why is this check changed?

src/server/auth-client.ts
tusharpandey13: added code for hook call

src/server/auth-client.ts
tusharpandey13: this has been reverted

src/server/client.ts
tusharpandey13: this has been reverted

src/server/auth-client.test.ts
tusharpandey13, line 1: authClient.getTokenSet now returns {tokenset, user}
hence updated test

src/server/auth-client.ts
```diff
@@ -656,11 +662,28 @@ export class AuthClient {
...
+          updatedTokenSet.idToken ?? null
+        );
+        finalSession = {
+          ...updatedSession,
+          internal: finalSession.internal
+        };
+      } else {
+        finalSession.user = filterDefaultIdTokenClaims(finalSession.user);
+      }
+
```
tusharpandey13: - If idToken has updated (RT flow), update the new claims in the session.
- Call `beforeSessionSaved` hook if available, else filter claims using default rules.

src/server/auth-client.ts
```diff
@@ -771,6 +794,7 @@ export class AuthClient {
           ];
         }
 
+        const idTokenClaims = oauth.getValidatedIdTokenClaims(oauthRes)!;
```
tusharpandey13, line 784: Since `oauthRes: oauth.TokenEndpointResponse` is available here, parse the id_token claims here.

src/server/auth-client.ts
tusharpandey13, line 1: - `getTokenSet` now returns `tokenSet` and parsed `user` object from id_token claims on success.
- `handleAccessToken` updated to set session if `user` is avaialble from `getTokenSet`
- A new method, `finalizeSession` was added that calls `beforeSessionSaved` hook if supplied else fitlers the ID token claims in `session.user` using the default filtering rules.

src/server/client.ts
tusharpandey13, line 1: `getAccessToken` changed similar to `handleAccessToken` above.

src/server/get-access-token.test.ts
tusharpandey13, line 1: - Generate id token for intial and updated scenarios
- Check that updated id_token claim information is being saved to session in `session.user`.

src/server/auth-client.ts
```diff
@@ -1175,3 +1205,8 @@ const encodeBase64 = (input: string) => {
   }
   return btoa(arr.join(""));
 };
+
+export type GetTokenSetResponse = {
```
guabu: Perhaps we can consider renaming this to `idTokenClaims`?

Also, does this type need to be exported? Is it being used elsewhere?

src/server/client.ts
```diff
@@ -188,6 +188,7 @@ export class Auth0Client {
   private transactionStore: TransactionStore;
   private sessionStore: AbstractSessionStore;
   private authClient: AuthClient;
+  private readonly beforeSessionSaved?: BeforeSessionSavedHook;
```
guabu: IMO, we should avoid relying on this hook in `src/server/client.ts`. The way I look at is mostly to instantiate the `auth-client.ts` and provide the server contract to end-users.

src/server/client.ts
```diff
@@ -420,23 +423,39 @@ export class Auth0Client {
...
+          finalSession,
+          tokenSet.idToken ?? null
+        );
+        finalSession = {
+          ...updatedSession,
+          internal: finalSession.internal
+        };
+      } else {
+        finalSession.user = filterDefaultIdTokenClaims(finalSession.user);
+      }
```
guabu: Similar to the above comment, I believe we should keep this logic in the `auth-client`

guabu (changes requested):
The overall approach is in the right direction — left a few small comments mostly around organization.

src/server/auth-client.ts
```diff
@@ -716,7 +739,7 @@ export class AuthClient {
   async getTokenSet(
     tokenSet: TokenSet,
     forceRefresh?: boolean | undefined
-  ): Promise<[null, TokenSet] | [SdkError, null]> {
+  ): Promise<[null, GetTokenSetResponse] | [SdkError, null]> {
```
guabu, line 729: Let's add some comments to the `getTokenSet` method to be explicit about when we expect the `idTokenClaims` to be returned

src/server/auth-client.ts
frederikprijck: You mean renaming `user` right? Not the entire type?

guabu: That's right, just `user`.

tusharpandey13: Renamed to `idTokenClaims` and removed the redundant export.

src/server/client.ts
tusharpandey13: beforeSessionSaved hook logic was moved to `auth-client.ts` and this was removed.

src/server/client.ts
tusharpandey13: beforeSessionSaved hook logic was moved to `auth-client.ts` and this was removed.

src/server/auth-client.ts
tusharpandey13: Added docstring.

src/server/auth-client.ts
```diff
@@ -1161,6 +1190,41 @@ export class AuthClient {
...
+      );
+      session = {
+        ...updatedSession,
+        internal: session.internal
+      };
+    } else {
+      session.user = filterDefaultIdTokenClaims(session.user);
+    }
+    return session;
+  }
```
tusharpandey13, line 1206: `finalizeSession` was added that calls `beforeSessionSaved` hook if supplied else fitlers the ID token claims in `session.user` using the default filtering rules.

This logic was duplicated at 3 places so moved this to a seperate method. Also, moving to a method was necessary to remove this logic from `client.ts`

src/server/auth-client.ts
```diff
@@ -656,11 +654,20 @@ export class AuthClient {
...
     if (
       updatedTokenSet.accessToken !== session.tokenSet.accessToken ||
-      updatedTokenSet.refreshToken !== session.tokenSet.refreshToken ||
-      updatedTokenSet.expiresAt !== session.tokenSet.expiresAt
+      updatedTokenSet.expiresAt !== session.tokenSet.expiresAt ||
+      updatedTokenSet.refreshToken !== session.tokenSet.refreshToken
     ) {
+      if (user) {
+        session.user = user!;
+      }
```
guabu:
```suggestion
      if (user) {
        session.user = user;
      }
```

src/server/auth-client.ts
```diff
@@ -1175,3 +1239,8 @@ const encodeBase64 = (input: string) => {
   }
   return btoa(arr.join(""));
 };
+
+type GetTokenSetResponse = {
+  tokenSet: TokenSet;
+  idTokenClaims?: User;
```
guabu: Should the `idTokenClaims` be `{ [key: string]: any }`, similar to `filterDefaultIdTokenClaims`?

src/server/auth-client.ts
```diff
@@ -710,13 +717,28 @@ export class AuthClient {
...
+   *   - `[SdkError, null]` if an error occurred (missing refresh token, discovery failure, or refresh failure)
+   *   - `[null, {tokenSet, idTokenClaims}]` if a new token was retrieved, containing the new token set ID token claims
+   *   - `[null, {tokenSet, }]` if token refresh was not done and existing token was returned
+   *
+   * @remarks
+   * - If the access token has expired and no refresh token is available, returns a MISSING_REFRESH_TOKEN error
+   * - If a refresh token is present and the token is expired (or forceRefresh is true), attempts to refresh
+   * - Retrieves and returns id_token claims in the case of RT flow.
+   * - Handles refresh token rotation by updating the refresh token if a new one is provided
+   * - Maintains session lifetime by preserving the original `iat` claim from the existing token set
```
guabu: Q: do you find the `@param` and `@returns` useful here given we use TypeScript exclusively?

Also, I'm not sure the `@remarks` add much context that's valuable to the caller or can be gleaned from the code itself.

Just my 2c, feel free to ignore.

src/server/auth-client.ts
```diff
@@ -1161,6 +1190,41 @@ export class AuthClient {
...
+   * will be applied to remove standard ID token claims from the user object.
+   *
+   * @param session - The session data object containing user information and internal metadata
+   * @param idToken - The raw ID token string from the authentication provider
+   * @returns Promise that resolves when the session has been processed and filtered
+   *
+   * @remarks
+   * This method modifies the session object in place. When using a custom
+   * `beforeSessionSaved` callback, the internal session metadata is preserved
+   * and merged with the updated session data.
```
guabu: Same comment as the above.

src/server/auth-client.ts
tusharpandey13: I think `@returns` conveys useful information in this case.
`@param` and `@remarks` can be removed if needed.

src/server/auth-client.ts
tusharpandey13: I agree, removed `@param` and `@remarks`

src/server/auth-client.ts
tusharpandey13: done

src/server/auth-client.ts
tusharpandey13: removed redundant sections

src/server/auth-client.ts
tusharpandey13: yes, changed to `{ [key: string]: any }`

nandan-bhat (approved):

PR #2207, opened 2025-07-02
feat: add `withPageAuthRequired` for server

guabu:
### 📋 Changes

Adds `withPageAuthRequired` for server-rendered components.

### 📎 References

Fixes: https://github.com/auth0/nextjs-auth0/issues/1790

### 🎯 Testing

See `EXAMPLES.md` for testable examples using the App and Pages router.


guabu:
Thanks for taking a look @tusharpandey13. Since the SDK is the one returning the `redirect` directive from `getServerSideProps`, there shouldn't ever be a case where the `notFound` directive is returned unless we return it explicitly (which we don't).

tusharpandey13 (dismissed):
lgtm

arpit-jn (approved):

Issue #1895, opened 2025-02-10
v4: Customize auth handlers


mvvmm:
### Checklist

- [x] I have looked into the [Readme](https://github.com/auth0/nextjs-auth0#readme), [Examples](https://github.com/auth0/nextjs-auth0/blob/main/EXAMPLES.md), and [FAQ](https://github.com/auth0/nextjs-auth0/blob/main/FAQ.md) and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://auth0.github.io/nextjs-auth0/) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/nextjs-auth0/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

In v3, we could customize the auth handlers in `/api/auth/[auth0]` via something like:

```
export const GET = handleAuth({
  async logout(req: NextApiRequest, res: NextApiResponse) {

   // do custom stuff

   return await handleLogout(req, res);
  },
});
```
Reference: 
- auth0.github.io/nextjs-auth0/types/handlers_logout.HandleLogout.html
- https://auth0.github.io/nextjs-auth0/modules/handlers.html

I'm not finding that something similar exists in v4.

### Describe the ideal solution

Ability to customize the auth handlers in v4. 

The `handleLogout()` helper is an important piece of this. We need the ability to write our own custom code and then rely on the default auth handler as a fallback, or in addition to our custom code.

### Alternatives and current workarounds

Depends on the use case, but as far as I can tell, there is no way to run custom code in the middle of those new auth handlers,  `/auth/logout`, for example.

priley86:
+1'ing this request here. As an aside, if you are simply looking to update session post `/auth/callback`, you may want to look at the `beforeSessionSaved` and `onCallback` hooks. These ended up fitting my usecase ok here for the moment.
https://github.com/auth0/nextjs-auth0/blob/47e45ee9ea46b4670fed2566a9caa17b2d6fbc02/EXAMPLES.md#hooks

maninderpreetsingh:
It's also not taking the basePath from next.config.

priley86:
[...]

PR #2218, opened 2025-07-04
feat: Add customize auth handlers docs

tusharpandey13:
Fixes: https://github.com/auth0/nextjs-auth0/issues/1895

Adds documentation about customizing auth handlers at various parts of authentication:
- Before / after login
- Before / after logout
- Before callback handler

Includes example and prose


tusharpandey13:
@guabu does this look right?

nandan-bhat (dismissed):

EXAMPLES.md
```diff
@@ -1265,3 +1268,65 @@ export async function middleware(request: NextRequest) {
   return resWithCombinedHeaders;
 }
 ```
+
+## Customizing Auth Handlers
+
+In v4, authentication routes (`/auth/login`, `/auth/logout`, `/auth/callback`) are handled automatically by the middleware. While you can no longer customize individual route handlers directly like in v3, you can intercept these routes in your middleware to run custom logic before the auth handlers execute.
```
guabu: Do we need to mention older versions in the README? It feels more appropriate to leave those to the migration guide where it's more relevant.

EXAMPLES.md
```diff
@@ -1265,3 +1268,65 @@ export async function middleware(request: NextRequest) {
...
+- Modify the response (set cookies, headers, etc.)
+- Implement custom redirects or early returns when needed
+- Add business logic around authentication flows
+- Maintain compatibility with existing tracking and analytics systems
+
+The middleware-based approach provides the same level of control as v3's custom handlers while working seamlessly with v4's automatic route handling.
+
+### Run custom code before Auth Handlers
+
+Following example shows how to run custom logic before the `logout` handler:
```
guabu: This would actually be running after the logout handler but before the response is sent to the user

EXAMPLES.md
```diff
@@ -1265,3 +1268,65 @@ export async function middleware(request: NextRequest) {
...
+
+    // prepare NextResponse object from auth0 middleware
+    const authRes = await auth0.middleware(request);
+
+    // The following interceptUrls can be used:
+    //    "/auth/login" : intercept login auth handler
+    //    "/auth/logout" : intercept logout auth handler
+    //    "/auth/callback" : intercept callback auth handler
+    //    "/your/login/returnTo/url" : intercept redirect after login
+    //    "/your/logout/returnTo/url" : intercept redirect after logout
```
guabu: Where do these come from?

EXAMPLES.md
```diff
@@ -1265,3 +1268,65 @@ export async function middleware(request: NextRequest) {
...
+            value: 'anotherValue',
+            httpOnly: true,
+            path: '/',
+        });
+
+        // Example: Delete a cookie
+        // authRes.cookies.delete('cookieNameToDelete');
+
+        // you can also do an early return here with your own NextResponse object
+        // return NextResponse.redirect(new URL('/custom-logout-page'));
```
guabu, line 1322: nit (feel free to ignore): I would simplify the example to keep it concise

EXAMPLES.md
tusharpandey13: login `returnTo` and logout `returnTo`.

EXAMPLES.md
tusharpandey13: IMO one concise exhaustive example is easier to use than 5 different variations of the same code.

EXAMPLES.md
tusharpandey13: ACK

EXAMPLES.md
tusharpandey13: ACK, made the change

tusharpandey13:
Moving ahead with this.

PR #2221, opened 2025-07-08
bugfix/delete redundant txn cookies

tusharpandey13:
- **Release v4.8.0 (#2217)**
- **remove redundat txn cookies after login and callback calls**


tusharpandey13:
this is not relevant

Issue #2219, opened 2025-07-07
Headers.append: xxx is an invalid header value


brunoscopelliti:
### Checklist

- [x] The issue can be reproduced in the [nextjs-auth0 sample app](https://github.com/auth0-samples/auth0-nextjs-samples/tree/main/Sample-01) (or N/A).
- [x] I have looked into the [Readme](https://github.com/auth0/nextjs-auth0#readme), [Examples](https://github.com/auth0/nextjs-auth0/blob/main/EXAMPLES.md), and [FAQ](https://github.com/auth0/nextjs-auth0/blob/main/FAQ.md) and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://auth0.github.io/nextjs-auth0/) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/nextjs-auth0/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

I've reproduced the problem also with [nextjs-auth0 sample app](https://github.com/auth0-samples/auth0-nextjs-samples/tree/main/Sample-01).

Add a new endpoint, that uses `auth0.updateSession` to update session.

```js
export const PATCH = async function route(request) {
  try {
    const session = await auth0.getSession();

    const response = NextResponse.json({
      data: session.user,
      status: 'OK'
    });

    await auth0.updateSession(request, response, { ...session, user: { ...session.user, updated: true } });

    return response;
  } catch (error) {
    return NextResponse.json({ error: error.message }, { status: error.status || 500 });
  }
};
```

When I call this endpoint I get this error:

```
'Headers.append: "append(name, value) {
        webidl.brandCheck(this, _Headers);
        webidl.argumentLengthCheck(arguments, 2, "Headers.append");
        const prefix = "Headers.append";
        name = webidl.converters.ByteString(name, prefix, "name");
        value = webidl.converters.ByteString(value, prefix, "value");
        return appendHeader(this, name, value);
      }" is an invalid header value.'

```

The error is triggered in this function:

```js
createRequestCookies(req) {
    const headers = new Headers();
    for (const key in req.headers) {
        if (Array.isArray(req.headers[key])) {
            for (const value of req.headers[key]) {
[...]

PR #2225, opened 2025-07-13
bugfix: ignore enumerable functions when copying headers in Pages-router

tusharpandey13:
Updates `createRequestCookies` to avoid treating helper functions on `req.headers` as header values, preventing runtime crashes on Node 20+.

- [ ] I have added documentation for all new/changed functionality (N/A)

### 🔍 RCA
`for…in` over `req.headers` also enumerated prototype helper methods (`append`, `delete`, …) starting Node 20. Passing these functions to the WHATWG `Headers` API raised `TypeError: <fn> is an invalid header value` when `auth0.updateSession()` ran in a Pages-router route.

### 📋 Changes
Changed `nextjs-auth0/src/server/client.ts`: iterate over `Object.entries(req.headers)` and copy only `string|string[]` values; updated docstring
Added `nextjs-auth0/tests/createRequestCookies-header-copy.test.ts`: regression tests that reproduce the crash and assert correct cookie parsing

### 📎 References
- Fixes: auth0/nextjs-auth0#2219

### 🎯 Testing
Automated:
- New Vitest spec `tests/createRequestCookies-header-copy.test.ts` fails on pre-patch code, passes now.
- `pnpm run test:coverage` shows all 248 tests passing with > 84 % coverage.

Manual:
1. Build SDK: `pnpm run build`
2. Integrate via `yalc` into sample app on Node 22
3. Hit endpoint that calls `auth0.updateSession`
4. Confirm no `Headers.append` error and session updates as expected.

### DIFF
--- a/src/server/auth-client.test.ts
+++ b/src/server/auth-client.test.ts
@@ -520,10 +520,10 @@ ca/T0LLtgmbMmxSv/MmzIg==
         // assert session has been updated
         const updatedSessionCookie = response.cookies.get("__session");
         expect(updatedSessionCookie).toBeDefined();
-        const { payload: updatedSessionCookieValue } = await decrypt(
+        const { payload: updatedSessionCookieValue } = (await decrypt(
           updatedSessionCookie!.value,
           secret
-        );
+        )) as jose.JWTDecryptResult;
         expect(updatedSessionCookieValue).toEqual(
           expect.objectContaining({
             user: {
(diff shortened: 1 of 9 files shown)


src/server/client.ts
```diff
@@ -682,17 +682,35 @@ export class Auth0Client {
...
-        for (const value of req.headers[key]) {
-          headers.append(key, value);
+    for (const [key, val] of Object.entries(req.headers)) {
+      if (val == null) {
+        continue; // skip undefined/null
+      }
+
+      if (Array.isArray(val)) {
+        for (const v of val) {
+          headers.append(key, v);
```
nandan-bhat, line 708: Should we check if `v` is not a string here ?

PR #2230, opened 2025-07-14
feat: add support for `withApiAuthRequired` helper

guabu:
### 📋 Changes

Adds support for `withApiAuthRequired` helper.

### 📎 References

Fixes: https://github.com/auth0/nextjs-auth0/issues/1790

### 🎯 Testing

Set up an API route as documented in `EXAMPLES.md` to observe the 401 when unauthenticated and the successful 200 response when a session exists.

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -92,9 +92,9 @@ By default, the SDK uses OpenID Connect's RP-Initiated Logout when available, fa
 
 ```ts
 export const auth0 = new Auth0Client({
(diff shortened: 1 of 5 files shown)


mickaeltr:
Hello, I couldn't find out how to use `withApiAuthRequired` as it does not seem to be available for import from `@auth0/nextjs-auth0` or else (like any other function). Perhaps I misunderstood or something is missing here.

Issue #2237, opened 2025-07-18
Logout fails to clear cookies when AUTH0_COOKIE_DOMAIN is set, causing immediate re-authentication


getinnocuous:
### Checklist

- [x] The issue can be reproduced in the [nextjs-auth0 sample app](https://github.com/auth0-samples/auth0-nextjs-samples/tree/main/Sample-01) (or N/A).
- [x] I have looked into the [Readme](https://github.com/auth0/nextjs-auth0#readme), [Examples](https://github.com/auth0/nextjs-auth0/blob/main/EXAMPLES.md), and [FAQ](https://github.com/auth0/nextjs-auth0/blob/main/FAQ.md) and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://auth0.github.io/nextjs-auth0/) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/nextjs-auth0/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

## Describe the problem

When `AUTH0_COOKIE_DOMAIN` is set to enable cross-subdomain authentication (required for API calls), the logout process fails to properly clear Auth0 cookies from the parent domain, causing users to be immediately re-authenticated after logout.

## What was the expected behavior?

Logout should clear all Auth0 cookies regardless of the cookie domain setting, preventing automatic re-authentication.

**Configuration:**
```javascript
export const auth0 = new Auth0Client({
  authorizationParameters: {
    scope: env.AUTH0_SCOPE,
    audience: env.AUTH0_AUDIENCE,
    redirect_uri: `${URL}auth/callback`,
  },
  secret: env.AUTH0_SECRET,
  domain: env.AUTH0_DOMAIN,
  appBaseUrl: URL, // https://localhost.mydomain.com:3000/
  clientId: env.AUTH0_CLIENT_ID,
  clientSecret: env.AUTH0_CLIENT_SECRET,
})
```
```
AUTH0_COOKIE_DOMAIN="df.mydomain.com"  # Required for API authentication
AUTH0_DOMAIN="auth.mydomain.com"
AUTH0_BASE_URL="https://localhost.df.mydomain.com:3000/"
```

### Reproduction

**Steps:**
[...]

PR #2250, opened 2025-07-30
# bugfix: Logout fails to clear cookies when AUTH0_COOKIE_DOMAIN is set (#2237)

tusharpandey13:
Fixed logout functionality when using `AUTH0_COOKIE_DOMAIN` by ensuring cookie deletion includes the correct `path` parameter. This resolves immediate re-authentication after logout when custom cookie domains are configured.

### 🔍 RCA

The `deleteCookie()` function in `src/server/cookies.ts` was ignoring the optional `path` parameter passed through `options`. During logout, session cookies are written with an explicit `path` (defaults to `/`), but deletion attempts only included `domain` without `path`, causing browsers to treat the missing path as the current request path (e.g., `/auth/logout`) which doesn't match the original cookie's `Path=/`.

### 📋 Changes

This fix ensures proper cookie cleanup when `AUTH0_COOKIE_DOMAIN` is configured, preventing users from being immediately re-authenticated after logout.

Changed `src/server/cookies.ts`: Added `path` parameter handling to `deleteCookie()` function
Added `src/server/logout-cookie-domain-fix.test.ts`: Comprehensive test for session cookie deletion with domain configuration  
Added `src/server/cookie-domain-deletion.test.ts`: Unit tests for `deleteCookie()` and `deleteChunkedCookie()` functions
Changed `src/server/session/stateless-session-store.test.ts`: Updated existing tests to expect correct `path` parameter in cookie deletion

### 📎 References

Fixes: #2237

### 🎯 Testing

**Automated:**
- Added integration test `logout-cookie-domain-fix.test.ts` that verifies session deletion includes both `Domain` and `Path` attributes
- Added unit tests in `cookie-domain-deletion.test.ts` for `deleteCookie()` and `deleteChunkedCookie()` functions  
- Updated existing tests to expect the corrected behavior with `path` parameter

**Manual:**
1. Configure `AUTH0_COOKIE_DOMAIN=".example.com"` in sample app
2. Log in and verify session cookie is set with `Domain=.example.com; Path=/`
3. Navigate to `/auth/logout`
4. Verify cookie deletion header includes both `Domain=.example.com` and `Path=/`
5. Confirm user is not immediately re-authenticated after logout

### DIFF
--- a/src/server/cookies.ts
+++ b/src/server/cookies.ts
@@ -353,6 +353,13 @@ export function addCacheControlHeadersForSession(res: NextResponse): void {
   res.headers.set("Expires", "0");
 }
 
+/**
+ * Deletes a cookie from the response with optional domain and path specifications.
+ *
+ * @param resCookies - The response cookies object to manipulate.
+ * @param name - The name of the cookie to delete.
+ * @param options - Optional domain and path settings for cookie deletion.
+ */
 export function deleteCookie(
   resCookies: ResponseCookies,
   name: string,
@@ -366,7 +373,9 @@ export function deleteCookie(
     deleteOptions.domain = options.domain;
   }
 
-
+  if (options?.path) {
+    deleteOptions.path = options.path;
+  }
 
   resCookies.set(name, "", deleteOptions);
 }
--- a/src/server/cookies.ts
+++ b/src/server/cookies.ts
@@ -218,6 +218,7 @@ export function setChunkedCookie(
     reqCookies.set(name, value);
 
     // When we are writing a non-chunked cookie, we should remove the chunked cookies
+    // Remove any previously stored chunks for this cookie name
     getAllChunkedCookies(reqCookies, name).forEach((cookieChunk) => {
       deleteCookie(resCookies, cookieChunk.name);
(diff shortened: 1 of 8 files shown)


src/server/cookies.ts
```diff
@@ -350,8 +353,29 @@ export function addCacheControlHeadersForSession(res: NextResponse): void {
...
+
+  if (options?.domain) {
+    deleteOptions.domain = options.domain;
+  }
+
+  if (options?.path) {
+    deleteOptions.path = options.path;
+  }
+
+  resCookies.set(name, "", deleteOptions);
```
tusharpandey13: main change of this PR. use domain passed in parameters.

package.json
```diff
@@ -6,8 +6,8 @@
   "scripts": {
     "build": "tsc",
     "build:watch": "tsc -w",
-    "test:unit": "vitest",
-    "test:coverage": "vitest run --coverage",
+    "test:unit": "vitest --run",
+    "test:coverage": "vitest run --run --coverage",
```
tusharpandey13, line 10: Make vitest not stop for user input after test run. This was considerably slowing down llm workflows.

PR #2261, opened 2025-08-07
feat: add support for backchannel authentication

guabu:
### 📋 Changes

Adds support for Client-Initiated Backchannel Authentication. A new method `getTokenByBackchannelAuth` is exposed on the client that enables developers to initiate a backchannel authentication request and poll the token endpoint until it's complete.

The `openid-client` was added to handle the polling and retry logic without having to re-implement it ourselves using `oauth4webapi` and to start moving the implementation closer to that of `auth0-auth-js`.

### 🎯 Testing

1. Call the `getTokenByBackchannelAuth` method from a server route/action
2. Authorize or reject the authorization via a Guardian app
3. The SDK will return the token set or error accordingly


adamjmcgrath:
> The openid-client was added to handle the polling and retry logic without having to re-implement it ourselves using oauth4webapi and to start moving the implementation closer to that of auth0-auth-js.

Would it make more sense to use auth0-auth-js for CIBA now if that's what the sdk is going to end up using?

src/server/auth-client.ts
```diff
@@ -863,6 +869,99 @@ export class AuthClient {
...
+      ];
+    } catch (e: any) {
+      return [
+        new BackchannelAuthenticationError({
+          cause: new OAuth2Error({
+            code: e.error,
+            message: e.error_description
+          })
+        }),
+        null
```
tusharpandey13, line 970: > Errors that come from Auth0 in the redirect_uri callback may contain reflected user input via the OpenID Connect error and error_description query parameter. You should not render the error message, or error and error_description properties without properly escaping them first.

See https://github.com/auth0/nextjs-auth0/blob/b685e5c444e4cb82ee9372851e698bc3ed12c8b7/src/errors/index.ts#L7

tusharpandey13: Is passing `e.error_description` safe here?

src/server/client.ts
```diff
@@ -735,6 +736,25 @@ export class Auth0Client {
...
 
+  /**
+   * Authenticates using Client-Initiated Backchannel Authentication and returns the token set and optionally the ID token claims and authorization details.
+   *
+   * This method will initialize the backchannel authentication process with Auth0, and poll the token endpoint until the authentication is complete.
+   *
+   * Using Client-Initiated Backchannel Authentication requires the feature to be enabled in the Auth0 dashboard.
+   * @see https://auth0.com/docs/get-started/authentication-and-authorization-flow/client-initiated-backchannel-authentication-flow
+   */
+  async getTokenByBackchannelAuth(options: BackchannelAuthenticationOptions) {
```
tusharpandey13, line 747: > Authenticates using Client-Initiated Backchannel Authentication and returns the token set and optionally the ID token claims and authorization details.

Should this be named as `authenticateWithBackchannel` instead?

src/server/auth-client.ts
guabu: The `e.error_description` would not be rendered — the caller would have to do that explicitly (and the same caveats would hold).

src/server/client.ts
guabu: We'd like to make it explicit that the method is only going to return a token and not establish/update a method

guabu:
Thanks for the reviews!

> Do we think that exporting openid-client polling configuration can be useful here instead of hardcoding the default config?
https://github.com/panva/openid-client/blob/main/docs/interfaces/BackchannelAuthenticationGrantPollOptions.md

We might want to consider that in the future if/when the use case arises but for the time being I would say it's better to keep the surface area of the API limited.

> Would it make more sense to use auth0-auth-js for CIBA now if that's what the sdk is going to end up using?

We chatted about this and decided to rely on openid-client directly since that's what auth0-auth-js uses under the hood as well.

src/server/client.ts
tusharpandey13: Yeah, but instead of getting a token for an already authenticated session, this method can potentially also establish said session on server (authenticate) and then also return the token via polling. I think the authenticate verb here is important to highlight.

Maybe something like `authenticateBackchannelAndGetToken`?
Open to a discussion on this.

PR #2313, opened 2025-09-10
feat: add federated logout

tusharpandey13:


src/server/auth-client.ts
```diff
@@ -461,6 +467,10 @@ export class AuthClient {
         url.searchParams.set("id_token_hint", session.tokenSet.idToken);
       }
 
+      if (federated) {
+        url.searchParams.set("federated", "");
```
nandan-bhat, line 472: This would set `?federated=`. Is (`=` at the end) it expected ?

gyaneshgouraw:
@claude

src/server/auth-client.ts
tusharpandey13: Please look at https://auth0.com/docs/authenticate/login/logout/log-users-out-of-idps#alternative-logout.
It does not mention if a value should be passed.
https://community.auth0.com/t/federated-logout/124280/3 talks about whether a boolean should be passed here or not.

nandan-bhat: Sounds good ! Thanks !

nandan-bhat (approved):

PR #2333, opened 2025-09-22
Add support for access tokens with difference audiences (MRRT)

frederikprijck:

### 📋 Changes

This PR adds support to pass `audience` and `scope` to the `getAccessToken` method.

There are still some things to complete before being ready for review:

- [ ] Add support for Audience and Scope in `/auth/access-token`

### 📎 References

N/A

### 🎯 Testing

These changes require [MRRT](https://auth0.com/docs/secure/tokens/refresh-tokens/multi-resource-refresh-token/configure-and-implement-multi-resource-refresh-token) to be configured accordingly, so that the RefreshToken can be re-used to request tokens from different audiences.

### DIFF
--- a/src/server/auth-client.ts
+++ b/src/server/auth-client.ts
@@ -41,6 +41,7 @@ import {
   normalizeWithBasePath,
   removeTrailingSlash
 } from "../utils/pathUtils.js";
+import { findAccessTokenSet } from "../utils/token-set-helpers.js";
 import { toSafeRedirect } from "../utils/url-helpers.js";
 import { addCacheControlHeadersForSession } from "./cookies.js";
 import { AbstractSessionStore } from "./session/abstract-session-store.js";
@@ -693,9 +694,7 @@ export class AuthClient {
       );
     }
 
-    const [error, getTokenSetResponse] = await this.getTokenSet(
-      session.tokenSet
-    );
+    const [error, getTokenSetResponse] = await this.getTokenSet(session);
 
     if (error) {
       return NextResponse.json(
@@ -784,6 +783,59 @@ export class AuthClient {
   }
 
   /**
+   * Retrieves the token set from the session data, considering optional audience and scope parameters.
+   * When audience and scope are provided, it checks if they match the global ones defined in the authorization parameters.
+   * If they match, it returns the top-level token set from the session data.
+   * If they don't match, it searches for a corresponding access token in the session's `accessTokens` array.
+   * @param sessionData The session data containing the token sets.
+   * @param options Optional parameters for audience and scope to filter the token set.
+   * @returns A partial token set that matches the provided audience and scope, or the top-level token set if they match the global ones.
+   */
+  #getTokenSetFromSession(
+    session: SessionData,
+    options: { scope?: string | null; audience?: string | null }
+  ): Partial<TokenSet> {
+    let tokenSet: Partial<TokenSet> = session.tokenSet;
+
+    // When audience and scope are provided, we need to compare them with the original ones provided in the Auth0Client constructor.
+    // When they are identical, we should read from the top-level `SessionData.tokenSet`.
+    // If not, we should look for the corresponding access token in `SessionData.accessTokens`
+    const isAudienceTheGlobalAudience =
+      !options.audience ||
+      options.audience === this.authorizationParameters.audience;
+    const isScopeTheGlobalScope =
+      !options.scope || options.scope === this.authorizationParameters.scope;
+
+
+    if (isAudienceTheGlobalAudience && isScopeTheGlobalScope) {
+      return tokenSet;
+    }
+
+    const audience = options.audience ?? this.authorizationParameters.audience;
+    const scope =
+      options.scope ?? this.authorizationParameters.scope ?? undefined;
+
+    // If there is no audience, we cannot find the correct access token in the array
+    if (!audience) {
+      return {
+        ...tokenSet,
+        accessToken: undefined,
+        expiresAt: undefined,
+        scope: undefined
+      };
+    }
+
+    const accessTokenSet = findAccessTokenSet(session, { scope, audience });
+
+    return {
+      ...tokenSet,
+      accessToken: accessTokenSet?.accessToken,
+      expiresAt: accessTokenSet?.expiresAt,
+      scope: accessTokenSet?.scope,
+      audience: accessTokenSet?.audience
+    };
+  }
+  /**
    * Retrieves OAuth token sets, handling token refresh when necessary or if forced.
    *
    * @returns A tuple containing either:
@@ -792,11 +844,39 @@ export class AuthClient {
    *   - `[null, {tokenSet, }]` if token refresh was not done and existing token was returned
    */
   async getTokenSet(
-    tokenSet: TokenSet,
-    forceRefresh?: boolean | undefined
+    sessionData: SessionData,
+    options: {
+      refresh?: boolean | undefined;
+      scope?: string;
+      audience?: string;
+    } = {}
   ): Promise<[null, GetTokenSetResponse] | [SdkError, null]> {
-    // the access token has expired but we do not have a refresh token
-    if (!tokenSet.refreshToken && tokenSet.expiresAt <= Date.now() / 1000) {
+    const tokenSet: Partial<TokenSet> = this.#getTokenSetFromSession(
+      sessionData,
+      {
+        scope: options.scope,
+        audience: options.audience
+      }
+    );
+
+    // no access token was found that matches the, optional, provided audience and scope
+    if (!tokenSet.refreshToken && !tokenSet.accessToken) {
+      return [
+        new AccessTokenError(
+          AccessTokenErrorCode.MISSING_REFRESH_TOKEN,
+          "No access token found and a refresh token was not provided. The user needs to re-authenticate."
+        ),
+        null
+      ];
+    }
+
+    // the access token was found, but it has expired and we do not have a refresh token
+    if (
+      !tokenSet.refreshToken &&
+      tokenSet.accessToken &&
+      tokenSet.expiresAt &&
+      tokenSet.expiresAt <= Date.now() / 1000
+    ) {
       return [
         new AccessTokenError(
           AccessTokenErrorCode.MISSING_REFRESH_TOKEN,
@@ -808,12 +888,26 @@ export class AuthClient {
 
     if (tokenSet.refreshToken) {
       // either the access token has expired or we are forcing a refresh
-      if (forceRefresh || tokenSet.expiresAt <= Date.now() / 1000) {
+      if (
+        options.refresh ||
+        !tokenSet.expiresAt ||
+        tokenSet.expiresAt <= Date.now() / 1000
+      ) {
         const [discoveryError, authorizationServerMetadata] =
           await this.discoverAuthorizationServerMetadata();
 
         if (discoveryError) {
           return [discoveryError, null];
+        }
+
+        const additionalParameters = new URLSearchParams();
+
+        if (options.scope) {
+          additionalParameters.append("scope", options.scope);
+        }
+
+        if (options.audience) {
+          additionalParameters.append("audience", options.audience);
         }
 
         const refreshTokenRes = await oauth.refreshTokenGrantRequest(
@@ -824,7 +918,8 @@ export class AuthClient {
           {
             ...this.httpOptions(),
             [oauth.customFetch]: this.fetch,
-            [oauth.allowInsecureRequests]: this.allowInsecureRequests
+            [oauth.allowInsecureRequests]: this.allowInsecureRequests,
+            additionalParameters
           }
         );
 
@@ -878,7 +973,7 @@ export class AuthClient {
       }
     }
 
-    return [null, { tokenSet, idTokenClaims: undefined }];
+    return [null, { tokenSet: tokenSet as TokenSet, idTokenClaims: undefined }];
   }
 
   async backchannelAuthentication(
--- a/src/server/client.ts
+++ b/src/server/client.ts
@@ -17,9 +17,14 @@ import {
   SessionData,
   SessionDataStore,
   StartInteractiveLoginOptions,
+  TokenSet,
   User
 } from "../types/index.js";
 import { isRequest } from "../utils/request.js";
+import {
+  accessTokenSetFromTokenSet,
+  findAccessTokenSet
+} from "../utils/token-set-helpers.js";
 import {
   AuthClient,
   BeforeSessionSavedHook,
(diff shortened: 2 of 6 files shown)


frederikprijck:
@claude

src/server/client.ts
```diff
@@ -434,7 +437,7 @@ export class Auth0Client {
     arg2?: PagesRouterResponse | NextResponse,
     arg3?: GetAccessTokenOptions
   ): Promise<{ token: string; expiresAt: number; scope?: string }> {
-    const defaultOptions: Required<GetAccessTokenOptions> = {
+    const defaultOptions: GetAccessTokenOptions = {
```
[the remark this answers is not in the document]
frederikprijck: This looks fine.

src/server/client.ts
```diff
@@ -20,6 +20,7 @@ import {
   User
 } from "../types/index.js";
 import { isRequest } from "../utils/request.js";
+import { getSessionChangesAfterGetAccessToken } from "../utils/session-changes-helpers.js";
```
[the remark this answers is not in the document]
frederikprijck: This is used.

src/utils/session-changes-helpers.ts
```diff
@@ -0,0 +1,105 @@
...
+      accessTokens: [
+        ...(session.accessTokens || []),
+        accessTokenSetFromTokenSet(tokenSet, { scope, audience })
+      ]
+    };
+  } else {
+    if (
+      tokenSet.accessToken !== existingAccessTokenSet.accessToken ||
+      tokenSet.expiresAt !== existingAccessTokenSet.expiresAt ||
+      tokenSet.refreshToken !== session.tokenSet.refreshToken
```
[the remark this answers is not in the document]
frederikprijck: This looks fine, the suggested change has the same before and after.

src/utils/token-set-helpers.ts
```diff
@@ -0,0 +1,157 @@
...
+    );
+  });
+
+  // If no potential matches were found, we can exit early.
+  if (allMatches.length === 0) {
+    return;
+  }
+
+  // 3. Sort the valid matches to find the best one.
+  // The best match is the one with the smallest scope array, as it has the fewest
```
nandan-bhat, line 173: What if the `scope` is set as `'read write write'` (with duplicates)?  
Since we use `.length` here to decide the "best" match, duplicates inflate the count and could affect sorting.

src/utils/token-set-helpers.ts
```diff
@@ -0,0 +1,157 @@
...
+  if (scopes === requiredScopes) {
+    return true;
+  }
+
+  if (!scopes || !requiredScopes) {
+    return false;
+  }
+
+  const scopesSet = new Set(parseScopesToArray(scopes));
+  const requiredScopesArray = parseScopesToArray(requiredScopes);
```
nandan-bhat: I think we should get rid of duplicates before comparing. Better to make `requiredScopeArray` a set.

src/utils/token-set-helpers.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  // 1. When there are no access tokens, we can exit early.
+  if (!accessTokenSets || accessTokenSets.length === 0) {
+    return;
+  }
+
+  // 2. Filter the list to find all AccessTokenSet's that are valid matches.
+  // A valid match's audience must match the provided `options.audience`,
+  // and its scope must contain all items from `options.scope`.
+  const allMatches = accessTokenSets.filter((accessTokenSet) => {
```
nandan-bhat, line 154: Can multiple tokens exist for a single `audience`? From the current implementation it looks like we can store more than one. I’ll try to dig into the use cases, but I had a couple of observations:

1. As I understand it, calling `getAccessToken('resource-1', 'read')` followed by `getAccessToken('resource-2', 'write')` would create and store two tokens in the session. Is that expected/required?
2. I’m also wondering how this impacts the `__session` cookie if we end up with multiple ATs per audience, especially when this scales across several audiences.

Would it make sense to instead store a single AT per audience with the broader set of scopes, and reuse that for subset requests?

src/server/auth-client.ts
```diff
@@ -857,6 +982,15 @@ export class AuthClient {
...
           idToken: oauthRes.id_token,
+          // We store the requested scopes on the tokenSet, so we know what scopes were requested.
+          // We do not store the returned scopes from the `oauthRes` object.
+          // The server may return less scopes than requested.
+          // This ensures we can return the same token again when a token for the same or less scopes is requested.
+          //
+          // E.g. When requesting a token with scope `a b`, and we return one for scope `a` only,
+          // - If we store the returned scopes, we cannot return this token when the user requests a token for scope `a b` again.
+          // - If we store the requested scopes, we can return this token when the user requests a token for scope `a` or `a b` again.
+          scope: tokenSet.scope ?? scope,
```
nandan-bhat: I see we’re saving the requested scopes here instead of the scopes the server actually returned.

This makes sense for caching (so a `scope=a b` request can reuse the token even if only `a` was granted), but it also means the session may show more scopes than the token really has. 

Might be good to mention in the docs that apps should always check the token itself (or the API) to know which scopes were really granted.

src/server/auth-client.ts
```diff
@@ -783,6 +806,57 @@ export class AuthClient {
...
+
+    // When audience and scope are provided, we need to compare them with the original ones provided in the Auth0Client constructor.
+    // When they are identical, we should read from the top-level `SessionData.tokenSet`.
+    // If not, we should look for the corresponding access token in `SessionData.accessTokens`
+    const isAudienceTheGlobalAudience =
+      !options.audience ||
+      options.audience === this.authorizationParameters.audience;
+    const isScopeTheGlobalScope =
+      !options.scope ||
+      options.scope ===
```
nandan-bhat: Should we normalize the scope comparison here ?
For example, if the global scope is `"read write"` but the requested scope is `"write read"`, this check would treat them as different even though they’re semantically identical.

src/utils/token-set-helpers.ts
```diff
@@ -0,0 +1,157 @@
...
+ * @param scopes1 The first scope string
+ * @param scopes2 The second scope string
+ * @returns Merged scope string with unique scopes
+ */
+export function mergeScopes(
+  scopes1: string | undefined | null,
+  scopes2: string | undefined | null
+): string {
+  return [
+    ...(scopes1 ? scopes1.split(" ") : []),
```
nandan-bhat: Right now `mergeScopes()` unions the `global` and `requested scopes`, so `global="read write"` + `requested="read"` results in `"read write"`. Is this intentional ?

This means `getAccessToken()` can only ever request the same or broader scopes, but not fewer. Should we allow overriding the global scopes instead of always combining them ?

If the goal here is to cache tokens with the broadest scope so they can be reused later, that makes sense. But I’m not fully clear on how we’re handling storage of multiple tokens for the same audience are we keeping just one per audience, or allowing several with different scopes ?

src/utils/session-changes-helpers.ts
```diff
@@ -0,0 +1,104 @@
...
+  options: { scope?: string | null; audience?: string | null },
+  globalOptions: {
+    scope?: string | null | undefined;
+    audience?: string | null | undefined;
+  }
+): Partial<SessionData> | undefined {
+  const isAudienceTheGlobalAudience =
+    !options.audience || options.audience === globalOptions.audience;
+  const isScopeTheGlobalScope =
+    !options.scope || options.scope === globalOptions.scope;
```
nandan-bhat: Same as https://github.com/auth0/nextjs-auth0/pull/2333/files#diff-f286f78bd0e95e3de4a2adaeabd5ec5d2033af74e1f65969c961e6dbc18eea19R832

src/server/auth-client.ts
```diff
@@ -719,11 +733,20 @@ export class AuthClient {
...
+          this.authorizationParameters?.scope,
+          audience ?? this.authorizationParameters?.audience
+        ),
+        audience: this.authorizationParameters?.audience
+      }
+    );
+
+    if (sessionChanges) {
       if (idTokenClaims) {
         session.user = idTokenClaims as User;
```
nandan-bhat, line 747: Is it necessary to update the ID token on every `getAccessToken` call / `sessionChange` ?

src/utils/token-set-helpers.ts
frederikprijck: Fixed in https://github.com/auth0/nextjs-auth0/pull/2333/commits/27f29603e1b0355655cace095e54e485a123b6c7

src/utils/token-set-helpers.ts
frederikprijck: Fixed in https://github.com/auth0/nextjs-auth0/pull/2333/commits/7440c63f2c68d63b5065fcc680966f1a81781ee9

src/utils/session-changes-helpers.ts
frederikprijck: Fixed in https://github.com/auth0/nextjs-auth0/pull/2333/commits/227fbc2b5ae4302feb9e44e23912e1f49fc71a1b

src/server/auth-client.ts
frederikprijck: Fixed in https://github.com/auth0/nextjs-auth0/pull/2333/commits/227fbc2b5ae4302feb9e44e23912e1f49fc71a1b

src/utils/token-set-helpers.ts
frederikprijck: I think we want to use the `scope` as the base set of scopes, to enable the following:

- If users want to ensure scope `a`, `b` and `c` are always included with all the tokens, they can set it in the global scope definition.
- if users do not want this, they shouldn't, but instead pass them to `getAccessToken`.

I would typically encourage to use a broad set of globally defined scopes, and try and limit the usage of `getAccessToken` to passing scopes that are part of the global scopes, as that would be the most optimal in terms of token storage but also communicating with Auth0 to issue the tokens. 

However, we still allow for the use-case where the user would provide little global scopes, and leaves it to the `getAccessToken` to pass more granural scopes.

src/utils/token-set-helpers.ts
frederikprijck: 1. Yes, it are two different audiences, so they have to be different tokens.
2. It will impact the session, so perhaps we want to more explicitly call it out that using MRRT has an impact, and may require the move to stateful sessions.

I still have to add something to examples.md, but I could mention it in there as well as in the JSDocs for the audience when calling `getAccessToken`.

src/server/auth-client.ts
tusharpandey13: More context:
#1884 
#2189

src/server/auth-client.ts
frederikprijck: I have made quite some changes since this comment, including incorporating both `requested` scopes (`requestedScope`), as well as the granted scopes (`scope`) in the decision making which item to use from the cache.

nandan-bhat (approved):

src/utils/session-changes-helpers.ts
```diff
@@ -0,0 +1,300 @@
...
+ */
+function addNewAccessTokenSet(
+  session: SessionData,
+  tokenSet: TokenSet,
+  audience: string
+): Pick<SessionData, "accessTokens"> {
+  return {
+    accessTokens: [
+      ...(session.accessTokens || []),
+      accessTokenSetFromTokenSet(tokenSet, { audience })
```
nandan-bhat, line 130: Here, the `audience` comes from `tokenSet.audience` (one of the possibilities). We’re updating it based on what the user requests, but that doesn’t guarantee the audience actually exists in the tenant.

I tested this and noticed that Auth0 doesn’t throw an error when a random audience is passed, it just returns the global token set instead. This could be an issue for us since we rely on the `audience` to determine the correct token.

What do you think about this ?

src/utils/token-set-helpers.ts
```diff
@@ -0,0 +1,188 @@
...
+ * @param options.strict If true, requires an exact match of scopes (no extra scopes allowed)
+ * @returns True if all required scopes are present in the scopes, false otherwise
+ */
+export const compareScopes = (
+  scopes: string | null | undefined,
+  requiredScopes: string | undefined,
+  options: { strict?: boolean } = {}
+): boolean => {
+  // When the scopes and requiredScopes are exactly the same, return true
+  if (scopes === requiredScopes) {
```
nandan-bhat, line 68: Should we order/normalize it comparing ?
Example: `read write`  vs `write read` ?

src/utils/session-changes-helpers.ts
frederikprijck: The problem is, we can never know if the token being issued is for the correct audience. When using MRRT, the following happens:

- You send a `refresh_token` + `audience` (and optionally `scope`) to `/oauth/token`
  - Important to note, a `refresh_token` is bound to an `audience` and `scope`, the values used initially when logging in.
- Auth0 tries to find the `audience` in the `refresh_token.policies` entries
- If it finds one, it uses MRRT (this is not the case for your example)
- If it can't find one, it uses pre-MRRT behavior (this is your use-case).

If we forget about MRRT, this is how the `/oauth/token` endpoint worked:

- You send a `refresh_token` + `audience` (and optionally `scope`) to `/oauth/token`
  - `scope` can only be used to request less scopes than the `resfresh_token` is linked to
  - `audience` is completely ignored, instead the audience will be picked from the `refresh_token`
  - it will return a token from the initial `audience`.
  
So in your case, when you log in with `audience-1` (which exist), and then u ask for a token for audience-2 (which doesn't exist, both today and with MRRT, Auth0 will send back a token for `audience-1`.

This is confusing, but we can not detect this as we shouldn't be decoding the `access_token` to verify the `aud` claim. Instead, our SDK treats it as if it is a token for the unexisting audience, even tho this may be incorrect. 
Assumably, the user us using this audience because they need a token for that audience, and pretty soon they will try and use the token and notice it doesn't work and they need to configure MRRT correctly.

Long story short, I think we need to ensure we correctly call this out, which I have been trying to do here:

- https://github.com/auth0/nextjs-auth0/pull/2333/files#diff-fca89f70e75e01836aad1fae82cabf73cbb31164a29fa0f83c5f2dc0e05cf054R6-R12
- https://github.com/auth0/nextjs-auth0/pull/2333/files#diff-f30942cb123a919517afd870448544fac040e10ac91e49d781a0db0ed31fbdefR437-R440
- https://github.com/auth0/nextjs-auth0/pull/2333/files#diff-f30942cb123a919517afd870448544fac040e10ac91e49d781a0db0ed31fbdefR977-R982

I am also working on adding an entry for `EXAMPLES.md`, and I believe this also needs a big call-out in there.

nandan-bhat (approved):

src/utils/token-set-helpers.ts
frederikprijck: No, this is used for exact comparison, to skip any unnecessary splitting, sorting etc.

nandan-bhat (approved):

PR #2344, opened 2025-10-02
feat: add support for connected accounts

guabu:
### 📋 Changes

This PR adds support for the Connected Accounts flow:

- Introduces an opt-in endpoint `/auth/connect` to initiate connect account flow
- Introduces a `connectAccount()` method for custom route implementations
- Extended `onCallback` hook to handle connected account responses

Mounted route:

```typescript
  enableConnectAccountEndpoint: true // Mounts /auth/connect endpoint (customizable)
});
```

Programmatically via: 

```typescript
const response = await auth0.connectAccount({
  connection: "my-connection",
  authorizationParams: {
    scope: "openid profile email",
    audience: "https://myapi.com"
  },
  returnTo: "/connected"
});
```

The `onCallback` hook now receives additional context for connected account flows:
- `connectedAccount`: Account linking response data
- `responseType`: Set to `RESPONSE_TYPES.CONNECT_CODE` for account linking
- `returnTo`: Original return URL from the linking initiation

### DIFF
--- a/src/errors/index.ts
+++ b/src/errors/index.ts
@@ -203,9 +203,21 @@ export class MyAccountApiError extends SdkError {
   public detail: string;
   public status: number;
   public validationErrors?: Array<{
+    /**
+     * A human-readable description of the specific error. Required.
+     */
     detail: string;
+    /**
+     * The name of the invalid parameter. Optional.
+     */
     field?: string;
+    /**
+     * A JSON Pointer that points to the exact location of the error in a JSON document being validated. Optional.
+     */
     pointer?: string;
+    /**
+     *  Specifies the source of the error (e.g., body, query, or header in an HTML message). Optional.
+     */
     source?: string;
   }>;
 


src/server/client.ts
```diff
@@ -221,6 +224,11 @@ export interface Auth0ClientOptions {
   noContentProfileResponseWhenUnauthenticated?: boolean;
 
   enableParallelTransactions?: boolean;
+
+  /**
+   * If true, the `/auth/connect` endpoint will be mounted to enable users to connect additional accounts.
+   */
```
frederikprijck, line 230: Is there anything the user needs to do to be able to use My Account as an audience when these endpoints are being hit?

src/server/transaction-store.ts
```diff
@@ -1,16 +1,18 @@
...
 export interface TransactionState extends jose.JWTPayload {
-  nonce: string;
   codeVerifier: string;
-  responseType: string;
+  responseType: RESPONSE_TYPES;
   state: string; // the state parameter passed to the authorization server
   returnTo: string; // the URL to redirect to after login
+  nonce?: string; // A string value used to associate a client session with an ID Token, and to mitigate replay attacks.  codeVerifier: string;
   maxAge?: number; // the maximum age of the authentication session
+  authSession?: string; // the auth session ID for connect accounts
```
frederikprijck, line 15: Should it be more clear in the state that this is for something isolated? E.g. `{ connect: { authSession } }`?

src/types/connected-accounts.ts
```diff
@@ -0,0 +1,130 @@
...
+ */
+export interface ConnectAccountOptions {
+  /**
+   * The name of the connection to link the account with (e.g., 'google-oauth2', 'facebook').
+   */
+  connection: string;
+  /**
+   * Authorization parameters to be passed to the authorization server.
+   */
+  authorizationParams?: AuthorizationParameters;
```
frederikprijck, line 15: It looks like the SDK inconsistently uses both `authorizationParameters` and `authorizationParams`, with login we use `authorizationParameters`, would it make sense to align?

src/types/connected-accounts.ts
```diff
@@ -0,0 +1,130 @@
...
+   * Authorization Code flow.
+   */
+  CODE = "code",
+  /**
+   * Connect Account flow.
+   */
+  CONNECT_CODE = "connect_code"
+}
+
+export interface ConnectAccountRequest {
```
frederikprijck, line 33: Should we keep non-exported interfaces in a separate file, to make it less likely to accidentally expose them?

src/types/connected-accounts.ts
```diff
@@ -0,0 +1,130 @@
...
+   */
+  authSession: string;
+  /**
+   * The lifetime in seconds of the connect account session.
+   */
+  expiresIn: number;
+}
+
+export interface CompleteConnectAccountRequest {
+  accessToken: string;
```
frederikprijck, line 89:
```suggestion
  /**
   * The access token with the `create:me:connected_accounts` scope.
   */ 
   accessToken: string;
```

src/server/auth-client.ts
```diff
@@ -1541,6 +1733,230 @@ export class AuthClient {
...
+    const completeConnectAccountUrl = new URL(
+      "/me/v1/connected-accounts/complete",
+      this.issuer
+    );
+
+    try {
+      const httpOptions = this.httpOptions();
+      const headers = new Headers(httpOptions.headers);
+      headers.set("Content-Type", "application/json");
+      headers.set("Authorization", `Bearer ${options.accessToken}`);
```
frederikprijck, line 1893: How will this handle DPoP?

src/server/auth-client.ts
```diff
@@ -1541,6 +1733,230 @@ export class AuthClient {
...
+    try {
+      const connectAccountUrl = new URL(
+        "/me/v1/connected-accounts/connect",
+        this.issuer
+      );
+
+      const httpOptions = this.httpOptions();
+      const headers = new Headers(httpOptions.headers);
+      headers.set("Content-Type", "application/json");
+      headers.set("Authorization", `Bearer ${options.accessToken}`);
```
frederikprijck, line 1813: How will this handle DPoP?

src/server/client.ts
guabu: They'll have to enable the My Account API via the Dashboard — that's a good call out to add to the docs!

src/types/connected-accounts.ts
guabu: Yeah this is a bit unfortunate and I went back and forth about it — ultimately, My Account API calls the property in the request body `authorizationParams` so I thought it would make the most sense to align with the API. Since we already export `AuthorizationParameters`, I don't think we'll be able to change that one.

src/server/auth-client.ts
guabu: These 2 calls will have to be updated once the DPoP implementation is ready.

src/server/auth-client.ts
guabu: These 2 calls will have to be updated once the DPoP implementation is ready.

src/types/connected-accounts.ts
guabu: I slightly refactored the files so that anything exported under index.ts will be public but anything under the other files (e.g.: connected-accounts.ts) will be internal by default unless explicitly exported from the index.ts file. Happy to iterate on it if you have other ideas in mind.

src/server/transaction-store.ts
guabu: I don't have a super strong opinion on this one! Could definitely update it if it would help make things more clear.

src/server/client.ts
frederikprijck: I assume they will also need to configure DPoP at some point, as it will be enforcing DPoP once My Account is GA. Would it make sense to call that out already, or as part of the follow up PR that integrates Connected Accounts with the DPoP implementation?

src/server/transaction-store.ts
adamjmcgrath: I don't think you need to isolate `authSession`, if you did you'd want to isolate the login only stuff as well - but I don't think there's much value in it.

fwiw I went with defining types for each transaction type then using a generic on the transaction store https://github.com/auth0/auth0-spa-js/pull/1422/files#diff-1f8cb53e1966eb1ab668d418f7fb31218912295c6bffb874f97f7f7dca8faeb6

Although tbh what you have here looks fine, it's not used externally

Side note - it's weird that `TransactionState` extends `jose.JWTPayload`

src/server/transaction-store.ts
```diff
@@ -1,16 +1,18 @@
...
 const TRANSACTION_COOKIE_PREFIX = "__txn_";
 
 export interface TransactionState extends jose.JWTPayload {
-  nonce: string;
   codeVerifier: string;
-  responseType: string;
+  responseType: RESPONSE_TYPES;
   state: string; // the state parameter passed to the authorization server
   returnTo: string; // the URL to redirect to after login
+  nonce?: string; // A string value used to associate a client session with an ID Token, and to mitigate replay attacks.  codeVerifier: string;
```
adamjmcgrath, line 13: You can remove this - we're not using nonce since there are now tokens produced from the connect flow to validate

src/server/auth-client.ts
adamjmcgrath: Could you use this https://github.com/panva/oauth4webapi/blob/7f2ae90e9c861f4cce54d9f5f3786b575a81566c/docs/functions/protectedResourceRequest.md - then you might get DPoP for free when it's shipped

src/server/auth-client.ts
```diff
@@ -1541,6 +1733,230 @@ export class AuthClient {
...
+    const transactionState: TransactionState = {
+      codeVerifier,
+      responseType: RESPONSE_TYPES.CONNECT_CODE,
+      state,
+      returnTo,
+      authSession: connectAccountResponse.authSession
+    };
+
+    const res = NextResponse.redirect(
+      `${connectAccountResponse.connectUri}?ticket=${connectAccountResponse.connectParams.ticket}`
```
pmalouin: should we uri-encode the ticket? I'd guess that we use a safe character set, but might be good to do it for additional safety and future-proofing?

src/server/transaction-store.ts
adamjmcgrath: Actually, nonce is used by the login flow when we do form post implicit so it can stay

frederikprijck:
@claude

src/server/client.ts
guabu: > or as part of the follow up PR that integrates Connected Accounts with the DPoP implementation?

I think we can do that as part of the DPoP PR #2357 👌

src/server/transaction-store.ts
guabu: Sounds good, I'll stick with the current approach then 👍

src/server/transaction-store.ts
guabu: Yup, I've removed the `nonce` from the connected accounts flows — this is for the code flow as you mentioned 👌

src/server/auth-client.ts
guabu: Good call! I've pushed a commit to encode it 👍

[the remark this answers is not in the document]
guabu: Done ✅

src/server/client.ts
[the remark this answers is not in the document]
guabu: DPoP docs will be added when the referenced PR is merged.

src/server/auth-client.ts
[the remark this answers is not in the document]
guabu: Once DPoP is added.

src/server/auth-client.ts
guabu: I believe the method `protectedResourceRequest` will be wrapped in a fetcher implementation as part of #2357

src/server/auth-client.ts
```diff
@@ -547,9 +584,78 @@ export class AuthClient {
...
+          onCallbackCtx,
+          req,
+          state
+        );
+      }
+
+      const [completeConnectAccountError, connectedAccount] =
+        await this.completeConnectAccount({
+          accessToken: tokenSetResponse.tokenSet.accessToken,
+          authSession: transactionState.authSession!,
```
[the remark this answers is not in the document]
guabu: This is expected as the API will return an error and we will propagate it if the auth session is missing or any reason.

src/server/auth-client.ts
```diff
@@ -547,9 +584,78 @@ export class AuthClient {
...
+          connectCode: req.nextUrl.searchParams.get("connect_code")!,
```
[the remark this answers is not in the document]
guabu: The API will return an error if the connect code is missing and it will be propagated to the caller.

PR #2386, opened 2025-10-29
fix: append intl headers in with-next-intl instead of overwrite

### LINKED ISSUE #2340: Sample application does not copy cookie headers from Auth0 response to Intl response correctly
### Checklist

### Description

In the example for how to setup nextjs-auth0 with intl, there is a middleware file that show how to copy header values form Auth0 response to intl response. [The code line](https://github.com/auth0/nextjs-auth0/blob/6daabeb68dbdd0bce72743eb2ca9383ed137d3bd/examples/with-next-intl/middleware.ts#L38C1-L38C37) that writes header values to intl response looks like this: `intlRes.headers.set(key, value);`. If the header value exists, the previous value will be overwritten. The Auth0 response contains multiple `set-cookie` headers. Since `headers.set` is used in stead of `headers.append`, only the last cookie is written to the response. Because of this error, the session is not updated correctly and it will expire before its absolute expiry time even though the user is actively using the application. Can we update the example so that it uses `append`? 

Also, I think it is a common use case to have other middleware that generates responses. Could the SDK be extended with a helper method for copying the header values? 

### Reproduction

.

### Additional context

Example logging of headers that are copied from Auth0 response to intl response: 
```
Copying cookie from auth response. "cache-control"="private, no-cache, no-sto..."
Copying cookie from auth response. "expires"="0..."
Copying cookie from auth response. "pragma"="no-cache..."
Copying cookie from auth response. "set-cookie"="__session__0=eyJlbmMiOiJB..."
Copying cookie from auth response. "set-cookie"="__session__1=rgPjj5rbkWdW..."
Copying cookie from auth response. "set-cookie"="__session=; Path=/; Max-A..."
Copying cookie from auth response. "set-cookie"="appSession=; Path=/; Max-..."
Copying cookie from auth response. "x-middleware-next"="1..."
Copying cookie from auth response. "x-middleware-set-cookie"="__session__0=eyJlbmMiOiJB..."
```

### nextjs-auth0 version

4.10.0

### Next.js version

15.5.3

### Node.js version

22.18.0

tusharpandey13:
Fixes #2340

Use `Headers.append()` instead of `Headers.set()`

See [Headers: append() method](https://developer.mozilla.org/en-US/docs/Web/API/Headers/append)

### DIFF
--- a/examples/with-next-intl/middleware.ts
+++ b/examples/with-next-intl/middleware.ts
@@ -34,14 +34,11 @@ export async function middleware(request: NextRequest) {
       }
     }
     // For all other headers, or if 'x-middleware-next' can be safely copied,
-    // copy them to intlRes. Use append() if the header already exists to support
-    // multi-value headers like set-cookie. HTTP allows multiple headers with the same name,
-    // and cookies must be preserved separately. Using set() would overwrite previous cookies.
-    if (intlRes.headers.has(key)) {
-      intlRes.headers.append(key, value);
-    } else {
-      intlRes.headers.set(key, value);
-    }
+    // append them to intlRes. The append() method adds the header if it doesn't exist,
+    // or appends the value if it does. This correctly preserves multi-value headers
+    // like set-cookie. HTTP allows multiple headers with the same name, and using
+    // append() ensures all cookie values are preserved instead of overwriting.
+    intlRes.headers.append(key, value);
   }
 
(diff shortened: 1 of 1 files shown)


tusharpandey13:
@claude please review this PR

nandan-bhat (approved):

PR #2400, opened 2025-11-03
feat: Proxy handler support enabling My Account and My Org

tusharpandey13:
This PR enables support for secure browser-initiated My-Account and My-Org API calls through a `handleProxy` handler, a transparent and secure authenticating proxy to handle secure API calls originating from browser.

## Changes
- `auth-client.ts`: added methods `handleMyAccount` and `handleMyOrg`, added matchers for these in `handler`.
- `utils/proxy.ts`: header filters, url transform, proxy matcher

## Tests
`proxy-handler.test.ts`: Added tests for proxy handler covering:
- Basic Proxy Routing & Session Management
- HTTP Method Routing
- URL Path Matching & Transformation
- HTTP Headers Forwarding
- Request Body Handling
- Bearer Token Handling
- DPoP Token Handling
- Session Update After Token Refresh
- Error Scenarios
- Concurrent Request Handling
- CORS Handling

`proxy.test.ts`: Tests for header filters, url transform, proxy matcher
`auth-client.proxy.test.ts`: proxy tests specific to my-account and my-org
`auth-client.test.ts`: updated test cases

## References
- RFC 2616
- RFC 7231
- RFC 6454 
- RFC 9449
- WHATWG Fetch Spec

## Implementation details
The proxy matcher uses the following rules:
```ts
if (sanitizedPathname.startsWith("/me/")) {
      return this.handleMyAccount(req);
} else if (sanitizedPathname.startsWith("/my-org/")) {
      return this.handleMyOrg(req);
}
```

`scope` header is used to supply scope (not `auth0-scope`)

### DIFF
--- a/src/server/auth-client.ts
+++ b/src/server/auth-client.ts
@@ -1209,122 +1209,21 @@ export class AuthClient {
         !tokenSet.expiresAt ||
         tokenSet.expiresAt <= Date.now() / 1000
       ) {
-        const [discoveryError, authorizationServerMetadata] =
-          await this.discoverAuthorizationServerMetadata();
+        const [error, response] = await this.#refreshTokenSet(tokenSet, {
+          audience: options.audience,
+          scope: options.scope ? scope : undefined,
+          requestedScope: scope
+        });
 
-        if (discoveryError) {
-          return [discoveryError, null];
-        }
-
-        const additionalParameters = new URLSearchParams();
-
-        if (options.scope) {
-          additionalParameters.append("scope", scope);
-        }
-
-        if (options.audience) {
-          additionalParameters.append("audience", options.audience);
-        }
-
-        // Create DPoP handle ONCE outside the closure so it persists across retries.
-        // This is required by RFC 9449: the handle must learn and reuse the nonce from
-        // the DPoP-Nonce header across multiple attempts.
-        const dpopHandle =
-          this.useDPoP && this.dpopKeyPair
-            ? oauth.DPoP(this.clientMetadata, this.dpopKeyPair)
-            : undefined;
-
-        const refreshTokenGrantRequestCall = async () =>
-          oauth.refreshTokenGrantRequest(
-            authorizationServerMetadata,
-            this.clientMetadata,
-            await this.getClientAuth(),
-            tokenSet.refreshToken!,
-            {
-              ...this.httpOptions(),
-              [oauth.customFetch]: this.fetch,
-              [oauth.allowInsecureRequests]: this.allowInsecureRequests,
-              additionalParameters,
-              ...(dpopHandle && {
-                DPoP: dpopHandle
-              })
-            }
-          );
-
-        const processRefreshTokenResponseCall = (response: Response) =>
-          oauth.processRefreshTokenResponse(
-            authorizationServerMetadata,
-            this.clientMetadata,
-            response
-          );
-
-        let oauthRes: oauth.TokenEndpointResponse;
-        try {
-          oauthRes = await withDPoPNonceRetry(
-            async () => {
-              const refreshTokenRes = await refreshTokenGrantRequestCall();
-              return await processRefreshTokenResponseCall(refreshTokenRes);
-            },
-            {
-              isDPoPEnabled: !!(this.useDPoP && this.dpopKeyPair),
-              ...this.dpopOptions?.retry
-            }
-          );
-        } catch (e: any) {
-          return [
-            new AccessTokenError(
-              AccessTokenErrorCode.FAILED_TO_REFRESH_TOKEN,
-              "The access token has expired and there was an error while trying to refresh it.",
-              new OAuth2Error({
-                code: e.error,
-                message: e.error_description
-              })
-            ),
-            null
-          ];
-        }
-
-        const idTokenClaims = oauth.getValidatedIdTokenClaims(oauthRes)!;
-        const accessTokenExpiresAt =
-          Math.floor(Date.now() / 1000) + Number(oauthRes.expires_in);
-
-        const updatedTokenSet = {
-          ...tokenSet, // contains the existing `iat` claim to maintain the session lifetime
-          accessToken: oauthRes.access_token,
-          idToken: oauthRes.id_token,
-          // We store the both requested and granted scopes on the tokenSet, so we know what scopes were requested.
-          // The server may return less scopes than requested.
-          // This ensures we can return the same token again when a token for the same or less scopes is requested by using `requestedScope` during look-up.
-          //
-          // E.g. When requesting a token with scope `a b`, and we return one for scope `a` only,
-          // - If we only store the returned scopes, we cannot return this token when the user requests a token for scope `a b` again.
-          // - If we only store the requested scopes, we lose track of the actual scopes granted.
-          //
-          // Scopes actually granted by the server
-          scope: oauthRes.scope,
-          // Scopes requested by the client
-          requestedScope: scope,
-          expiresAt: accessTokenExpiresAt,
-          // Keep the audience if it exists, otherwise use the one from the options.
-          // If not provided, use `undefined`.
-          audience: tokenSet.audience || options.audience || undefined,
-          // Store the token type from the OAuth response (e.g., "Bearer", "DPoP")
-          ...(oauthRes.token_type && { token_type: oauthRes.token_type })
-        };
-
-        if (oauthRes.refresh_token) {
-          // refresh token rotation is enabled, persist the new refresh token from the response
-          updatedTokenSet.refreshToken = oauthRes.refresh_token;
-        } else {
-          // we did not get a refresh token back, keep the current long-lived refresh token around
-          updatedTokenSet.refreshToken = tokenSet.refreshToken;
+        if (error) {
+          return [error, null];
         }
 
         return [
           null,
           {
-            tokenSet: updatedTokenSet,
-            idTokenClaims: idTokenClaims
+            tokenSet: response.updatedTokenSet,
+            idTokenClaims: response.idTokenClaims
           }
         ];
       }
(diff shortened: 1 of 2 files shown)


src/server/auth-client.ts
```diff
@@ -2199,6 +2236,198 @@ export class AuthClient {
...
+        method: "OPTIONS",
+        headers
+      });
+
+      // CORS preflight responses should be 204 No Content per spec
+      // Forward CORS headers from upstream but normalize status to 204
+      return new NextResponse(null, {
+        status: 204,
+        headers: buildForwardedResponseHeaders(preflightResponse)
+      });
```
frederikprijck, line 2180: Is it expected to return a 204, even in cases where the API would return something different (e.g. 3xx or 4xx)? Why not just forward the status from the `preflightResponse` and leave correct CORS implementation to the API itself?

tusharpandey13: makes sense, will do

src/utils/proxy.test.ts
```diff
@@ -0,0 +1,507 @@
...
+          proxyPath: "/me",
+          targetBaseUrl: "https://issuer/me/v1",
+          audience: "https://issuer/me/",
+          scope: null
+        };
+
+        const result = transformTargetUrl(req, options);
+
+        // Expected: https://issuer/me/v1/v1/some-endpoint
+        // NOT: https://issuer/me/v1/v1/some-endpoint (old buggy behavior)
```
frederikprijck: I assume we can remove this comment, and perhaps update the describe block as well to not mention bug.

src/server/auth-client.test.ts
```diff
@@ -6423,6 +6424,205 @@ ca/T0LLtgmbMmxSv/MmzIg==
     });
   });
 
+  describe("handleMyAccount", async () => {
```
frederikprijck: I believe we can delete this block, as these tests are included in auth-client.proxy.test.ts

src/utils/proxy.test.ts
tusharpandey13: done

src/server/auth-client.test.ts
tusharpandey13: done

nandan-bhat (approved):

src/server/auth-client.ts
```diff
@@ -1055,6 +1067,24 @@ export class AuthClient {
     return connectAccountResponse;
   }
 
+  async handleMyAccount(req: NextRequest): Promise<NextResponse> {
+    return this.#handleProxy(req, {
+      proxyPath: "/me",
+      targetBaseUrl: `${this.issuer}/me/v1`,
+      audience: `${this.issuer}/me/`,
+      scope: req.headers.get("scope")
```
nandan-bhat, line 1085: Should we use `req.headers.get("auth0-scope")` ? 

Because we are documenting that way here: https://github.com/auth0/nextjs-auth0/pull/2400/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R331

src/server/auth-client.ts
```diff
@@ -1055,6 +1067,24 @@ export class AuthClient {
...
+    });
+  }
+
+  async handleMyOrg(req: NextRequest): Promise<NextResponse> {
+    return this.#handleProxy(req, {
+      proxyPath: "/my-org",
+      targetBaseUrl: `${this.issuer}/my-org`,
+      audience: `${this.issuer}/my-org/`,
+      scope: req.headers.get("scope")
```
nandan-bhat, line 1094: Should we use `req.headers.get("auth0-scope")` ? 

Because we are documenting that way here: https://github.com/auth0/nextjs-auth0/pull/2400/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R336

src/server/auth-client.ts
```diff
@@ -399,6 +407,10 @@ export class AuthClient {
       this.enableConnectAccountEndpoint
     ) {
       return this.handleConnectAccount(req);
+    } else if (sanitizedPathname.startsWith("/me")) {
```
nandan-bhat: What's the behavior when we have this in the `next.config.js` ? 

```ts
module.exports = {
  basePath: "/app"
}
```

This it's a generic comment and is not specific to this implementation. I'd like to know if it can affect other route conditions too.

src/server/auth-client.ts
```diff
@@ -399,6 +407,10 @@ export class AuthClient {
...
+      return this.handleMyAccount(req);
```
nandan-bhat, line 421: Also, I think `sanitizedPathname.startsWith("/me")` would match all the endpoints which starts with `/me`.

Example: 
```
- /meaw/v1
- /meta/v1
```

src/server/auth-client.ts
```diff
@@ -2199,6 +2229,198 @@ export class AuthClient {
...
+    // This retry will not happen with bearer auth so no need to clone if DPoP is false.
+    const clonedReq = this.useDPoP ? req.clone() : req;
+    const targetUrl = transformTargetUrl(req, options);
+
+    // Set Host header to upstream host
+    headers.set("host", targetUrl.host);
+
+    // Buffer the body to allow retry on DPoP nonce errors
+    // ReadableStreams can only be consumed once, so we need to buffer for retry
+    const bodyBuffer = clonedReq.body
```
nandan-bhat, line 2236: What will happen if a malicious client sends a very large body to `/me` endpoint ? SDK is loading full body into memory.

src/server/auth-client.ts
```diff
@@ -2199,6 +2229,198 @@ export class AuthClient {
...
+      const response = await fetcher.fetchWithAuth(
+        targetUrl.toString(),
+        {
+          method: clonedReq.method,
+          headers,
+          body: bodyBuffer,
+          // @ts-expect-error duplex is not known, while we do need it for sending streams as the body.
+          // As we are receiving a request, body is always exposed as a ReadableStream when defined,
+          // so setting duplex to 'half' is required at that point.
+          duplex: bodyBuffer ? "half" : undefined
```
nandan-bhat: I just looked it up the behavior of `duplex`.

Looks like, if you make a fetch with `duplex: "half"` while the body is not a stream (like an ArrayBuffer), certain runtimes throw an error.

I don't know how to reproduce this though.

src/server/auth-client.ts
```diff
@@ -2199,6 +2229,198 @@ export class AuthClient {
...
+      // Forward preflight directly WITHOUT DPoP/auth headers
+      const preflightResponse = await this.fetch(targetUrl.toString(), {
+        method: "OPTIONS",
+        headers
+      });
+
+      // CORS preflight responses should be 204 No Content per spec
+      // Forward CORS headers from upstream but normalize status to 204
+      return new NextResponse(null, {
+        status: preflightResponse.status,
```
nandan-bhat, line 2178: This (`preflightResponse.status`) is not guaranteed to have `204 No Content` right ?

src/utils/proxy.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    // Forward if:
+    // 1. It's in the allow-list, AND
+    // 2. It's not a hop-by-hop header
+    const shouldForward =
+      DEFAULT_HEADER_ALLOW_LIST.has(lowerKey) &&
+      !HOP_BY_HOP_HEADERS.has(lowerKey);
+
+    if (shouldForward) {
+      forwardedHeaders.set(key, value);
```
nandan-bhat, line 84: Can upstream response contain `Set-Cookie` header ? If yes, we are blindly setting the cookie in the browser.

Just wondering if that could happen in this case.

src/server/auth-client.ts
tusharpandey13: will update docs

src/server/auth-client.ts
frederikprijck: You can reproduce it by executing this:

```js
const body = { foo: 'bar' }; // object
const request1 = new Request('http://www.google.com', { body, method: 'post' }); // body is object

console.log(request1.body); // request1.body is ReadableStream

const request2 = new Request('http://www.google.com', { body: request1.body, method: 'post' }); // body is ReadableStream, so fails if no duplex is set.
```

src/server/auth-client.ts
frederikprijck: Should we be more specific and check for `/me/v1/`?

src/server/auth-client.ts
frederikprijck: I think the comment is not updated after reverting some changes.

src/server/auth-client.ts
nandan-bhat: IMO, yes.
Because currently it's triggering  `myAccount` flow for `/me*`.  Same with `/my-org`.

src/server/auth-client.ts
tusharpandey13: will update docs

src/server/auth-client.ts
tusharpandey13: fixed, handler now strips basepath from final path to match.

src/server/auth-client.ts
tusharpandey13: This is a theoretical vulnerability with negligible practical risk for My Account and My Organization APIs.
We can solve this in a separate PR (if proxy handler needs to be made generic)

src/utils/proxy.ts
tusharpandey13: This is a theoretical vulnerability with negligible practical risk for My Account and My Organization APIs.
My Account/My Org APIs don't send `Set-Cookie`.
We can solve this in a follow-up PR, if needed for generic use-cases.

src/server/auth-client.ts
tusharpandey13: We can, however, look into adding a hard limit on the request buffer (configurable with options and env variable). Default for this can be set to something like 10MB.

src/server/auth-client.ts
frederikprijck: I removed the duplex option, as it didn't seem to be necessary anymore.

src/server/auth-client.ts
tusharpandey13: Added check for:
```
else if (
      sanitizedPathname.startsWith("/me/")
    )
```

src/server/auth-client.ts
tusharpandey13: updated docs

src/server/auth-client.ts
```diff
@@ -1055,6 +1077,24 @@ export class AuthClient {
     return connectAccountResponse;
   }
 
+  async handleMyAccount(req: NextRequest): Promise<NextResponse> {
+    return this.#handleProxy(req, {
```
nandan-bhat, line 1081: `/me` and `/my-org` use `DPoP`. When `DPoP` is disabled, `#handleProxy` falls back to `Bearer`. Does it make sense to return an error immediately when `/me` or `/my-org` are accessed when `DPoP` is disabled ?

EXAMPLES.md
```diff
@@ -1550,6 +1567,341 @@ const fetcher = await auth0.createFetcher(req, {
...
+      const data = await response.json();
+      setProfile(data);
+    } catch (error) {
+      console.error("Failed to fetch profile:", error);
+    } finally {
+      setLoading(false);
+    }
+  };
+
+  const updateProfile = async (updates) => {
```
nandan-bhat, line 1656: If the page makes a `My Account` request (`updateProfile` in this case) during page render, and that page is linked using the Next.js `<Link>` component, it could trigger an actual API request during prefetch in production.

I think we should either fix this or document this.

src/utils/proxy.ts
nandan-bhat, line 84: This is a minor one.

- If headers are case in-sensitive, we can set `forwardedHeaders.set(lowerKey, value);` anyway.
- If headers are case-sensitive, we should not do `key.toLowerCase();`


But, headers are case in-sensitive, so this should be fine.

src/server/auth-client.ts
tusharpandey13: We can plan to take this change live to error on DPoP disabled incrementally, upon confirming that upstream APIs do not, in-fact, work with DPoP disabled.

PR #2405, opened 2025-11-07
Adding support for Next 16

nandan-bhat:
### ✅ Next.js 16 Support (Node Proxy Compatibility)

This PR adds full compatibility for both **Next 15 (Edge middleware)** and **Next 16 (Node proxy)** runtimes, without requiring any app-level code changes.

#### 🔧 Changes

* **Request normalization:**
  Added internal `toNextRequest()` helper to safely convert a plain `Request` (as used in Next 16 proxy) to a `NextRequest`.
  Used in:

  * `middleware()`
  * `getSession()`
  * `updateSession()`
  * `getAccessTokenForConnection()`
  * `createFetcher()`

* **Response normalization:**
  Introduced internal `toNextResponse()` helper for consistent cookie and header handling when a plain `Response` is used.

* **Backward-compatibility:**
  All existing examples and apps continue to work.
  Developers only need to:

  * Rename **`middleware.ts` → `proxy.ts`**
  * Rename the exported function **`middleware` → `proxy`**

* **Edge vs Node behavior:**
  Next 15 still runs middleware on Edge.
  Next 16 runs proxy in Node runtime (middleware remains deprecated but supported under `experimental.nodeMiddleware`).

### DIFF
--- a/src/server/next-compat.ts
+++ b/src/server/next-compat.ts
@@ -13,7 +13,6 @@ export function toNextRequest(input: Request | NextRequest): NextRequest {
   return new NextRequest(input.url, {
     method: input.method,
     headers: input.headers,
-    // TODO: We might need to handle the case where body is a ReadableStream
     body: input.body as any,
     duplex: (input as any).duplex ?? "half"
   });


src/server/next-compat.test.ts
```diff
@@ -0,0 +1,114 @@
+import { NextRequest, NextResponse } from "next/server.js";
+import { describe, expect, it } from "vitest";
+
+import { toNextRequest, toNextResponse } from "./next-compat.js";
+
+describe("next-compact", () => {
```
tusharpandey13: rename "next-compact" to "next-compat"
```suggestion
describe("next-compat", () => {
```

src/server/next-compat.ts
```diff
@@ -0,0 +1,51 @@
...
+    return input;
+  }
+
+  return new NextRequest(input.url, {
+    method: input.method,
+    headers: input.headers,
+    body: input.body as any,
+    duplex: (input as any).duplex ?? "half"
+  });
+}
```
tusharpandey13, line 19: Do we want to consider cloning input request body before transforming? Just a suggestion, open to comments on this.

src/server/next-compat.test.ts
nandan-bhat: Thanks. I have updated it in this commit https://github.com/auth0/nextjs-auth0/pull/2405/commits/a716c40f1b1034e3210f2a45ea381420e77ce280

src/server/next-compat.ts
nandan-bhat: Any specific reason ? We are not writing into the original `input` object, we are just reading from it.

nandan-bhat: Added a `TODO` here https://github.com/auth0/nextjs-auth0/pull/2405/commits/6575ee2219a2a3fdd381819e07aca0f90f4f7fb8

nandan-bhat: We aren’t reading the stream, so `body: input.body` does not affect or consume it. No changes are required here.

PR #2407, opened 2025-11-10
feat: add support `scopes` parameter for connected accounts

guabu:
### 📋 Changes

Adds support for a top-level `scopes` parameter when initiating a connect account flow.

### 🎯 Testing

See included tests.


jfromaniello (approved):

src/server/auth-client.ts
```diff
@@ -1056,6 +1058,7 @@ export class AuthClient {
       await this.connectAccount({
         tokenSet: tokenSet,
         connection,
+        scopes,
```
pmalouin: Let's make sure `scopes` is never an empty array, rather, omit the param if empty. (`getAll()` above will return an empty array when there is no `scopes` in the search params)

guabu: Done as part of: https://github.com/auth0/nextjs-auth0/pull/2407/commits/8792639ace1f64cda9e410e70691fbcdc6936834

arpit-jn:
@claude

Issue #2394, opened 2025-10-31
Cookie not updating after getting a new access token


vortexture:
### Checklist

- [ ] ~~The issue can be reproduced in the [nextjs-auth0 sample app](https://github.com/auth0-samples/auth0-nextjs-samples/tree/main/Sample-01) (or N/A).~~
- [x] I have looked into the [Readme](https://github.com/auth0/nextjs-auth0#readme), [Examples](https://github.com/auth0/nextjs-auth0/blob/main/EXAMPLES.md), and [FAQ](https://github.com/auth0/nextjs-auth0/blob/main/FAQ.md) and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://auth0.github.io/nextjs-auth0/) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/nextjs-auth0/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

I am attempting to upgrade the version of this library I am using from `4.0.2` to `4.11.0`
I am using the pages router and I use rotating refresh tokens

I have been using `getSession(request)` (to confirm there is a session) and `getAccessToken(request, response)` in my api route to get the access token so I can retrieve my users roles from the access token (this is all after checking for a valid session in the middleware).

Since updating I have been having an issue where my session does not get updated with the new access token & refresh token, so once my token expires the call to `getAccessToken(request, response)` will get a new token but then after the grace period for rotating refresh token reuse passes I get an error for refresh token reuse any time I call `getAccessToken`

I have tried moving the retrieval of the access token to the the middleware and passing the access token directly to the api routes via the headers, this seems to have the same issue (this is following the recomendation in the second example for getting an access token in the middleware and setting the authRes headers - https://github.com/auth0/nextjs-auth0/blob/main/EXAMPLES.md#middleware-2)

I have tracked the problem down to to the version change `4.2.1` -> `4.4.4` https://github.com/auth0/nextjs-auth0/compare/v4.2.1...v4.4.0
After upgrading beyond this I encounter the above issue

Any guidance on what I might be doing wrong, or what changed that might cause this would be appreciated

[...]

PR #2447, opened 2025-12-04
bugfix: session write not happening for pages router with chunked cookies

tusharpandey13:
When using `getAccessToken(req, res)` or `updateSession(req, res, session)` in Pages Router with chunked sessions, only the last cookie chunk was being persisted. This caused `invalid_grant` errors on subsequent requests because the session was corrupted.

Fixes: #2394

## Root Cause
`Headers.entries()` yields each Set-Cookie header separately, but `res.setHeader()` overwrites previous values instead of appending in `saveToSession`.

## Changes
- client.ts: Updated logic to use `pagesRouterRes.appendHeader` for `set-cookie` header.
- added a UT for the same.

## Reproduction steps
- Set Access token lifetime to 10s in auth0 dashboard
- Added a pages router page that fetches Access token
- Page refreshed after 10s multiple times
- On second refresh, an `invalid_grant` error was thrown.

Reproduction failed after fixes.


vortexture:
I can confirm this appears to resolve the issue I raised when I apply the same change as a patch to the package https://github.com/auth0/nextjs-auth0/issues/2394

PR #2484, opened 2026-01-12
chore: introduce generic request to be able to handle different kind of requests

frederikprijck:
### 📋 Changes

Introduces an abstraction layer to decouple authentication logic from Next.js-specific `NextRequest`/`NextResponse` types. This refactor enables future support for API routes with Pages Router without duplicating authentication code.

**Impact:** No breaking changes to external APIs. This is purely an internal refactor - middleware continues to work exactly as before.

#### Motivation

The current v4 implementation is tightly coupled to Next.js middleware types (`NextRequest` / `NextResponse`). Supporting API routes or Pages Router would require duplicating all authentication logic for `NextApiRequest` / `NextApiResponse`, or adding some `if/else` brancing.

This abstraction establishes a foundation for multi-paradigm support through polymorphism rather than code duplication.

#### Overview

**New HTTP layer** (`src/server/http/`):
- `Auth0Request<TRequest>` - Abstract base class for request operations
- `Auth0Response<TResponse>` - Abstract base class for response operations
- `Auth0NextRequest` / `Auth0NextResponse` - Concrete implementations for middleware
- `Auth0RequestCookies` / `Auth0ResponseCookies` - Unified cookie interfaces

**Key design decisions:**

1. **Generic type preservation**: `Auth0Request<NextRequest>` maintains full type information through the stack
2. **Mutable response pattern**: Response methods mutate internal state.
3. **Cookie abstraction unification**: Single interface for both `RequestCookies` and `ReadonlyRequestCookies`

##### 1. Handler Abstraction Pattern

All authentication handlers now accept abstracted types instead of concrete Next.js types:

```typescript
// Before
async handleLogin(req: NextRequest): Promise<NextResponse>

// After
async handleLogin(req: Auth0Request, res: Auth0Response): Promise<Auth0Response>
```

This enables the same handler logic to work with any request/response implementation (middleware, API routes, route handlers).

##### 2. Response Wrapper & Header Merging

Introduced `#unwrapHandler()` helper in `auth-client.ts` that:
- Calls handlers that now return Auth0Response instead of NextResponse
- Unwraps the concrete `NextResponse` from `Auth0Response.res
- Returns the `NextResponse` instance.

The response wrapper pattern allows handlers to remain agnostic of the concrete response type while maintaining compatibility with Next.js middleware expectations in a non-breaking way.

##### 3. Session & Transaction Store Refactoring

All storage abstractions now operate on `Auth0RequestCookies` / `Auth0ResponseCookies`:

- `AbstractSessionStore` - Updated interface for get/set/delete operations
- `StatelessSessionStore` - Cookie encoding/chunking logic works with abstractions
- `StatefulSessionStore` - Database operations use abstracted cookie interfaces
- `TransactionStore` - State parameter handling via abstractions

This enables future storage implementations to work seamlessly across different Next.js contexts.

##### 4. Response 

When using NextResponse, we typically need to create a new instance and return it. However, to support Pages Router, which uses NextApiResponse, we know that it will be different in the sense that we will accept a NextApiResponse and call methods on that instance rather than return new instances. Therefore, this PR already ensures we create an Auth0Response in the beginning of all handlers, and returns the underlying `Auth0Response.res`.

This may be a bit weird looking at it purely from the  App Router perspective, but is done to support Pages Router in a follow up PR.

### 📎 References

N/A

### 🎯 Testing

- All existing middleware tests pass without modification
- New test cases covering abstraction layer behavior

### DIFF
--- a/src/server/auth-client.test.ts
+++ b/src/server/auth-client.test.ts
@@ -7956,7 +7956,9 @@ ca/T0LLtgmbMmxSv/MmzIg==
         }
       );
 
-      await authClient.startInteractiveLogin(new Auth0NextResponse(new NextResponse()));
+      await authClient.startInteractiveLogin(
+        new Auth0NextResponse(new NextResponse())
+      );
 
       expect(authClient["transactionStore"].save).toHaveBeenCalled();
     });
@@ -7979,7 +7981,10 @@ ca/T0LLtgmbMmxSv/MmzIg==
         }
       );
 
-      await authClient.startInteractiveLogin(new Auth0NextResponse(new NextResponse()), { returnTo });
+      await authClient.startInteractiveLogin(
+        new Auth0NextResponse(new NextResponse()),
+        { returnTo }
+      );
 
       expect(authClient["transactionStore"].save).toHaveBeenCalled();
     });
@@ -8002,7 +8007,10 @@ ca/T0LLtgmbMmxSv/MmzIg==
         }
       );
 
-      await authClient.startInteractiveLogin(new Auth0NextResponse(new NextResponse()), { returnTo });
+      await authClient.startInteractiveLogin(
+        new Auth0NextResponse(new NextResponse()),
+        { returnTo }
+      );
 
       expect(authClient["transactionStore"].save).toHaveBeenCalled();
     });
@@ -8028,7 +8036,10 @@ ca/T0LLtgmbMmxSv/MmzIg==
         }
       );
 
-      await authClient.startInteractiveLogin(new Auth0NextResponse(new NextResponse()), { returnTo: unsafeReturnTo });
+      await authClient.startInteractiveLogin(
+        new Auth0NextResponse(new NextResponse()),
+        { returnTo: unsafeReturnTo }
+      );
 
       expect(authClient["transactionStore"].save).toHaveBeenCalled();
     });
@@ -8050,7 +8061,10 @@ ca/T0LLtgmbMmxSv/MmzIg==
         return originalAuthorizationUrl.call(authClient, params);
       });
 
-      await authClient.startInteractiveLogin(new Auth0NextResponse(new NextResponse()), { authorizationParameters });
+      await authClient.startInteractiveLogin(
+        new Auth0NextResponse(new NextResponse()),
+        { authorizationParameters }
+      );
 
       expect(authClient["authorizationUrl"]).toHaveBeenCalled();
     });
@@ -8084,7 +8098,9 @@ ca/T0LLtgmbMmxSv/MmzIg==
         fetch: mockFetch
       });
 
-      await authClient.startInteractiveLogin(new Auth0NextResponse(new NextResponse()));
+      await authClient.startInteractiveLogin(
+        new Auth0NextResponse(new NextResponse())
+      );
 
       // Verify that PAR was used
       expect(parRequestCalled).toBe(true);
@@ -8115,7 +8131,10 @@ ca/T0LLtgmbMmxSv/MmzIg==
         }
       );
 
-      await authClient.startInteractiveLogin(new Auth0NextResponse(new NextResponse()), { returnTo });
+      await authClient.startInteractiveLogin(
+        new Auth0NextResponse(new NextResponse()),
+        { returnTo }
+      );
 
       expect(authClient["transactionStore"].save).toHaveBeenCalled();
     });
@@ -8142,12 +8161,15 @@ ca/T0LLtgmbMmxSv/MmzIg==
         return originalAuthorizationUrl.call(authClient, params);
       });
 
-      await authClient.startInteractiveLogin(new Auth0NextResponse(new NextResponse()), {
-        authorizationParameters: {
-          scope: methodScope,
-          audience: methodAudience
+      await authClient.startInteractiveLogin(
+        new Auth0NextResponse(new NextResponse()),
+        {
+          authorizationParameters: {
+            scope: methodScope,
+            audience: methodAudience
+          }
         }
-      });
+      );
 
       expect(authClient["authorizationUrl"]).toHaveBeenCalled();
     });
--- a/src/server/auth-client.test.ts
+++ b/src/server/auth-client.test.ts
@@ -22,6 +22,7 @@ import {
 import { DEFAULT_SCOPES } from "../utils/constants.js";
 import { AuthClient } from "./auth-client.js";
 import { decrypt, encrypt } from "./cookies.js";
+import { Auth0NextRequest, Auth0NextResponse } from "./http/index.js";
 import { StatefulSessionStore } from "./session/stateful-session-store.js";
 import { StatelessSessionStore } from "./session/stateless-session-store.js";
 import { TransactionState, TransactionStore } from "./transaction-store.js";
@@ -443,7 +444,7 @@ ca/T0LLtgmbMmxSv/MmzIg==
       const request = new NextRequest("https://example.com/auth/login", {
         method: "GET"
       });
-      authClient.handleLogin = vi.fn();
+      authClient.handleLogin = vi.fn().mockResolvedValue({});
       await authClient.handler(request);
       expect(authClient.handleLogin).toHaveBeenCalled();
     });
@@ -474,7 +475,7 @@ ca/T0LLtgmbMmxSv/MmzIg==
       const request = new NextRequest("https://example.com/auth/callback", {
         method: "GET"
       });
-      authClient.handleCallback = vi.fn();
+      authClient.handleCallback = vi.fn().mockResolvedValue({});
       await authClient.handler(request);
       expect(authClient.handleCallback).toHaveBeenCalled();
     });
@@ -505,7 +506,7 @@ ca/T0LLtgmbMmxSv/MmzIg==
       const request = new NextRequest("https://example.com/auth/logout", {
         method: "GET"
       });
-      authClient.handleLogout = vi.fn();
+      authClient.handleLogout = vi.fn().mockResolvedValue({});
       await authClient.handler(request);
       expect(authClient.handleLogout).toHaveBeenCalled();
     });
@@ -536,7 +537,7 @@ ca/T0LLtgmbMmxSv/MmzIg==
       const request = new NextRequest("https://example.com/auth/profile", {
         method: "GET"
       });
-      authClient.handleProfile = vi.fn();
+      authClient.handleProfile = vi.fn().mockResolvedValue({});
       await authClient.handler(request);
       expect(authClient.handleProfile).toHaveBeenCalled();
     });
@@ -568,7 +569,7 @@ ca/T0LLtgmbMmxSv/MmzIg==
       const request = new NextRequest("https://example.com/auth/access-token", {
         method: "GET"
       });
-      authClient.handleAccessToken = vi.fn();
+      authClient.handleAccessToken = vi.fn().mockResolvedValue({});
       await authClient.handler(request);
       expect(authClient.handleAccessToken).toHaveBeenCalled();
     });
@@ -600,7 +601,7 @@ ca/T0LLtgmbMmxSv/MmzIg==
       const request = new NextRequest("https://example.com/auth/access-token", {
         method: "GET"
       });
-      authClient.handleAccessToken = vi.fn();
+      authClient.handleAccessToken = vi.fn().mockResolvedValue({});
       const response = await authClient.handler(request);
       expect(authClient.handleAccessToken).not.toHaveBeenCalled();
       // When a route doesn't match, the handler returns a NextResponse.next() with status 200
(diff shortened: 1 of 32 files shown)


src/server/auth-client.test.ts
```diff
@@ -443,7 +444,7 @@ ca/T0LLtgmbMmxSv/MmzIg==
       const request = new NextRequest("https://example.com/auth/login", {
         method: "GET"
       });
-      authClient.handleLogin = vi.fn();
+      authClient.handleLogin = vi.fn().mockResolvedValue({});
```
frederikprijck, line 447: The original mock for`handleX()` is actually not correctly configured. Every `handleX()` method is configured not to return undefined, while `vi.fn()` on the left does return undefined.

This worked fine for this test, as we never expected any response and `handler()` just returned whatever `handleX()` returns.

However, `handleX()` now returns an `Auth0Response`, which `handler()` now unwraps using `#unwrapHandler()`. Because of that, `handler()` expects `handleX` to return something.

In this case, we return `{}`, which would mean the unwrapped `NextResponse` will be undefined, just like the original test.

src/server/auth-client.test.ts
```diff
@@ -1262,7 +1263,13 @@ ca/T0LLtgmbMmxSv/MmzIg==
...
       );
 
-      const response = await authClient.handleLogin(request);
+      const auth0Req = new Auth0NextRequest(request);
+
+      const auth0Res = new Auth0NextResponse(NextResponse.next());
+
+      await authClient.handleLogin(auth0Req, auth0Res);
+
+      const response = auth0Res.res;
```
frederikprijck, line 1272: The original tests call `handleX()`. In my opinion, `handleX` is an implementation detail. It's the handler called by `handler()` for the corresponding route (e.g. `/auth/login` for `handleLogin()`). In my opinion, this should have been a private method, and call `handler()` instead in the tests.

If this test would have called `const response = await authClient.handler(request);` instead of `const response = await authClient.handleLogin(request);`, this would have not required any change at all.

Personally, I think we should change this test to go through the `handler()`, and treat `handleX`  as an implementation detail.

Happy to update the PR, or do a follow up PR. Additionally, also happy to leave the tests as is, as they are public methods, so they aren't configured as the real implementation details I consider them.

src/server/auth-client.ts
```diff
@@ -389,7 +396,10 @@ export class AuthClient {
   }
 
   async handler(req: NextRequest): Promise<NextResponse> {
-    let { pathname } = req.nextUrl;
+    const auth0Req = new Auth0NextRequest(req);
+    const auth0Res = new Auth0NextResponse(new NextResponse());
```
frederikprijck, line 399: As `handler()` is our main public API, I ensured this does not require any change. Instead, the first thing it does now is create the `Auth0NextRequest` instance, holding the original `NextRequest` instance.

Additionally, it also prepares an `Auth0NextResponse`, which contains a placeholder response.

This change for `Auth0NextResponse` is a bit biased by the fact that I know we want to introduce support for `NextApiResponse` in a next PR. NextResponse and NextApiResponse are very different in usage:

- `NextResponse`: Next.js expects you to create and return the instance.
- `NextApiResponse`: Next.js gives you the NextApiResponse for the NextApiRequest, and expects you to `mutate` it.

Because of that knowledge, the `Auth0Request` and `Auth0Response` classes already account for this.

Admittedly, this bias comes from the fact that this PR is created by pulling out the `Auth0Request`, `Auth0Response` and Middleware / AppRouter specific logic from a single branch that contains the full integration with `PagesRouter` and `AppRouter` without using middleware.

src/server/auth-client.ts
```diff
@@ -389,7 +396,10 @@ export class AuthClient {
...
+
+    let { pathname } = auth0Req.getUrl();
```
frederikprijck, line 401: We no longer use `nextUrl`, instead we go through the `Auth0Request.getUrl` method to handle the abstraction. Internally, for `NextRequest`, it will return `nextUrl`.

src/server/auth-client.ts
```diff
@@ -401,57 +411,69 @@ export class AuthClient {
     }
 
     const sanitizedPathname = removeTrailingSlash(pathname);
-    const method = req.method;
+    const method = auth0Req.getMethod();
```
frederikprijck, line 413: We no longer use `.method`, instead we go through the `Auth0Request.getMethod()` method to handle the abstraction. Internally, for `NextRequest`, it will return `method`.

src/server/auth-client.ts
```diff
@@ -401,57 +411,69 @@ export class AuthClient {
...
 
     if (method === "GET" && sanitizedPathname === this.routes.login) {
-      return this.handleLogin(req);
+      return this.#unwrapHandler(() => this.handleLogin(auth0Req, auth0Res));
```
frederikprijck, line 416: `handleX()` no longer returns a `NextResponse`. Instead, it returns an `Auth0NextResponse`. Instead of returning whatever `handleX()` returns, we have to unwrap the `res` property, containing the `NextResponse` instance, and return that one.

Because of the reason mentioned [here](https://github.com/auth0/nextjs-auth0/pull/2484/files#r2682066147), every `handleX()` method now also accepts an `Auth0Response` instance to do response manipulations directly on the provided `Auth0Response` instance.

src/server/auth-client.ts
```diff
@@ -401,57 +411,69 @@ export class AuthClient {
...
-      return this.handleMyOrg(req);
+      return this.#unwrapHandler(() => this.handleMyOrg(auth0Req, auth0Res));
     } else {
       // no auth handler found, simply touch the sessions
-      // TODO: this should only happen if rolling sessions are enabled. Also, we should
-      // try to avoid reading from the DB (for stateful sessions) on every request if possible.
-      const res = NextResponse.next();
-      const session = await this.sessionStore.get(req.cookies);
-
+      const session = await this.sessionStore.get(auth0Req.getCookies());
```
frederikprijck, line 456: Same as `nextUrl`, and `method`, we should not try and read the cookies directly from `NextRequest`. Instead, we read them through `Auth0Request`, handling the `NextRequest` specific cookie handling.

src/server/auth-client.ts
```diff
@@ -401,57 +411,69 @@ export class AuthClient {
...
-        });
-        addCacheControlHeadersForSession(res);
+        await this.sessionStore.set(
+          auth0Req.getCookies(),
+          auth0Res.getCookies(),
+          {
+            ...session
+          }
+        );
+        auth0Res.addCacheControlHeadersForSession();
```
frederikprijck, line 468: As manipulating headers differs between `NextResponse` and `NextApiResponse`, we should call this from `Auth0Response` instead.

src/server/auth-client.ts
```diff
@@ -401,57 +411,69 @@ export class AuthClient {
...
+      return this.#unwrapHandler(() => this.handleMyOrg(auth0Req, auth0Res));
     } else {
       // no auth handler found, simply touch the sessions
-      // TODO: this should only happen if rolling sessions are enabled. Also, we should
-      // try to avoid reading from the DB (for stateful sessions) on every request if possible.
-      const res = NextResponse.next();
-      const session = await this.sessionStore.get(req.cookies);
-
+      const session = await this.sessionStore.get(auth0Req.getCookies());
+      const auth0Res = new Auth0NextResponse(NextResponse.next());
```
frederikprijck, line 457: In this case, we just want to continue the next middleware, so we need to call `NextResponse.next()`, but wrapped in a `new Auth0NextResponse`.

Important: this is still bound to `NextResponse`. But this is fine as this is middleware only. We will need solve this in the. follow up PR to add support for using the SDK without middleware.

src/server/auth-client.ts
```diff
@@ -401,57 +411,69 @@ export class AuthClient {
...
+          auth0Res.getCookies(),
+          {
+            ...session
+          }
+        );
+        auth0Res.addCacheControlHeadersForSession();
       }
 
-      return res;
+      return auth0Res.res;
```
frederikprijck, line 471: We need to return the underlying`NextResponse`.

src/server/auth-client.ts
```diff
@@ -401,57 +411,69 @@ export class AuthClient {
...
   }
 
+  /**
+   * Unwraps an Auth0Response by extracting the underlying NextResponse.
+   * This utility simplifies the pattern of awaiting handler calls and accessing .res
+   */
+  async #unwrapHandler(handler: () => Promise<Auth0Response>): Promise<NextResponse> {
+    const auth0Response = await handler();
+    return auth0Response.res;
+  }
```
frederikprijck: Currently, this only returns `Promise<NextResponse>`. This is fine for now, but in the follow-up PR that adds support for mounting the auth routes without middelware, we will change `unwrapHandler` to unwrap based on the context it's used in.

src/server/auth-client.ts
```diff
@@ -401,60 +411,75 @@ export class AuthClient {
...
+
   async startInteractiveLogin(
+    auth0Res: Auth0Response,
```
frederikprijck, line 487: This is the internal method, not exposed to our users. This is used by `server/client.ts`, the instance exposed to the user.

src/server/auth-client.ts
```diff
@@ -616,67 +651,82 @@ export class AuthClient {
...
+      res.redirect(url.toString());
     };
 
     // Determine logout strategy and create appropriate response
-    let logoutResponse: NextResponse;
-
     if (this.logoutStrategy === "v2") {
       // Always use v2 logout endpoint
-      logoutResponse = createV2LogoutResponse();
+      createV2LogoutResponse(auth0Res);
```
frederikprijck, line 667: Instead of returning a new instance, we have to call methods on the Auth0Response.

src/server/auth-client.ts
```diff
@@ -616,67 +651,82 @@ export class AuthClient {
...
     const transactionStateCookie = await this.transactionStore.get(
-      req.cookies,
+      auth0Req.getCookies(),
       state
     );
     if (!transactionStateCookie) {
-      return this.onCallback(new InvalidStateError(), {}, null);
+      const errorRes = await this.onCallback(new InvalidStateError(), {}, null);
+      auth0Res.setResponse(errorRes);
+      return auth0Res;
```
frederikprijck, line 731: This is a bit of a weird thing, but `onCallback` is a property that gets set to the user-provided `options.onCallback`, or used the `defaultOnCallback`. It's configured to return a `NextResponse`.

To make this as generic as possible, the `Auth0Response` contains a `setResponse()` that allows you to just set the underlying response.

src/server/auth-client.ts
```diff
@@ -748,9 +801,10 @@ export class AuthClient {
         session
       );
 
-      await this.transactionStore.delete(res.cookies, state);
+      auth0Res.setResponse(res);
```
frederikprijck, line 803: Same as https://github.com/auth0/nextjs-auth0/pull/2484/files#r2682265151

src/server/auth-client.ts
```diff
@@ -894,41 +944,50 @@ export class AuthClient {
     // if not then filter id_token claims with default rules
     session = await this.finalizeSession(session, oidcRes.id_token);
 
-    await this.sessionStore.set(req.cookies, res.cookies, session, true);
-    addCacheControlHeadersForSession(res);
+    auth0Res.setResponse(res);
```
frederikprijck, line 950: Same as https://github.com/auth0/nextjs-auth0/pull/2484/files#r2682265151

src/server/chunked-cookies.test.ts
```diff
@@ -95,18 +97,28 @@ const createMocks = () => {
...
+  let auth0ReqCookies: Auth0RequestCookies;
+  let auth0ResCookies: Auth0ResponseCookies;
   let cookieStore: Map<any, any>;
 
   beforeEach(() => {
     const mocks = createMocks();
     reqCookies = mocks.reqCookies;
     resCookies = mocks.resCookies;
+    auth0ReqCookies = mocks.auth0ReqCookies;
+    auth0ResCookies = mocks.auth0ResCookies;
```
frederikprijck, line 121: We need both the original cookies, and the Auth0Cookies, because we need to pass Auth0Cookies instances down to setChunkedCookie, but we want to still expect the `set` method to have been called on the original `cookie` instance.

src/server/client.ts
```diff
@@ -1018,7 +1035,11 @@ export class Auth0Client {
   async startInteractiveLogin(
     options: StartInteractiveLoginOptions = {}
   ): Promise<NextResponse> {
-    return this.authClient.startInteractiveLogin(options);
+    const auth0Res = await this.authClient.startInteractiveLogin(
+      new Auth0NextResponse(new NextResponse()),
```
frederikprijck, line 1039: startInteractiveLogin needs to set a cookie on the response. Before this PR, it would internally create and return NextResponse instances. However, we are changing this with this PR by providing the Auth0NextResponse object, wrapping a placeholder NextResponse, and have the startInteractive call corresponding method on the Auth0NextResponse instance (e.g. redirect(), as well as getCookies() to retrieve the cookies for the response, allowing the SDK to add a cookie as needed.

This is mostly changed with the knowledge that for Pages Router, Next.js will provide the NextApiResponse initially together with the NextApiRequest. But for the App Router, we only receive a NextRequest, and we should instantiate and return NextResponse ourselves. To unify this, we now always expect an Auth0Response implementation.

PR #2488, opened 2026-01-13
Expose default onCallback

greenlynx:
- [✅] All new/changed/fixed functionality is covered by tests (or N/A)
- [✅] I have added documentation for all new/changed functionality (or N/A)

### 📋 Changes

We would like to define an `onCallback` hook in our code that takes an action but does not override default callback behaviour. This is currently difficult. This PR sends an extra `defaultOnCallback` parameter to `onCallback` that can be called if default callback behaviour is desired.

### 🎯 Testing

Automated unit tests added.

### DIFF
--- a/src/server/auth-client.ts
+++ b/src/server/auth-client.ts
@@ -118,7 +118,11 @@ export type OnCallbackContext = {
 export type OnCallbackHook = (
   error: SdkError | null,
   ctx: OnCallbackContext,
-  session: SessionData | null
+  session: SessionData | null,
+  defaultOnCallback: (
+    error: SdkError | null,
+    ctx: OnCallbackContext
+  ) => Promise<NextResponse>
 ) => Promise<NextResponse>;
 
 // params passed to the /authorize endpoint that cannot be overwritten
@@ -676,7 +680,12 @@ export class AuthClient {
       state
     );
     if (!transactionStateCookie) {
-      return this.onCallback(new InvalidStateError(), {}, null);
+      return this.onCallback(
+        new InvalidStateError(),
+        {},
+        null,
+        this.defaultOnCallback
+      );
     }
 
     const transactionState = transactionStateCookie.payload;
@@ -745,7 +754,8 @@ export class AuthClient {
           ...onCallbackCtx,
           connectedAccount
         },
-        session
+        session,
+        this.defaultOnCallback
       );
 
       await this.transactionStore.delete(res.cookies, state);
@@ -888,7 +898,12 @@ export class AuthClient {
       }
     };
 
-    const res = await this.onCallback(null, onCallbackCtx, session);
+    const res = await this.onCallback(
+      null,
+      onCallbackCtx,
+      session,
+      this.defaultOnCallback
+    );
 
     // call beforeSessionSaved callback if present
     // if not then filter id_token claims with default rules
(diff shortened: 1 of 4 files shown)


src/server/auth-client.ts
```diff
@@ -118,7 +118,11 @@ export type OnCallbackContext = {
 export type OnCallbackHook = (
```
Piyush-85, line 118: Could we add JSDoc on type parameters, IDEs (VS Code, WebStorm) surface these on hover. Something like below should be good.
```
/**
 * Hook called once the user has been redirected back from Auth0 to your
 * application's callback route, after the authorization code has been
 * exchanged (or an error has occurred).
 *
 * Return a `NextResponse` to control where the user is sent next.
 * If you do not need custom routing, simply return the result of calling
 * the fourth argument (`defaultOnCallback`) to invoke the SDK's default behaviour.
 *
 * @param error - The error returned from Auth0 or during token exchange.
 *   `null` when the callback completed successfully.
 * @param ctx - Context about the transaction that initiated the auth flow
 *   (e.g. `returnTo` URL, `responseType`, `challengeMode`).
 * @param session - The `SessionData` that will be persisted after a
 *   successful callback. `null` when `error` is present.
 * @param defaultOnCallback - The SDK's built-in callback handler. Call this
 *   to keep the default redirect/error behaviour while still running your own
 *   side effects (e.g. syncing the user to your database, setting extra
 *   cookies). Signature: `(error, ctx) => Promise<NextResponse>`.
 *
 * @example
 * ```ts
 * const auth0 = new Auth0Client({
 *   async onCallback(error, ctx, session, defaultOnCallback) {
 *     if (!error && session) {
 *       await db.upsertUser(session.user); // side effect
 *     }
 *     return defaultOnCallback(error, ctx); // keep default redirect behaviour
 *   }
 * });
 * ```
 */
```

src/server/auth-client.ts
```diff
@@ -676,7 +680,12 @@ export class AuthClient {
       state
     );
     if (!transactionStateCookie) {
-      return this.onCallback(new InvalidStateError(), {}, null);
+      return this.onCallback(
+        new InvalidStateError(),
+        {},
+        null,
+        this.defaultOnCallback
```
Piyush-85, line 687: `this.defaultOnCallback` is passed as a plain function reference. Since it's a regular method (not an arrow function), this won't be bound when the consumer calls it. 
Please use `this.defaultOnCallback.bind(this)` at each callsite,  so if defaultOnCallback ever gets updated to call `this.someInternalHelper()`, it does not silently break.

Piyush-85:
Hi @greenlynx, thanks for your contribution - this is a useful addition. 
The change looks good. Two minor comments to address before we merge, please take a look when you get a chance, and also rebase to resolve the merge conflicts.

Piyush-85:
Hey @greenlynx — thanks for taking the time to put this together.
                                                                                                              
  After reviewing this more carefully, we've decided not to move ahead with exposing defaultOnCallback as a
  parameter. The default callback behaviour is intentionally simple,  a redirect to returnTo on success and a 
  500 on error and users who need side effects can replicate that in a few lines, as shown in our existing
  docs example.                                                                                               
                                                                                                       
  The bigger concern is the popup flow: defaultOnCallback would produce a redirect response there, which is   
  the wrong behaviour for a postMessage flow. Making the contract accurate for popup would add meaningful
  complexity for a convenience gain that doesn't justify it.                                                  
                                                                                                       
  Closing this out, but appreciate the contribution.

Issue #2454, opened 2025-12-10
Make the request available to onCallback


dstapleton92:
### Checklist

- [x] I have looked into the [Readme](https://github.com/auth0/nextjs-auth0#readme), [Examples](https://github.com/auth0/nextjs-auth0/blob/main/EXAMPLES.md), and [FAQ](https://github.com/auth0/nextjs-auth0/blob/main/FAQ.md) and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://auth0.github.io/nextjs-auth0/) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/nextjs-auth0/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

Due to how we implemented step-up auth for MFA management on v3, we need the ability to access the "previous session" in `onCallback`, so that we can set the new session to be a combination of the previous and incoming session, with an mfa access token set as a new property onto the "old session" and persisted. The signature of `onCallback` does not expose the `request`, so we are unable to call `getSession` within `onCallback` to accomplish this.

Example v3 code, using `afterCallback`:
```typescript
export const afterCallback = async (
 req: NextApiRequest,
 res: NextApiResponse,
 session: Session,
 state: LoginOptions | undefined
) => {
 // Previous session is null on initial login
 const prevSession = await getSession(req, res)


 if (prevSession && session.accessTokenScope?.includes('someParticularScope')) {
     prevSession.mfaAccessToken = {
       accessToken: session.accessToken,
       accessTokenScope: session.accessTokenScope,
       accessTokenExpiresAt: session.accessTokenExpiresAt,
     }
     return prevSession
  }
  ...
}
```

Note: We established this pattern with guidance from Auth0 Professional Services when onboarding a few years back.

### Describe the ideal solution

I know that `handleCallback` receives the request. Our issue could be resolved by either:
* Having `handleCallback` retrieve the "previous session" and pass it into onCallback. This would be pretty niche and would probably not be useful for most users.
[...]

PR #2489, opened 2026-01-13
feat: add support for mounting auth routes as api routes

frederikprijck:
### 📋 Changes

This PR adds support for mounting the authentication routes with both Next.js' App Router and Pages Router, enabling developers to use our SDK without handling these in the middleware.

#### Usage

This PR introduces a new `handleAuth` method that works seamlessly with both routing patterns:

**App Router usage:**
```typescript
// app/api/auth/[...auth0]/route.ts
import { auth0 } from "@/lib/auth0";

export const GET = auth0.handleAuth;
export const POST = auth0.handleAuth;
```

**Pages Router usage:**
```typescript
// pages/api/auth/[...auth0].ts
import { auth0 } from "@/lib/auth0";

export default auth0.handleAuth;
```

#### Technical Changes

- **Enhanced `AuthClient.handler` (internal) method** with method overloading to support both `NextRequest` (App Router/Middleware) and `NextApiRequest` (Pages Router)
- **Added new HTTP adapters** (`Auth0NextApiRequest`, `Auth0NextApiResponse`, `Auth0ApiResponseCookies`) to handle Pages Router request/response objects
- **Extended `OnCallbackHook`** context to include `request` and `response` objects for Pages Router compatibility
- **Added runtime detection** using `isRequest()` to automatically determine the routing pattern and use appropriate handlers

### 📎 References

N/A

### DIFF
--- a/src/server/auth-client.ts
+++ b/src/server/auth-client.ts
@@ -1,3 +1,4 @@
+import { NextApiRequest, NextApiResponse } from "next";
 import { NextResponse, type NextRequest } from "next/server.js";
 import * as jose from "jose";
 import * as oauth from "oauth4webapi";
@@ -71,6 +72,7 @@ import {
   buildForwardedResponseHeaders,
   transformTargetUrl
 } from "../utils/proxy.js";
+import { isRequest } from "../utils/request.js";
 import {
   ensureDefaultScope,
   getScopeForAudience
@@ -83,6 +85,7 @@ import {
   tokenSetFromAccessTokenSet
 } from "../utils/token-set-helpers.js";
 import { toSafeRedirect } from "../utils/url-helpers.js";
+import { unwrapHandler } from "./auth-client.utils.js";
 import {
   AccessTokenFactory,
   Fetcher,
@@ -91,6 +94,8 @@ import {
   FetcherMinimalConfig
 } from "./fetcher.js";
 import {
+  Auth0NextApiRequest,
+  Auth0NextApiResponse,
   Auth0NextRequest,
   Auth0NextResponse,
   Auth0Request,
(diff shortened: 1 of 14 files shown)


src/server/auth-client.ts
```diff
@@ -120,12 +124,27 @@ export type OnCallbackContext = {
...
+   * which we do not have initially and should be created and returned instead.
+   */
+  response?: NextApiResponse;
+
+  /**
+   * The request object (NextRequest for App Router or Middleware on Edge, NextApiRequest for Pages Router, or Request for Middleware or Node.js or Proxy).
+   *
+   * @remark This property is undefined when using the App Router.
+   */
+  request?: NextRequest | NextApiRequest | Request;
```
frederikprijck, line 142: In App Router, users can return a `NextResponse` instance from `onCallback`. However, in Pages Router, the user should mutate the original `NextApiResponse`, so we need to expose `response` when using Pages Router.

Exposing `request` made things a bit simpler for the defaultOnCallback we have, and has been requested [here](https://github.com/auth0/nextjs-auth0/issues/2454), so feels like a useful thing to consider.

src/server/client.ts
```diff
@@ -497,15 +502,91 @@ export class Auth0Client {
...
+        undefined,
+        res as NextApiResponse
+      );
+    }
+  }
+
+  /**
+   * Mounts the SDK proxy routes to run as a proxy function.
+   */
+  async proxy(req: NextRequest) {
```
frederikprijck, line 586: As our SDK has a capability to expose certain proxy endpoints, we keep these as part of the proxy integration.

When using our middleware, it would mount both the authentication routes and the proxy routes automatically. With this PR, authentication routes can be mounted through the API routes with either Pages or App Router, which does not include the proxy functionality. That needs to be configured in the customers `proxy.ts` file, by calling `handleProxy()`.

Issue #2428, opened 2025-11-20
Make access token refresh configurable


dnbkr:
### Checklist

- [x] I have looked into the [Readme](https://github.com/auth0/nextjs-auth0#readme), [Examples](https://github.com/auth0/nextjs-auth0/blob/main/EXAMPLES.md), and [FAQ](https://github.com/auth0/nextjs-auth0/blob/main/FAQ.md) and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://auth0.github.io/nextjs-auth0/) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/nextjs-auth0/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

The frontend of my NextJS app is fetching an access token via `/auth/token` (the route provided by this library) and keeping it in memory, and creating a setTimeout to proactively refetch the token from that route once the one in memory expires. I would like to refresh my token a little bit before the token actually expires to create a safe buffer.

Can I suggest a new url parameter, `refresh-buffer`, which would allow me to express a number of seconds. If the token in the current session is within that number of seconds to expiry, and refresh token is available, then the token would refresh.

### Describe the ideal solution

```
/auth/token?scope=openid+offline_access&refresh-buffer=2
```

this would mean that if my current token was within 2 seconds of expiring then it would be refreshed

nandan-bhat:
Hi @dnbkr ,

Thanks for posting this issue.

You can do this today by adding a custom endpoint that calls `auth0.getAccessToken with { refresh: true }`. Call this endpoint a few seconds before the token expires, so the token refreshes early even if it has not expired yet (as long as a refresh token exists). 

I will also discuss the feature request with the team and share an update on our decision.

nandan-bhat:
@dnbkr,
 
We looked into your request and decided to implement it in a slightly different way. This PR has been merged: https://github.com/auth0/nextjs-auth0/pull/2508.

With this change, you should be able to achieve what you need. The feature will be available in the next release. I’m closing this ticket for now.

If you’d like to reopen it or discuss anything further, please let me know.

PR #2508, opened 2026-01-28
feat: added tokenRefreshBuffer for early access‑token refreshes

nandan-bhat:
### Changes
- Added `tokenRefreshBuffer` (`seconds`) to `Auth0Client` config.
- `getAccessToken` now refreshes when a token is within the `buffer` window and a refresh token is available.
- Tests cover early refresh, no‑refresh outside buffer, missing refresh token cases, and invalid config.

### Fixes
- #2428

### DIFF
--- a/src/server/auth-client.ts
+++ b/src/server/auth-client.ts
@@ -166,6 +166,8 @@ export type OnCallbackHook = (
   session: SessionData | null
 ) => Promise<NextResponse>;
 
+type ExpiresAtInput = number | string | null | undefined;
+
 // params passed to the /authorize endpoint that cannot be overwritten
 const INTERNAL_AUTHORIZE_PARAMS = [
   "client_id",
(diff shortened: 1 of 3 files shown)


src/server/auth-client.ts
```diff
@@ -1417,6 +1420,13 @@ export class AuthClient {
         audience: options.audience ?? this.authorizationParameters.audience
       }
     );
+    const now = Date.now() / 1000;
+    const expiresAt =
+      typeof tokenSet.expiresAt === "number" ? tokenSet.expiresAt : undefined;
+    const isExpired = typeof expiresAt === "number" && expiresAt <= now;
+    const isWithinRefreshWindow =
+      typeof expiresAt === "number" &&
+      expiresAt <= now + this.tokenRefreshBuffer;
```
frederikprijck: nit: I think we can clean this up a bit.

We can create a utility.
```ts
const isBeforeOrEqual = (left: number | undefined, right: number) => {
    return left !== undefined && left <= right;
}
```

Then, here we can replace line 1423 until 1429 with:

```ts
const isExpired = isBeforeOrEqual(tokenSet.expiresAt, now);
```
And then below we use `tokenSet.expiresAt` instead of `expiresAt`.

Then, inside `if (tokenSet.refreshToken)`, we can add something like this.

```ts
const shouldRefresh = isBeforeOrEqual(tokenSet.expiresAt, now + this.tokenRefreshBuffer);
```

PR #2524, opened 2026-02-16
feat: add mfa.stepUpWithPopup() for reactive MFA step-up via Universal Login

tusharpandey13:
Adds client-side `mfa.stepUpWithPopup()` that completes MFA step-up in a popup window using Auth0 Universal Login, without a full-page redirect.

How it works: When `getAccessToken()` throws `MfaRequiredError`, the app calls `mfa.stepUpWithPopup({ audience })`. A popup opens, the user completes MFA, and the resulting token is cached in the session and returned to the caller.

## Changes

- Added `mfa.stepUpWithPopup()` to client helpers
- Added `challengeMode?: "redirect" | "popup"` to `StartInteractiveLoginOptions`
- Added popup mode handling to `handleCallback()`
- Updated `getTokenSet`, `handleAccessToken` to enable conditional scope merging
- Added tests for all flows
- Updated `EXAMPLES.md` with clear documentation.

## Usage

```js
try {
  const token = await getAccessToken({ audience: 'https://api.example.com' });
} catch (err) {
  if (err instanceof MfaRequiredError) {
    // Trigger popup MFA step-up
    const { token } = await mfa.stepUpWithPopup({
      audience: 'https://api.example.com'
    });
  }
}
```


gyaneshgouraw-okta:
@copilot+claude-opus-4.6 Review changes in this PR

src/client/helpers/get-access-token.ts
```diff
@@ -57,13 +57,33 @@ export type AccessTokenOptions = {
...
    */
   includeFullResponse?: boolean;
+
+  /**
+   * Control scope merging behavior server-side.
+   * When true (default): merge global scopes for default audience.
+   * When false: use ONLY requested scope (no global merge).
+   * Passed as query param to /auth/access-token endpoint.
+   */
+  mergeScopes?: boolean;
```
gyaneshgouraw-okta, line 85: @tusharpandey13 , I see we have introduced `mergeScopes` a new property for the `getAccessToken` method. This appears to be needed only internally by `challengeWithPopup()`  to prevent scope
 cache misses during token retrieval. 

 Is this absolutely necessary to introduce this for `challengeWithPopup` use-case. Did we consider alternatives that would avoid expanding the public API surface? 

Either way If this is required for `challengeWithPopup` use-case.
Will it be worth it documenting the use case in `Example.md` under corresponding section ?

src/server/auth-client.ts
```diff
@@ -1078,22 +1136,30 @@ export class AuthClient {
...
+      // processAuthorizationCodeResponse would throw AuthorizationCodeGrantError.
       oidcRes = await oauth.processAuthorizationCodeResponse(
         authorizationServerMetadata,
         this.clientMetadata,
         codeGrantResponse,
         {
           expectedNonce: transactionState.nonce,
           maxAge: transactionState.maxAge,
-          requireIdToken: true
+          requireIdToken: challengeMode !== "popup"
```
gyaneshgouraw-okta, line 1162: @tusharpandey13 For `requireIdToken: challengeMode !== "popup"` will there security concerns with skipping ID token validation completely for popup flows?

EXAMPLES.md
```diff
@@ -3705,10 +3719,191 @@ The SDK provides typed error classes for all MFA operations:
...
+
+export const auth0 = new Auth0Client({
+  cspNonce: 'your-generated-nonce'
+});
+```
+
+The nonce is injected into the `<script>` tag of the popup callback HTML response, making it compliant with `script-src 'nonce-...'` CSP policies.
+
+> [!IMPORTANT]
+> The nonce must contain only base64 characters (`A-Za-z0-9+/=-_`). Invalid characters will throw an `InvalidConfigurationError`. The nonce should be generated per-request on the server side for maximum security.
```
gyaneshgouraw-okta: @tusharpandey13 here we are recommending "_The nonce should be generated **per-request** on the server side for maximum security_"
 However `cspNonce` is set once in the constructor and the `Auth0Client` is typically a singleton. How do you see users providing a dynamic per-request nonce with this API? 

I see this is the current way of setting the nonce
```ts
  cspNonce: 'your-generated-nonce'
});
```

gyaneshgouraw-okta:
The PR in general looks well structured. Have added few minor comments.

gyaneshgouraw-okta (dismissed):

src/client/helpers/get-access-token.ts
tusharpandey13: `mergeScopes` is used internally by `challengeWithPopup()`, but it has legitimate value as a public option beyond that use case — any MRRT scenario where a developer stores tokens for multiple audiences with specific scopes benefits from exact-scope cache lookups without global scope pollution.
 
Keeping it public gives advanced users control over cache matching behavior (e.g., retrieving a token that was cached with `read:sensitive` without having openid profile email merged into the lookup key, which would cause a cache miss).
 
I've improved the JSDoc with a usage example and clearer explanation of when to use it.

src/server/auth-client.ts
tusharpandey13: No security concern here. The popup flow targets API-only audiences for step-up tokens — Auth0 does not always return an ID token for these (per OIDC spec, ID tokens are optional when the openid scope isn't the primary purpose).

The security model doesn't rely on the ID token in this path because:
1. PKCE protects the authorization code exchange
2. State parameter (encrypted in transaction cookie) prevents CSRF
3. Existing session required — the popup merges into an already-authenticated session (user's identity was validated via ID token during initial login)
4. No client exposure — the token stays server-side in the encrypted session cookie; `postMessage` only sends `{ sub, email }` metadata

If an ID token IS returned (e.g., openid scope was included), it's still validated via `getValidatedIdTokenClaims()` — we just don't require its presence.

EXAMPLES.md
tusharpandey13: Updated the docs to be transparent about this tradeoff:

- A static nonce still protects against 'unsafe-inline' (attacker must know the nonce value)
- It's weaker than per-request rotation, but sufficient for most apps since the popup callback HTML is a controlled, minimal page
- For apps requiring per-request nonces, the docs now suggest creating the client per-request or using middleware injection

Making cspNonce accept a () => string function is a possible future enhancement but adds complexity for a narrow use case.

src/server/auth-client.ts
```diff
@@ -1127,29 +1198,105 @@ export class AuthClient {
...
-        ...(this.provider?.isResolverMode && {
-          mcd: {
-            domain: this.domain,
-            issuer: this.issuer
-          }
-        })
+    // ★ POSTMESSAGE BRANCH
+    if (challengeMode === "popup") {
+      // Merge new token into existing session (do NOT create fresh session)
+      const existingSession = await this.sessionStore.get(req.cookies);
```
nandan-bhat: Should this popup path use `getSessionWithDomainCheck(req.cookies)` instead of `this.sessionStore.get(req.cookies)` ?

In `MCD` mode, this branch seems to skip the domain mismatch check. It also seems to skip the normal `internal.mcd` add/validate path before saving the session. 

Could this let a popup token be merged into a session from a different domain ? Can you confirm ?

src/client/mfa/index.ts
```diff
@@ -374,6 +454,133 @@ class ClientMfaClient implements MfaClient {
...
+    // When omitted, startInteractiveLogin uses the global scope config,
+    // so transactionState.scope matches what a default getAccessToken()
+    // lookup computes — preventing cache misses on subsequent calls.
+    // See session-helpers.ts for detailed explanation of why
+    // requestedScope must use transactionState.scope (not oidcRes.scope).
+    if (options.scope) {
+      params.set("scope", options.scope);
+    }
+
+    const loginUrl =
```
nandan-bhat, line 541: Should `challengeWithPopup()` use `process.env.NEXT_PUBLIC_LOGIN_ROUTE || "/auth/login"` here instead of hardcoding `"/auth/login"` ?

We already support custom login routes in other places. Could popup mode break for apps that use a custom login route ?

src/server/auth-client.ts
```diff
@@ -1127,29 +1198,105 @@ export class AuthClient {
...
+
+      // Call onCallback for side-effects (logging, analytics) - discard redirect response.
```
nandan-bhat: Should popup flows keep the same `onCallback` behavior as normal callback flows ?

In the popup success path, `onCallback` runs before the popup token is merged and before the session is finalized, so the hook does not seem to get the session that will actually be saved. 

In the popup error path (refer the line 2104), `onCallback` does not seem to run at all. Is that intended ?

src/server/auth-client.ts
tusharpandey13: Yes
In MCD mode `this.sessionStore.get(req.cookies)` bypasses domain isolation. A popup callback on `auth.brand2.com` could merge into a session created by `auth.brand1.com`.
Updated to use `getSessionWithDomainCheck` and handle the domain mismatch error as a popup error response.

src/client/mfa/index.ts
tusharpandey13: Yes, updated

src/server/auth-client.ts
tusharpandey13: 1. Updated
2. Not intended, updated

nandan-bhat (approved):

nandan-bhat:
LGTM

PR #2530, opened 2026-02-20
fix: prevent rolling session from re-creating a deleted session on concurrent logout (#2335)

### LINKED ISSUE #2335: Rolling session race condition
### Checklist

### Description

On my site with dozens of components I noticed that _sometimes_ when a user tries to logout, the user is redirected through the logout process, but finally is still logged in. This seemed to happen mostly when trying to logout while parts of the site are still loading,

I analyzed this in the browser network tab and noticed that the reason are requests that are still in-flight while the user presses the logout button.
While the `/auth/logout` request correctly clears the session cookies, the in-flight requests sometimes get responses after `/auth/logout` and set the session back to active.

Timeline:
Image

A walkaround for me was to use the auth0/nextjs-auth0 middleware only on `/auth` paths, contrary to what is recommended in the documentation:
```
  if (path.startsWith("/auth/")) { // added this condition
    const authResponse = await auth0.middleware(request);
    // other auth code
  }
  // other site specific middleware
```

Obviously, this prevents rolling sessions from working.

### Reproduction

Logging out while there are ongoing requests.

### Additional context

_No response_

### nextjs-auth0 version

4.9.0

### Next.js version

15.5.2

### Node.js version

20.19.2

sleitor:
## Problem

Fixes #2335.

When a user logs out while concurrent in-flight requests are being processed, the middleware was writing a fresh `Set-Cookie` header to roll the session in *every* response. If any of those responses arrived at the browser **after** the logout response the session cookie was silently restored, leaving the user apparently still authenticated.

## Root Causes

Two separate issues contribute to the bug:

### 1. StatefulSessionStore – no existence check before rolling write

`StatefulSessionStore.set()` called `store.set(sessionId, session)` unconditionally. If the session had already been deleted by a concurrent logout (via `store.delete()`) the write would re-insert it, reviving the session.

### 2. Middleware – rolling update fired even when `rolling: false`

The `else` branch in `AuthClient.handler()` that touches sessions on every non-auth request was reading and re-writing the session even when the session store was configured with `rolling: false`. This is the TODO that has existed in that branch since the feature was written.

### Fix 1 – `StatefulSessionStore.set()`: guard against reviving a deleted session

Before persisting a rolling-session update the store is now queried with the session ID found in the incoming request cookie. If the session no longer exists (a concurrent logout already called `store.delete()`) the write is skipped and no `Set-Cookie` header is emitted. The browser never receives a revived session cookie.

The guard only applies to rolling updates (sessions with an existing cookie, `isNew = false`). Brand-new sessions (fresh login, first visit) bypass the check so those flows remain unaffected.

```ts
// src/server/session/stateful-session-store.ts
if (existingSessionId !== null) {
  const existingSession = await this.store.get(existingSessionId);
  if (!existingSession) {
    return; // session was deleted (logout) – do not re-create it
  }
}
```

### Fix 2 – Middleware: honour `rolling: false`

A new `isRolling` getter is exposed on `AbstractSessionStore`. The middleware now gates the session-touch block behind `this.sessionStore.isRolling`, eliminating unnecessary reads from the session store (and the associated `Set-Cookie` responses) when rolling sessions are disabled.

```ts
// src/server/auth-client.ts
if (this.sessionStore.isRolling) {
  const session = await this.sessionStore.get(req.cookies);
  if (session) {
    await this.sessionStore.set(req.cookies, res.cookies, { ...session });
    addCacheControlHeadersForSession(res);
  }
}
```

## Limitations

The stateless (cookie-only) session store cannot be fixed with a pure server-side guard because all session state lives in the encrypted cookie itself — there is no server-side record to check for revocation. For stateless sessions the recommended mitigation remains to scope the middleware matcher to `/auth/*` routes only, or to migrate to a stateful session store.

## Tests

- **`stateful-session-store.test.ts`** – three new cases under *rolling session race condition*:
  - A session deleted in the store (simulating a concurrent logout) is **not** re-created.
  - A session that still exists in the store **is** rolled normally.
  - A new-login session (`isNew = true`) bypasses the guard (session fixation prevention takes priority).

- **`auth-client.test.ts`** – one new case verifying that no `Set-Cookie` header is emitted when `rolling: false`.


sleitor:
Rebased on latest `main` (through `Fix/gh 2440` / v4.16.0) — the conflict in `with-page-auth-required.ts` was from an unrelated PR that got merged upstream. Our rolling-session fix is unchanged; only the rebase was needed to clear the merge conflict.

Piyush-85:
Thanks for this contribution @sleitor.
The fix works perfectly vast majority of real-world cases since logout typically completes well before an in-flight request finishes.
The `get + set` guard reduces the race window but doesn't close it. Logout can still land between the two calls:
```
store.get("ses_abc")   → exists ✓
  ← logout deletes the row here
store.set("ses_abc")   → upsert re-creates it  
```

For a 100% fix, adding an optional `update?()` to `SessionDataStore` that does a conditional write — `UPDATE WHERE id = $1`, not an upsert. 
The check and the write become one atomic DB operation so there is no gap for logout to slip through. Stores that don't implement it fall back to this `get + set` behaviour unchanged, fully backward-compatible.

- [PR for reference ](https://github.com/auth0/nextjs-auth0/pull/2590)

Piyush-85 (approved):

PR #2545, opened 2026-03-06
feat: MCD (Multiple Custom Domains) Support

tusharpandey13:
This change adds support for [Multiple Custom Domains](https://auth0.com/docs/customize/custom-domains/multiple-custom-domains) using a dynamic `DomainResolver` in client configuration.

- `Auth0ClientOptions.domain` widened to `string | DomainResolver` where `DomainResolver` is a callback that returns a domain string.
- `AuthClientProvider` replaces singleton `AuthClient` with shared discovery/JWKS caches, `AuthClient` made domain-aware.
- Domain/issuer session storage and mismatch check logic.
- Callback pins issuer in the transaction cookie.
- Domain normalization logic.
- `openid` scope enforced in resolver mode for `iss` claim availability.
- Logout, backchannel, MFA, and proxy are domain-aware via provider delegation.

## Usage

```ts
  domain: async ({ headers, url }) => {
    // your code here
    const domain = db.getDomain() || "auth.brand1.com";
    return domain;
  },
});
```

Full docs: [EXAMPLES.md — Multiple Custom Domains](https://github.com/auth0/nextjs-auth0/blob/94eda6139f89c54c361f2a2af9f69de503ccebd1/EXAMPLES.md#multiple-custom-domains-mcd)

## Tests

277 new tests (flow tests and UTs) across 7 files covering:
- Domain normalization
- Error handling
- Shared discovery cache
- AuthClient factory
- Session domain gating
- Auth flows

## References
- [Multiple Custom Domains - Auth0 Docs](https://auth0.com/docs/customize/custom-domains/multiple-custom-domains)


tusharpandey13:
## Review guide
Changes can be categorized as:
### New MCD logic
These changes add the MCD capability to the SDK, including the new AuthClient factory, caches, types and errors.
- `auth-client-provider.ts`
- `discovery-cache.ts`
- `types/mcd.ts` & `errors/mcd.ts`
- tests for these files
### Core integration
These changes plug MCD into existing logic
- `auth-client.ts`
- `client.ts`
- `transaction-store.ts`
- `server-mfa-client.ts`
- tests for these & exports 
### Updates to existing test files
Tests for:
`fetcher`, `get-access-token`, `with-api-auth-required`, `with-page-auth-required`, `mfa-error-bubbling`, `mfa-server`, `proxy-handler`
### Docs
`EXAMPLES.md`, `README.md`

src/server/auth-client.ts
```diff
@@ -543,17 +545,21 @@ export class AuthClient {
...
-      if (session) {
-        // we pass the existing session (containing an `createdAt` timestamp) to the set method
-        // which will update the cookie's `maxAge` property based on the `createdAt` time
-        await this.sessionStore.set(req.cookies, res.cookies, {
-          ...session
-        });
-        addCacheControlHeadersForSession(res);
+      // Skip touch silently if domain mismatch or no session
+      if (error || !session) {
+        return res;
```
nandan-bhat: Should we log a warning for the `Domain Mismatch` case ?
Otherwise, the developers wouldn't know the difference between not having session vs domain mismatch.

src/server/discovery-cache.ts
```diff
@@ -0,0 +1,195 @@
...
+
+  /**
+   * Maximum number of domain entries to cache. Default: 100
+   */
+  maxEntries?: number;
+
+  /**
+   * Maximum number of JWKS entries in the co-located cache. Default: 10
+   */
+  maxJwksEntries?: number;
```
nandan-bhat: Since each domain has exactly one `JWKS URI`, both caches will always grow at the same rate one entry per domain. Having a separate  `maxJwksEntries` option doesn't seem necessary.

Would it be simpler to just use `maxEntries` for both caches and remove `maxJwksEntries` entirely ?

Or am I missing any specific use-case here ?

src/server/discovery-cache.ts
```diff
@@ -0,0 +1,195 @@
...
+   */
+  ttl?: number;
+
+  /**
+   * Maximum number of domain entries to cache. Default: 100
+   */
+  maxEntries?: number;
+
+  /**
+   * Maximum number of JWKS entries in the co-located cache. Default: 10
```
nandan-bhat: `Default: 10` looks like a mistake. Can you confirm ?
If the default is intentionally `10`, what is the reason to fix on this number ?

src/types/mcd.ts
```diff
@@ -0,0 +1,121 @@
...
+  url?: URL;
+}) => Promise<string> | string;
+
+/**
+ * Configuration for the OIDC discovery metadata cache.
+ * Applies in both static and resolver modes.
+ *
+ * @property ttl - Time-to-live for cached discovery metadata in seconds. Default: 600 (10 minutes).
+ * @property maxEntries - Maximum number of cached issuers. Default: 100. LRU eviction.
+ * @property maxJwksEntries - Maximum number of cached JWKS entries. Default: 10. Independent LRU eviction.
```
nandan-bhat: Can you confirm that `Default: 10` is intentional ?

src/server/auth-client.ts
```diff
@@ -1908,10 +2056,83 @@ export class AuthClient {
...
+        error: null,
+        session,
+        exists: true
+      };
+    }
+
+    // Pre-MCD session (no mcd field): need to backfill
+    // Fail-open in both modes: backfill with current domain instead of throwing error
+    session.internal = session.internal || {};
+    session.internal.mcd = {
```
nandan-bhat, line 2187: If we backfill the issuer / domain like this, if app cookies are shared across subdomains, wouldn’t that let the first MCD domain that sees an old session claim it ?

src/server/auth-client.ts
```diff
@@ -986,6 +1057,21 @@ export class AuthClient {
     }
 
     const idTokenClaims = oauth.getValidatedIdTokenClaims(oidcRes)!;
+
+    // Secondary issuer check for defense-in-depth
+    if (transactionState.originIssuer) {
```
nandan-bhat, line 1079: This `if` block can be removed I guess. Can you confirm if this defensive check is really required ?

Since this SDK only deals with `Auth0` domains, `this.domain` would most likely match with the `issuer` (normalized) returned by `discoverAuthorizationServerMetadata`.

So, `idTokenClaims` would contain the `iss` claim same as normalized `this.domain`.

src/server/client.ts
```diff
@@ -767,8 +826,10 @@ export class Auth0Client {
     token_type?: string;
     audience?: string;
   }> {
-    const session: SessionData | null = req
-      ? await this.getSession(req)
+    const { authClient, normalizedReq } = await this.resolveRequestContext(req);
```
nandan-bhat, line 857: When `getAccessToken` is called,
Do we really need to resolve the `dynamic resolver` function twice ?

In `getAccessToken` flow:
`getAccessToken` > `executeGetAccessToken` (calls `resolveRequestContext`) > `getSession` > calling `resolveRequestContext` again.


I know that `getSession` can be called independently, but can we check if it's possible to avoid multiple calls ?

src/server/auth-client.ts
nandan-bhat: Should we use `discoveryCache.getJwksCacheForUri(jwks_uri)` here instead of `this.jwksCache` ?

EXAMPLES.md
```diff
@@ -3628,3 +3645,464 @@ The SDK provides typed error classes for all MFA operations:
...
+2. Explicitly compares the ID token `iss` claim against the stored `originIssuer`
+
+This prevents cross-domain token confusion where an attacker redirects the callback to a different domain.
+
+### Backward Compatibility
+
+MCD is fully backward compatible:
+
+- **Existing apps**: No changes required. `domain: "string"` and `AUTH0_DOMAIN` env var work as before
+- **Pre-MCD sessions**: Sessions without `internal.mcd` metadata continue to work in static mode. In resolver mode, pre-MCD sessions are rejected (fail-closed) to prevent domain confusion
```
nandan-bhat, line 4119: What will happen if the resolver resolves to the same `domain` as pre-MCD static `domain` ?

If we `fail`, that  would be a breaking change.

src/errors/mcd.ts
```diff
@@ -0,0 +1,135 @@
...
+
+/**
+ * Error thrown when backchannel logout fails in an MCD context.
+ *
+ * Renamed from `BackchannelLogoutError` to avoid collision with the generic
+ * `BackchannelLogoutError` in `oauth-errors.ts`.
+ *
+ * @public
+ */
+export class McdBackchannelLogoutError extends SdkError {
```
nandan-bhat: Where are we using this error ?

src/server/auth-client-provider.ts
```diff
@@ -0,0 +1,296 @@
...
+  }
+
+  /**
+   * Gets the shared discovery cache used across all domains.
+   *
+   * @returns The DiscoveryCache instance
+   *
+   * @internal
+   */
+  getDiscoveryCache(): DiscoveryCache {
```
nandan-bhat: Where are we using `getDiscoveryCache()` ?
I could see it referenced only in the tests.

src/server/auth-client-provider.ts
```diff
@@ -0,0 +1,296 @@
...
+   * Creates a new AuthClientProvider instance.
+   *
+   * @param options - Configuration options
+   * @throws {InvalidConfigurationError} If configuration is invalid (neither domain nor resolver)
+   *
+   * @internal
+   */
+  constructor(options: AuthClientProviderOptions) {
+    this.createAuthClientFactory = options.createAuthClient;
+    this.discoveryCache = new DiscoveryCache(options.discoveryCacheOptions);
```
nandan-bhat: If `getDiscoveryCache` is un-used, we don't need `this.discoveryCache`. I saw `Auth0Client` already creates a separate `DiscoveryCache` and passes that into each `AuthClient`

src/server/auth-client-provider.ts
```diff
@@ -0,0 +1,296 @@
...
+   *
+   * @internal
+   */
+  forDomainSync(domain: string): AuthClient {
+    // Normalize and validate the domain
+    const normalized = normalizeDomain(domain);
+    const normalizedDomain = normalized.domain;
+
+    // Check cache first
+    let client = this.domainClients.get(normalizedDomain);
```
nandan-bhat: The comments say `LRU`, but these caches do not update recency on cache hits. 

Here (in `forDomainSync`) a cache hit returns directly, and eviction removes the first inserted key irrespective of whether it's recently used or not.

src/server/auth-client-provider.ts
```diff
@@ -0,0 +1,296 @@
...
+   * @internal
+   */
+  async getProxyFetcher(
+    key: string,
+    factory: () => Promise<any>
+  ): Promise<any> {
+    let fetcher = this.proxyFetchers.get(key);
+    if (!fetcher) {
+      fetcher = await factory();
+      this.proxyFetchers.set(key, fetcher);
```
nandan-bhat: Same as https://github.com/auth0/nextjs-auth0/pull/2545/changes#r2946759603

src/server/auth-client.ts
tusharpandey13: Yes, adding this change 👍

src/server/discovery-cache.ts
tusharpandey13: Technically 2 custom doamins can have the same `jwks_uri` so there can be a difference in size of the 2 caches.
But this suggestion is good and removes complexity, adding 👍

src/server/discovery-cache.ts
tusharpandey13: This is a remnant of older design where it was capped to 10, changing 👍

src/server/auth-client.ts
tusharpandey13: Yes, this is true.

Instead of having `Fail-open in both modes`, we can just do backfill in `static` mode (customer code has MCD backfill logic but MCD is OFF (no resolver configured)) and when backfill is done, customer can configure resolver and this will not be needed.

Another solution can be to infer domain from session's iss claim in resolver mode. This can potentially fail if domain in `iss` is not part of possible domains returned by `Resolver`

What do you think on this? I think 1st one is a simpler approach but option 2 provides better UX (with an edge case that can be handled)

src/server/auth-client.ts
tusharpandey13: This was kept for the following cases:
- Transaction cookie manipulation (attacker changes which domain processes the callback)
- Race conditions where the resolver returns a different domain for the callback than for login initiation (bounding txn cookie to an issuer)

src/server/client.ts
tusharpandey13: Will fix

src/server/auth-client.ts
tusharpandey13: Will fix 👍

EXAMPLES.md
tusharpandey13: With [this change (option 1)](https://github.com/auth0/nextjs-auth0/pull/2545#discussion_r2954802000), this will be a documented 2-phase migration flow.

The flow becomes:
> If you're switching from static to resolver mode (even with the same domain),
  deploy first with the static string to backfill sessions, then switch to the
  resolver. Zero re-auth required.

With option 2, this is solved and no longer a concern.

src/errors/mcd.ts
tusharpandey13: Unused, will remove

src/server/auth-client-provider.ts
tusharpandey13: Will remove

src/server/auth-client-provider.ts
tusharpandey13: Will remove

src/server/auth-client-provider.ts
tusharpandey13: This is true, will add delete and re-set on hit logic to make this an LRU

src/server/auth-client-provider.ts
tusharpandey13: Same as https://github.com/auth0/nextjs-auth0/pull/2545#discussion_r2954926937

src/types/mcd.ts
tusharpandey13: This is a remnant of older design where it was capped to 10, changing 👍

src/server/auth-client.ts
tusharpandey13: Going with approach 2.

tusharpandey13:
@nandan-bhat review comments have been addressed. Unresolved conversations contain some explanations for reference.

EXAMPLES.md
tusharpandey13: Option 2 has been implemented, no breaking change.

nandan-bhat (approved):

PR #2552, opened 2026-03-12
fix: delete v3 appSession cookie on logout

gmurphey:
## Summary
- When upgrading from v3 to v4, the legacy `appSession` cookie was only cleaned up during login (`set()`), not during logout (`delete()`). This caused users with a lingering v3 cookie to bypass authentication on their next login attempt.
- Adds `appSession` cookie deletion to `StatelessSessionStore.delete()` and `StatefulSessionStore.delete()`, mirroring the existing cleanup logic in their `set()` methods.

## Test plan
- [x] Added test: stateless `delete()` removes legacy chunked `appSession` cookie when present
- [x] Added test: stateful `delete()` removes legacy `appSession` cookie when present
- [x] Added test: stateful `delete()` skips redundant deletion when session cookie name is already `appSession`
- [x] All 891 existing tests continue to pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Piyush-85:
@gmurphey Thanks for the contribution. This fix was critical and LGTM.
Approving and merging this.

**Note** - Will add a section **Session Cookie Migration** in the `V4 Migration Guide.md` to reflect this change.

Piyush-85 (approved):

PR #2555, opened 2026-03-13
fix: MFA http contract parity fixes

tusharpandey13:
Align MFA route handlers with Auth0 HTTP API contract. 
Route handlers become transparent proxies: accept snake_case requests, return snake_case responses.
Camelization moves to SDK client boundary (`ClientMfaClient`, `ServerMfaClient`).

## Changes

**Wire Format**
- Token extraction: `Authorization: Bearer` header only (query param removed)
- Camelization logic moved from business layer -> mfa client. This enables Auth0-like HTTP calls to and from the SDK's MFA proxy layer.
- Renamed `/auth/mfa/enroll` → `/auth/mfa/associate` (matches Auth0)

**`factorType`**: Added support for `factorType` instead of raw `authenticatorTypes`/`oobChannels`

## Tests

888 pass, 0 fail (50 files)


gyaneshgouraw-okta:
@tusharpandey13 in the PR description we should highlight breaking changes for customers under a breaking change section.
That would help community members visiting this PR.

gyaneshgouraw-okta (dismissed):

src/client/mfa/index.ts
gyaneshgouraw-okta: @tusharpandey13 since  ```/auth/mfa/enroll  is updated to /auth/mfa/associate``` should this `url.includes("/enroll")` needs to be updated with `url.includes("/associate")` 

Is the route update applicable here also ?

src/utils/mfa-transform-utils.ts
nandan-bhat: Should we return an error if `authenticatorType` is not `otp` or `oob` here ? Right now this looks like it will treat any other value as `OTP`.

src/utils/mfa-transform-utils.ts
```diff
@@ -201,3 +222,76 @@ export const getVerifyGrantType = (params: URLSearchParams) => {
...
+ * });
+ * // Returns unchanged
+ * ```
+ */
+export function normalizeEnrollOptions(
+  options: EnrollOptions
+): EnrollOobOptions | EnrollOtpOptions {
+  if ("factorType" in options) {
+    const mapping = FACTOR_MAPPING[options.factorType];
+    if (!mapping) throw new Error(`Unknown factorType: ${options.factorType}`);
```
nandan-bhat: Can we throw an `SDK` error here instead of a plain `Error` ?
It looks like client code and server code may get different error types for the same bad input.

src/utils/mfa-transform-utils.ts
```diff
@@ -201,3 +222,76 @@ export const getVerifyGrantType = (params: URLSearchParams) => {
...
+ *   mfaToken: 'token123',
+ *   authenticatorTypes: ['otp']
+ * });
+ * // Returns unchanged
+ * ```
+ */
+export function normalizeEnrollOptions(
+  options: EnrollOptions
+): EnrollOobOptions | EnrollOtpOptions {
+  if ("factorType" in options) {
```
nandan-bhat, line 281: Should we check here that `phoneNumber` is present when `factorType` is `sms` or `voice` ?
The type comment says it is required, but this still allows it to be missing: https://github.com/auth0/nextjs-auth0/blob/81bea1859a9985cad0e6671fcf3fe7bee861cb33/src/types/mfa.ts#L164

tusharpandey13: Done.

src/utils/mfa-transform-utils.ts
tusharpandey13: Done

src/utils/mfa-transform-utils.ts
tusharpandey13: Done

nandan-bhat:
LGTM

nandan-bhat (approved):

PR #2580, opened 2026-03-31
fix: DPoP nonce retry race issue

nandan-bhat:
### Background:
The proxy code reused shared fetcher state for requests to the same `audience`. As part of that reuse, it replaced the fetcher’s token lookup function on every request. When `DPoP nonce` retry happened, `request A` could pause, `request B` could come in and replace that function, and then `request A’s` retry could continue using `request B’s` token.

### Fix:
This PR stops sharing request-specific token lookup state across requests. It now only reuses the `DPoP` handle per `audience`, and creates a fresh fetcher for each request with that request’s own token lookup function. That keeps `DPoP nonce` state reusable, while making sure retries stay bound to the correct session.


src/server/auth-client.ts
```diff
@@ -3508,38 +3509,24 @@ export class AuthClient {
...
-    // Get/create fetcher instance with domain-aware caching
-    // In MCD mode, use provider.getProxyFetcher for shared caching across AuthClient instances
-    // In static mode, use local proxyFetchers cache
-    const cacheKey = this.provider
-      ? `${this.domain}:${options.audience}`
-      : options.audience;
-
-    let fetcher: Fetcher<Response>;
-    if (this.provider) {
-      fetcher = await this.provider.getProxyFetcher(cacheKey, async () => {
```
Piyush-85, line 3520: Removing getProxyFetcher from this file makes `getProxyFetcher` a dead code, that can be removed.

src/server/auth-client.ts
```diff
@@ -2951,9 +2951,10 @@ export class AuthClient {
     const fetcherConfig: FetcherConfig<TOutput> = {
       // Fetcher-scoped DPoP handle and nonce management
       dpopHandle:
-        this.useDPoP && (options.useDPoP ?? true)
+        options.dpopHandle ??
```
Piyush-85: Not sure about this implementation even if useDPoP is false and user passes some dpopHandle this will pass.

src/server/auth-client.ts
tusharpandey13:
```
auth-client-provider.ts:42: const MAX_PROXY_FETCHERS = 100
auth-client-provider.ts:64: private proxyFetchers: LruMap<string, any>
auth-client-provider.ts:79: this.proxyFetchers = new LruMap(MAX_PROXY_FETCHERS
auth-client-provider.ts:207-220: getProxyFetcher() method
auth-client-provider.test.ts:519-600: Tests for getProxyFetcher  
mcd.integration.test.ts:750-780: MCD integration tests for proxy fetcher caching  
```
These are dead code after this change

nandan-bhat: Addressed in 372d576ae210d9c5f2003bb7d6cf61f9caa4c868

src/server/auth-client.ts
nandan-bhat: Addressed in 372d576ae210d9c5f2003bb7d6cf61f9caa4c868

tusharpandey13:
Was dropping support for MCD cross-instance DPoP handle sharing intentional?
This change moves the cache from provider to internal client.
Before: `AuthClientProvider` (`this.provider.getProxyFetcher(cacheKey)`) 
New: `AuthClient` (`this.proxyDpopHandles[audience]`).

nandan-bhat:
> Was dropping support for MCD cross-instance DPoP handle sharing intentional? This change moves the cache from provider to internal client. Before: `AuthClientProvider` (`this.provider.getProxyFetcher(cacheKey)`) New: `AuthClient` (`this.proxyDpopHandles[audience]`).

This is intentional. In practice we still share the `DPoP` handle for the same `domain`, because the provider already reuses a single `AuthClient` per `domain`. So moving the cache from the provider to `AuthClient` does not lose the useful `MCD` sharing, but it does avoid sharing request-specific fetcher state which was the root cause of the issue.

tusharpandey13 (approved):

Piyush-85 (approved):

PR #2582, opened 2026-04-01
fix: stale session reuse on proxy DPoP nonce retry

nandan-bhat:
Fix a proxy-only `DPoP` retry bug where a nonce challenge immediately after a token refresh could trigger a second refresh using the stale pre-refresh session snapshot.

## Problem

`#handleProxy()` creates a per-request `getAccessToken()` closure over the session loaded at the start of the request.

When that proxied request does all of the following in one flow:

1. starts with an expired access token,
2. refreshes successfully,
3. sends the refreshed access token upstream,
4. receives a `use_dpop_nonce` challenge from the protected resource,

the fetcher retries the protected-resource request by calling `getAccessToken()` again.

Before this change, the second `getAccessToken()` call still read from the original session snapshot loaded at the start of the proxy request. The refreshed token set was only persisted after `fetchWithAuth()` completed, so the retry path could call `getTokenSet()` again and perform a second refresh with the old refresh token.

With rotating refresh tokens enabled, that second refresh uses a token that was already invalidated by the first refresh, causing the proxy request to fail with `invalid_grant` and return a `500`.

## Root Cause

The proxy retry path reused the original session snapshot instead of the refreshed token state produced earlier in the same request.

The `DPoP` nonce retry itself was valid, the issue was that the retry reacquired credentials from stale request-local session state rather than reusing the first refreshed token set.

## Fix

This PR makes proxy token retrieval request-local and retry-safe:

- `#handleProxy()` now keeps a request-local session snapshot for the lifetime of the proxy request.
- After a successful `getTokenSet()` call, that request-local snapshot is updated immediately so subsequent retries in the same request see the refreshed token state.
- Token responses are memoized per proxy request for the same access-token request parameters, so a `DPoP` nonce retry reuses the first token response instead of performing another refresh.
- Session persistence behavior is unchanged: the final session is still written once after the proxy request completes successfully.
- The MFA-required path now uses the request-local mutated snapshot so any in-request state changes are preserved there as well.

### DIFF
--- a/src/server/auth-client.ts
+++ b/src/server/auth-client.ts
@@ -3718,10 +3718,48 @@ export class AuthClient {
       ? await clonedReq.arrayBuffer()
       : undefined;
 
+    // Keep an isolated, request-local session snapshot so retries within this
+    // proxy request can observe token refreshes immediately without writing the
+    // session cookie until the proxy response succeeds.
+    const requestSession = structuredClone(session);
+    const tokenSetResponsesByRequest = new Map<string, GetTokenSetResponse>();
+    const applyTokenSetResponseToRequestSession = (
+      tokenSetResponse: GetTokenSetResponse
+    ) => {
+      const sessionChanges = getSessionChangesAfterGetAccessToken(
+        requestSession,
+        tokenSetResponse.tokenSet,
+        {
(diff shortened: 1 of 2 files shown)


Piyush-85 (approved):
LGTM

Piyush-85 (approved):

PR #2590, opened 2026-04-07
feat: add optional update() to SessionDataStore

### LINKED ISSUE #2335: Rolling session race condition

Piyush-85:
References [PR#2530](https://github.com/auth0/nextjs-auth0/pull/2530)
Fixes https://github.com/auth0/nextjs-auth0/issues/2335.

### Background

In stateful session mode, the browser cookie holds only an encrypted session ID. All session data lives in the external store (Redis, Postgres, etc). On every rolling-session request, `StatefulSessionStore.set()` is called to persist the refreshed session back to the store.
Prior to this change, `set()` performed an unconditional **upsert** — it would write the session regardless of whether it still existed in the store. This opened a resurrection bug:

```
T1  User calls /logout  →  store.delete(sessionId)
T2  In-flight request   →  store.set(sessionId, session)   ← session recreated
T3  User is still "logged in" despite having explicitly logged out
```

### Previous fix and its limitation (TOCTOU gap)

A prior [PR](https://github.com/auth0/nextjs-auth0/pull/2530) introduced a guard to `set()` that reads the session before writing:

```ts
const existing = await store.get(existingSessionId);  // T1: exists
// ← logout can land HERE, deleting the row
await store.set(sessionId, session);                  // T2: re-creates it
```

This is a classic **TOCTOU (Time-Of-Check-Time-Of-Use)** race. The `get` and `set` are two separate async DB round-trips with a real window between them. A concurrent logout that lands in that gap still resurrects the session — the guard catches the obvious case but not the race.

---

### This change: optional `update()` on `SessionDataStore`

A new optional method is added to the `SessionDataStore` interface:

```ts
update?(id: string, session: SessionData): Promise<boolean>;
```

**Semantics:** behaves like `UPDATE … WHERE id = $1` — write only if the row exists, return `true` if updated, `false` if not found. It must **not** insert a new row (no upsert).

When implemented, the SDK uses it instead of the `get + set` pair for rolling-session writes. Because the check and write happen in a **single atomic DB operation**, there is no window for a concurrent logout to slip through:

```
store.update(id, session)
  → row exists → updated in one op → true   (cookie refreshed)
  → row deleted by logout → false            (no cookie, no resurrection)
```

For stores that do not implement `update()`, the SDK falls back to the existing non-atomic `get + set` guard, which is still a significant improvement over the original unconditional upsert.

### Backwards compatibility

- Fully backwards compatible. `update()` is optional — existing store implementations require no changes. 

- The `get + set` fallback path is unchanged.


src/server/session/stateful-session-store.ts
```diff
@@ -157,11 +157,25 @@ export class StatefulSessionStore extends AbstractSessionStore {
...
+        const updated = await this.store.update(existingSessionId, session);
+        if (!updated) {
+          return;
+        }
+      } else {
+        const existingSession = await this.store.get(existingSessionId);
+        if (!existingSession) {
+          return;
+        }
+        await this.store.set(sessionId, session);
```
sleitor: **Bug: `store.set()` called with `sessionId` instead of `existingSessionId` in the fallback path**

In the `else` branch (stores without `update()`), the call is:

```ts
await this.store.set(sessionId, session);
```

But `sessionId` may have been regenerated earlier in the method (when `isNew` is true and the old session is deleted). In the TOCTOU-fallback path we already confirmed the session exists under `existingSessionId` via `get()` — writing to a potentially different key creates an orphaned entry.

This should be:

```ts
await this.store.set(existingSessionId, session);
```

The `update()` path on line 13 correctly uses `existingSessionId` — the fallback should match.

src/server/session/stateful-session-store.ts
```diff
@@ -157,11 +157,25 @@ export class StatefulSessionStore extends AbstractSessionStore {
...
+    //
+    // If the store implements the optional update() method we use it as an atomic
+    // check-and-write (UPDATE WHERE id = $1). Otherwise we fall back to a non-atomic
+    // get() + set() pair.
     if (existingSessionId !== null) {
-      const existingSession = await this.store.get(existingSessionId);
-      if (!existingSession) {
-        return;
+      if (typeof this.store.update === "function") {
+        const updated = await this.store.update(existingSessionId, session);
```
sleitor, line 166: **Correct: `update()` uses `existingSessionId` ✅**

Just noting for contrast — this correctly targets the ID that was originally in the cookie. The fallback `set()` a few lines below uses `sessionId` instead, which can diverge after session regeneration.

src/types/index.ts
```diff
@@ -46,6 +46,12 @@ export interface SessionDataStore {
    */
   set(id: string, session: SessionData): Promise<void>;
 
+  /**
+   * Optional: update the session by its ID only if it already exists.
+   * Return `true` if updated, `false` if not found.
+   */
+  update?(id: string, session: SessionData): Promise<boolean>;
+
```
sleitor, line 63: **Consider documenting TTL refresh expectation**

The `set()` method doesn't accept a TTL either (stores own their expiry logic), so this signature is consistent. However it's worth explicitly documenting that `update()` implementations should also refresh the session TTL — otherwise a successful `update()` returning `true` could still serve an expired session on the next `get()`.

Suggested JSDoc addition:

```ts
/**
 * Optional: update the session by its ID only if it already exists.
 * Return `true` if updated, `false` if not found.
 *
 * Implementations should also reset the session TTL/expiry on a successful
 * update, matching the behaviour of `set()`.
 */
update?(id: string, session: SessionData): Promise<boolean>;
```

EXAMPLES.md
```diff
@@ -2634,6 +2634,12 @@ export const auth0 = new Auth0Client({
     async delete(id) {
       // delete the session using its ID
     },
+
+    async update(id, sessionData) {
```
sleitor, line 2655: **Nit: empty example body may confuse implementors**

The other methods in this block have inline comments showing intent, but `update()` only has a description. The critical requirement here is **atomicity** — a `get` + conditional `set` inside `update()` would recreate the exact TOCTOU gap this PR is trying to close. Worth calling that out:

```ts
async update(id, sessionData) {
  // MUST be a single atomic operation — do not use get() + set() here.
  // Example (PostgreSQL):
  //   const r = await db.query(
  //     'UPDATE sessions SET data=$2, expires_at=NOW()+interval\'1 day\' WHERE id=$1',
  //     [id, sessionData]
  //   );
  //   return r.rowCount > 0;
  //
  // Example (Redis — Lua script for atomicity):
  //   const lua = `if redis.call('exists',KEYS[1])==1 then
  //     redis.call('set',KEYS[1],ARGV[1],'EX',ARGV[2]) return 1
  //     else return 0 end`;
  //   return await redis.eval(lua, 1, id, JSON.stringify(sessionData), ttl) === 1;
},
```

sleitor (commented):
Good follow-up to #2530 — the atomic `update()` approach cleanly closes the TOCTOU gap for stores that can support it, and the optional fallback preserves full backward compatibility.

Found one real bug in the fallback path (`store.set()` targeting the wrong session ID), plus a few smaller observations around the type signature, docs atomicity warning, and test coverage. Details inline.

src/server/session/stateful-session-store.test.ts
```diff
@@ -463,6 +463,101 @@ describe("Stateful Session Store", async () => {
     });
 
     describe("rolling session race condition", async () => {
+      it("should use store.update() atomically when implemented and bail if it returns false", async () => {
```
sleitor, line 466: **Missing test: new session with `update()` implemented**

The refactor moved `store.set()` into a conditional `else` block. The existing tests above this `describe` cover new sessions without `update()`, but there's no test asserting that when a store *does* implement `update()`, a brand-new session (no request cookie → `existingSessionId === null`) still calls `store.set()` and NOT `store.update()`.

Suggested test:

```ts
it("should call store.set() and not store.update() for a brand-new session even when update() is implemented", async () => {
  const secret = await generateSecret(32);
  const session: SessionData = {
    user: { sub: "user_123" },
    tokenSet: { accessToken: "at", refreshToken: "rt", expiresAt: 123456 },
    internal: { sid: "sid", createdAt: Math.floor(Date.now() / 1000) }
  };
  const store = {
    get: vi.fn(),
    set: vi.fn(),
    delete: vi.fn(),
    update: vi.fn(), // must NOT be called for new sessions
  };
  const requestCookies = new RequestCookies(new Headers()); // no existing cookie
  const responseCookies = new ResponseCookies(new Headers());

  const sessionStore = new StatefulSessionStore({ secret, store, rolling: true });
  await sessionStore.set(requestCookies, responseCookies, session);

  expect(store.update).not.toHaveBeenCalled();
  expect(store.set).toHaveBeenCalledOnce();
  expect(responseCookies.get("__session")).toBeDefined();
});
```

Piyush-85:
@sleitor Thanks for the thorough review. 
All above comments addressed, please take another look and resolve if everything looks good!

sleitor (approved):
All review comments addressed — thank you!

- The `existingSessionId` bug in the fallback `set()` is fixed ✅
- TTL contract documented at the interface level (nice touch — applies to both `set` and `update`) ✅
- `EXAMPLES.md` now warns about the atomicity requirement with concrete DB examples ✅
- New test for brand-new session with `update()` implemented ✅

One tiny nit (not blocking): the JSDoc for `update()` has a trailing blank line before `*/` — cosmetic only.

LGTM 🚀

tusharpandey13 (approved):

PR #2602, opened 2026-04-15
fix(server): defer InvalidConfigurationError when neither domain nor resolver is configured

### LINKED ISSUE #2601: Auth0 Domain is now required during build (v4.17.0)
### Checklist

### Description

Prior to this change https://github.com/auth0/nextjs-auth0/pull/2545 the AUTH0_DOMAIN was only required at runtime.  It should only be required at runtime since it is not prefixed with NEXT_PUBLIC.  After this change, the domain is required at build time or the following error is thrown:

`Error [InvalidConfigurationError]: You must provide either a domain string or a DomainResolver function.`

### Reproduction

1. Create a new Next.js application with output set to `standalone`
2. Following the setup instructions to add Auth0
3. Build the application without the AUTH0_DOMAIN env variable

### Additional context

_No response_

### nextjs-auth0 version

4.17.0

### Next.js version

16.2.3

### Node.js version

24.14.0

SAY-5:
Per #2601, since v4.17.0 (#2545) the `AuthClientProvider` constructor throws `You must provide either a domain string or a DomainResolver function` whenever `AUTH0_DOMAIN` is missing from the environment. Next.js builds (especially `output: 'standalone'`) instantiate the provider during static analysis even when the domain is intentionally only available at runtime, which now hard-fails the build.

This PR defers the error: stores it as `deferredConfigError` and re-throws the exact same `InvalidConfigurationError` the first time the provider is actually used at runtime (`forRequest` / `getAuthClientForStaticMode`). Behaviour is unchanged for any configuration that *was* valid before — the missing-config case now fails at runtime instead of build time, matching the pre-4.17.0 contract.

Closes #2601

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -2651,6 +2651,28 @@ export const auth0 = new Auth0Client({
     async delete(id) {
       // delete the session using its ID
     },
+
+    async update(id, sessionData) {
+      // Optional: update the session by its ID only if it already exists; return true if updated, false if not found.
+      // Prevents a session deleted by logout from being re-created by a concurrent in-flight rolling write.
+      //
+      // MUST be a single atomic operation — do not use get() + set() here,
+      // as that would recreate the TOCTOU race condition this method is designed to prevent.
+      //
+      // Example (PostgreSQL):
+      //   const r = await db.query(
+      //     "UPDATE sessions SET data=$2, expires_at=NOW()+interval'1 day' WHERE id=$1",
+      //     [id, sessionData]
+      //   );
+      //   return r.rowCount > 0;
+      //
+      // Example (Redis — Lua script for atomicity):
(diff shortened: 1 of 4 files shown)


tusharpandey13:
Hi @SAY-5 , thanks for this contribution.
I took an initial look at the changes and the overall approach looks good to me. 
However, we have a failing build here: https://github.com/auth0/nextjs-auth0/actions/runs/24439033359/job/71422348405?pr=2602

Some reccomendations:
- Please fix the failing tests as described above.
- Since this is new behaviour, tests do not exist for this yet. It would be great if you could add tests for this so that future builds are correct.
- This repo requires commit signing, I see that your commit does not have a signature. Please take a look at [Signing Commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and sign your commits with your GPG key.

Thanks!

PR #2608, opened 2026-04-15
feat(passwordless): Add Next.js route handlers for passwordless flows

Piyush-85:
# Passwordless support - PR 4

## Summary

This PR is the **Next.js layer** — it exposes passwordless as HTTP routes that the frontend can POST to, and wires everything into the public `Auth0Client` API.

Background: the frontend can't call Auth0 directly because client credentials live server-side. Instead, it POSTs to the app's own `/auth/*` routes. The SDK ships a single catch-all route file (`app/auth/[auth0]/route.ts`) that intercepts every `/auth/*` request and dispatches it to the right handler based on the pathname. This PR adds `POST /auth/passwordless/start` and `POST /auth/passwordless/verify` to that dispatch table.

### `src/server/auth-client.ts`

Adds two route handler methods:

- `handlePasswordlessStart()` — parses the incoming request, validates the connection type, and calls PR3's `passwordlessStart()`. For magic link requests (`send: "link"`), it also generates a full PKCE set (code verifier, code challenge, state, nonce), passes it to `passwordlessStart()` via `authParams`, and saves a `TransactionState` cookie — so that when the user clicks the emailed link and lands on `/auth/callback`, the existing callback handler can complete the PKCE exchange without any extra steps.
- `handlePasswordlessVerify()` — parses the OTP submission, calls PR3's `passwordlessVerify()`, and then creates and sets the session cookie. Only handles `email` and `sms` connections — magic links go through `/auth/callback`, not this route.

### `src/server/client.ts`

Adds `POST /auth/passwordless/start` and `POST /auth/passwordless/verify` to the main handler dispatch table. Also exposes `auth0.passwordless` as a getter returning a `ServerPasswordlessClient` instance.

### `src/server/passwordless/server-passwordless-client.ts` (new)

`ServerPasswordlessClient` — the server-side public API, returned by `auth0.passwordless` on the `Auth0Client` instance. Exposes:

- `start(options)` — delegates to `authClient.passwordlessStart()` (the raw HTTP call). **Note:** calling `auth0.passwordless.start()` with `send: "link"` from a Server Action does **not** set up PKCE — PKCE is only generated by the HTTP route handler (`POST /auth/passwordless/start`). Callers who need magic link support should POST to the route rather than calling `start()` directly.
- `verify(options)` — App Router overload; reads/writes cookies via `next/headers`.
- `verify(req, res, options)` — Pages Router / Middleware overload with explicit `req`/`res`.

Same dual-overload pattern as `ServerMFAClient`.

### `src/server/passwordless-server.flow.test.ts` (new)

Flow tests for the route handlers, instantiating `AuthClient` directly with MSW mocking HTTP. Follows the same pattern as `mfa-server.flow.test.ts`. Covers:

- `passwordlessStart`: correct request body for email and SMS connections; `PasswordlessStartError` on Auth0 API error and network failure
- `passwordlessVerify`: correct token endpoint params (`realm`/`otp`/`username`) for email and SMS; `PasswordlessVerifyError` on `invalid_grant` and discovery failure
- Route handler `POST /auth/passwordless/start`: 204 for valid email/SMS start; 400 for missing or unknown connection
- Route handler `POST /auth/passwordless/verify`: 200 + Set-Cookie + session created for email and SMS; 400/403 for invalid input and Auth0 errors

### `src/test/defaults.ts`, `src/test/mcd-test-fixtures.ts`, `src/server/dynamic-base-url.test.ts`

Test utilities and fixtures extended to include `passwordlessStart` and `passwordlessVerify` route entries, keeping them consistent with the new routes added to `client.ts`.

## Depends on

[feat(passwordless): add passwordlessStart and passwordlessVerify to AuthClient](#)

## Summary by CodeRabbit

* **New Features**
  * Server-side passwordless auth: email (OTP & magic link) and SMS start/verify endpoints, plus automatic session creation on successful verify
  * Server-side passwordless client API for both App and Pages Router usage and a server passwordless getter

* **Documentation**
  * Updated passwordless option docs (language header and authParams examples)

* **Tests**
  * Expanded integration tests covering passwordless flows, routes, DPoP retry, and security cases

* **Chore**
  * Transaction state now allows an optional PKCE verifier

[Review Change Stack](https://app.coderabbit.ai/change-stack/auth0/nextjs-auth0/pull/2608?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/server/passwordless/server-passwordless-client.ts
```diff
@@ -0,0 +1,116 @@
...
+ * }
+ * ```
+ */
+export class ServerPasswordlessClient implements PasswordlessClient {
+  constructor(private provider: AuthClientProvider) {}
+
+  private async getAuthClient(): Promise<AuthClient> {
+    const { headers } = await import("next/headers.js");
+    const reqHeaders = await headers();
+    return this.provider.forRequest(reqHeaders, undefined);
```
tusharpandey13: `forRequest(reqHeaders, undefined)` : second arg is `nextUrl` which MCD resolver uses to determine which custom domain to route to. Passing `undefined` means in multi-domain deployments, `auth0.passwordless.start()` called from server components always hits the static/default domain rather than the one matching the user's request origin.

The `PasswordlessClient` interface is shared between browser and server, but their capabilities differ : server has access to `NextRequest`/`URL` context that browser doesn't. Having the server contract diverge slightly via overloads (like `verify` already does with `(req, res, options)`) would let MCD resolver work here. The browser implementation just wouldn't use that overload.

Worth noting: `ServerMfaClient` (`src/server/mfa/server-mfa-client.ts:58`) has the same gap. MFA shipped before MCD landed (v4.17.0), so it's not a deliberate design choice : it's pre-existing tech debt. Both passwordless and MFA server clients would benefit from accepting optional request context for MCD resolution.

src/server/auth-client.ts
```diff
@@ -3670,6 +3873,54 @@ export class AuthClient {
...
+    if (!tokenResponse.id_token) {
+      throw new PasswordlessVerifyError(
+        "missing_id_token",
+        "No id_token in passwordless verify response. Ensure 'openid' scope is requested."
+      );
+    }
+
+    // jose.decodeJwt does a non-validating base64 decode — safe here because the
+    // token was received directly from Auth0's HTTPS token endpoint.
+    const claims = jose.decodeJwt(tokenResponse.id_token);
```
tusharpandey13, line 3958: The trust model here (HTTPS origin, server-to-server, no browser in path) matches how MCD domain inference uses `decodeJwt` on already-trusted id_tokens from session store (line 2652). Both skip signature verification because the token source is already authenticated.

For comparison, backchannel logout (line 2395) uses full `jose.jwtVerify()` with issuer/audience/algorithm checks : but that's because logout tokens arrive from external sources and could be spoofed.

Given the server-to-server trust model, skipping signature verification is defensible. But a lightweight `iss`/`aud` sanity check (no JWKS fetch needed, just compare decoded claims against `this.issuer` and `this.clientMetadata.client_id`) would catch configuration errors early, wrong tenant or wrong client_id would surface as a clear error rather than silently creating a session with mismatched claims.

src/server/passwordless/server-passwordless-client.ts
tusharpandey13, line 1: **Line:** ~3686 (inside `passwordlessStart`)
The `/passwordless/start` endpoint respects `x-request-language` for email template localization. Currently the SDK doesn't forward the user's locale preference. Not blocking for initial release, jsut notifying.

tusharpandey13 (commented):
PTAL

tusharpandey13:
Current tests cover happy paths and basic error cases well. A few scenarios that would increase confidence:
- DPoP-enabled passwordless verify (nonce retry, DPoP header presence)
- Scope/audience injection attempt (verify that user-supplied values can't override global config)
- MCD resolver mode (multi-domain routing for passwordless endpoints)

src/server/auth-client.ts
Piyush-85: oauth4webapi's processGenericTokenEndpointResponse already performs full id_token validation — JWKS signature verification + iss/aud claim comparison against the discovered ASM — and throws with code === JWT_CLAIM_COMPARISON on mismatch. These errors were previously falling through to the generic unknown_error catch.

The fix intercepts JWT_CLAIM_COMPARISON and re-throws as named PasswordlessVerifyError("invalid_issuer") / PasswordlessVerifyError("invalid_audience"). Discrimination uses err.cause.claim ('iss' | 'aud') — structured data that oauth4webapi explicitly sets on the error — rather than message string-sniffing, so it won't silently regress if error messages change. Added test cases for both mismatch scenarios.

src/server/passwordless/server-passwordless-client.ts
Piyush-85: Fixed

Piyush-85:
> Current tests cover happy paths and basic error cases well. A few scenarios that would increase confidence:
> 
> * DPoP-enabled passwordless verify (nonce retry, DPoP header presence)
> * Scope/audience injection attempt (verify that user-supplied values can't override global config)
> * MCD resolver mode (multi-domain routing for passwordless endpoints)

Added additional tests to cover these

tusharpandey13 (dismissed):
Approved

A small follow-up:
`allow_magiclink_verify_without_session` : Is this documented customer-facing? Users enabling magic link will need this tenant setting.

Piyush-85:
> Approved
> 
> A small follow-up: `allow_magiclink_verify_without_session` : Is this documented customer-facing? Users enabling magic link will need this tenant setting.

No, this is a gap that exists today in Auth0 customer facing documentation. 
Will look into getting this added. As of today users trying magic link with RWAs face issue

src/server/passwordless-server.flow.test.ts
```diff
@@ -0,0 +1,1270 @@
...
+} from "../test/defaults.js";
+import { generateSecret } from "../test/utils.js";
+import type { SessionData } from "../types/index.js";
+import { generateDpopKeyPair } from "../utils/dpopRetry.js";
+import { AuthClientProvider } from "./auth-client-provider.js";
+import { AuthClient } from "./auth-client.js";
+import { encrypt } from "./cookies.js";
+import { ServerPasswordlessClient } from "./passwordless/server-passwordless-client.js";
+import { StatelessSessionStore } from "./session/stateless-session-store.js";
+import { TransactionStore } from "./transaction-store.js";
```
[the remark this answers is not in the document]
Piyush-85: This is not valid. It's applying a consumer-facing rule to internal SDK test code. The finding is a false positive.

tusharpandey13 (approved):

PR #2604, opened 2026-04-15
fix: defer AuthClientProvider construction when AUTH0_DOMAIN not set at build time

sleitor:
## Problem

PR #2545 (MCD support) introduced `AuthClientProvider` which validates the `domain` option eagerly in its constructor. If `domain` is neither a `string` nor a `DomainResolver` function (i.e. `undefined`), it throws:

> `InvalidConfigurationError: You must provide either a domain string or a DomainResolver function.`

This breaks Next.js **standalone** builds where `AUTH0_DOMAIN` is injected at container/runtime and is **not** available during `next build`. Prior to v4.17.0, domain was only validated at request time.

## Fix

When `domain` is `undefined` at module evaluation time, pass a `DomainResolver` function to `AuthClientProvider` instead of `undefined`. The resolver reads `AUTH0_DOMAIN` lazily on the first request — restoring the pre-4.17.0 behaviour.

If `AUTH0_DOMAIN` is still absent at request time, the resolver throws `InvalidConfigurationError` with a clear message.

When `domain` is already a `string` or a `DomainResolver` function, it is passed through unchanged.

## Testing

All 1252 existing unit tests pass.

Fixes #2601

PR #2609, opened 2026-04-16
test(client): add tests for deferred domain resolution in standalone builds

sleitor:
Follow-up to #2604.

tusharpandey13 noted in the review that a test should be added to prevent regressions. This PR adds three tests covering the deferred domain resolution behaviour:

1. **No throw at construction** — `new Auth0Client()` must succeed even when `AUTH0_DOMAIN` is absent (Next.js standalone build scenario).
2. **Domain resolved at request time** — when `AUTH0_DOMAIN` is injected after construction (container runtime), the first `getSession()` call succeeds.
3. **Clear error at request time** — when `AUTH0_DOMAIN` is still absent at request time, a `DomainResolutionError` is thrown with an `InvalidConfigurationError` as `.cause`, containing a descriptive message.

### DIFF
--- a/src/server/client.test.ts
+++ b/src/server/client.test.ts
@@ -1,7 +1,10 @@
-import { NextResponse } from "next/server.js";
+import { NextRequest, NextResponse } from "next/server.js";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
-import { InvalidConfigurationError } from "../errors/index.js";
+import {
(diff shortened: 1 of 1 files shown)


sleitor:
@tusharpandey13 @Piyush-85 — follow-up on your review comment from #2604. Added the tests you requested!

Piyush-85 (approved):

PR #2667, opened 2026-05-16
Expose default oncallback

Piyush-85:
### Summary                                                                                               
                                                                                                              
  The onCallback hook currently requires users to re-implement the SDK's default redirect/error behaviour from
   scratch if they want any custom side effects (e.g. syncing the user to a database). This change passes the 
  SDK's built-in callback handler as a fourth argument (defaultOnCallback) to every onCallback invocation, so 
  custom hooks can run their own logic and then delegate to the default behaviour without duplicating it.     
                                                                                                            
  Changes:                                                                                                    
  - Added defaultOnCallback as the fourth parameter of OnCallbackHook, typed as (error, ctx) => 
  Promise<NextResponse>                                                                                       
  - Passed this.defaultOnCallback.bind(this) at all onCallback call sites (including the popup/postMessage  
  flow added after this was originally authored)                                                              
  - Added full JSDoc to OnCallbackHook with @param descriptions and a usage example — surfaces on hover in VS 
  Code and WebStorm                                                                                          
                                                                                                              
  Usage                                                                                                     
                                                                                                              
```
  const auth0 = new Auth0Client({                                                                           
    async onCallback(error, ctx, session, defaultOnCallback) {                                                
      if (!error && session) {                                                                                
        await db.upsertUser(session.user); // your side effect                                                
      }                                                                                                       
      return defaultOnCallback(error, ctx); // keep default redirect behaviour                                
    }                                                                                                         
  });  
```                                                                                                       
                                                                                                              
  Previously, users had to manually inspect ctx.returnTo and construct a NextResponse.redirect(...) themselves
   to replicate what the SDK does by default.                                                               
                                                                                                                                                                                                                                                          
  ---                                                                                                         
  Co-authored-by: Dan Haughey (greenlynx) — original implementation in #2488

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -2535,11 +2535,12 @@ Do realize that this has an impact on the size of the cookie being issued, so it
 
 The `onCallback` hook is run once the user has been redirected back from Auth0 to your application with either an error or the authorization code which will be verified and exchanged.
 
-The `onCallback` hook receives 3 parameters:
+The `onCallback` hook receives 4 parameters:
 
 1. `error`: the error returned from Auth0 or when attempting to complete the transaction. This will be `null` if the transaction was completed successfully.
 2. `context`: provides context on the transaction that initiated the transaction.
 3. `session`: the `SessionData` that will be persisted once the transaction completes successfully. This will be `null` if there was an error.
+4. `defaultOnCallback`: the default onCallback hook that would have been called instead of the user-provided hook. This can be used if you want to take an action on callback, but keep the default callback behaviour instead of overriding it.
 
 The hook must return a Promise that resolves to a `NextResponse`.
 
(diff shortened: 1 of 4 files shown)


Piyush-85:
After reviewing this more carefully, we've decided not to move ahead with exposing defaultOnCallback as a
parameter. The default callback behaviour is intentionally simple, a redirect to returnTo on success and a
500 on error and users who need side effects can replicate that in a few lines, as shown in our existing
docs example.

The bigger concern is the popup flow: defaultOnCallback would produce a redirect response there, which is
the wrong behaviour for a postMessage flow. Making the contract accurate for popup would add meaningful
complexity for a convenience gain that doesn't justify it.

Closing this out, but appreciate the contribution.

PR #2690, opened 2026-06-02
feat: cte delegation

Piyush-85:

### 📋 Changes

Adds delegation and impersonation support to Custom Token Exchange (CTE) per the SDK requirements doc and [RFC 8693 §4.1](https://datatracker.ietf.org/doc/html/rfc8693#section-4.1).

#### Types added

- **`ActClaim`** (`src/types/index.ts`) — represents the `act` claim in an ID token issued via token exchange. `sub` is required (per RFC 8693), supports nested `act` for delegation chains.
- **`act?: ActClaim`** added to the `User` interface — makes the actor claim accessible on `session.user.act` after a session-creating flow.
- **`act?: ActClaim`** added to `CustomTokenExchangeResponse` (`src/types/token-vault.ts`) — the decoded actor claim is returned directly on the response object, so callers do not need to manually decode the `idToken` JWT.

#### Behaviour changes

**`customTokenExchange()`**

- `actorTokenType` is now validated client-side using the same URI rules as `subjectTokenType` (10–100 chars, valid URL or URN format). The error message explicitly names `actor_token_type` to distinguish it from `subjectTokenType` errors.
- The returned `id_token` is decoded (via `jose.decodeJwt`) and the `act` claim is extracted and included on the response as `result.act`. No additional network call or dependency — `jose` is already imported.
- `refreshToken` JSDoc updated to document Auth0's behaviour of suppressing the refresh token when `actor_token` is present in the request.

**`filterDefaultIdTokenClaims` / `DEFAULT_ID_TOKEN_CLAIMS`** (`src/server/user.ts`)

- `"act"` added to `DEFAULT_ID_TOKEN_CLAIMS`. Without this, `filterDefaultIdTokenClaims` silently strips the `act` claim from every session built from an ID token (standard OIDC callback, passwordless, passkey, etc.), even when an Auth0 Action sets it via `api.authentication.setActor()`.

#### Usage

```typescript
// Delegation — service acting on behalf of a user
const result = await auth0.customTokenExchange({
  subjectToken: userToken,
  subjectTokenType: 'urn:acme:legacy-token',
  actorToken: agentJwt,
  actorTokenType: 'http://corporate-idp/id-token',
  audience: 'https://api.example.com',
});

// act claim decoded and available directly
console.log(result.act);
// { sub: 'agent|abc123' }

// act claim also available on the session user object (when set via Action in a session-creating flow)
const session = await auth0.getSession();
console.log(session?.user?.act);
// { sub: 'agent|abc123' }
```

### 📎 References

- [RFC 8693 — OAuth 2.0 Token Exchange](https://datatracker.ietf.org/doc/html/rfc8693)
- [RFC 8693 §4.1 — `act` claim](https://datatracker.ietf.org/doc/html/rfc8693#section-4.1)

## Summary by CodeRabbit

* **New Features**
  * Actor token format validation for custom token exchanges (URI/URN enforcement).
  * SDK returns decoded actor (act) claim from token responses and exposes it on session user objects; supports nested delegation chains.
  * Refresh tokens may be suppressed when an actor token is present; access tokens still returned.

* **Documentation**
  * Examples updated to show actor claim usage, session exposure, and refresh-token behavior.

* **Tests**
  * Expanded coverage for actor token validation, act parsing, and delegation/impersonation flows.


EXAMPLES.md
```diff
@@ -4113,18 +4113,32 @@ const result = await auth0.customTokenExchange({
 
 ### With Actor Token (Delegation)
 
-For delegation scenarios where a service acts on behalf of a user:
+For delegation scenarios where a service or agent acts on behalf of a user (RFC 8693 §4.4). The `act` claim in the returned ID token identifies the acting party and is automatically decoded and returned on the response:
```
kishore7snehil: This should be [Section 4.1](https://datatracker.ietf.org/doc/html/rfc8693#section-4.1)

kishore7snehil (changes requested):

src/server/auth-client.ts
```diff
@@ -2987,6 +2987,25 @@ export class AuthClient {
...
 
+    // Validate actorTokenType URI (same rules as subjectTokenType)
+    if (options.actorToken && options.actorTokenType) {
+      const actorTokenTypeError = this.validateSubjectTokenType(
+        options.actorTokenType
+      );
+      if (actorTokenTypeError) {
+        return [
+          new CustomTokenExchangeError(
+            actorTokenTypeError.code,
```
kishore7snehil: The error code here stays `INVALID_SUBJECT_TOKEN_TYPE` even though the failure is on the actor token type. Callers doing `switch (error.code)` will land in the subject-token branch for an actor-token problem. Since the enum already distinguishes `MISSING_ACTOR_TOKEN_TYPE`, suggest adding a dedicated `INVALID_ACTOR_TOKEN_TYPE` and returning it here.

src/server/auth-client.ts
```diff
@@ -2987,6 +2987,25 @@ export class AuthClient {
...
+            actorTokenTypeError.message.replace(
```
kishore7snehil: We can use `.replaceAll(...)` in place of `.replace()` as it only swaps the first occurrence.Right now, it's harmless with today's single-instance message, but fragile if the validation message ever contains `subject_token_type` twice.

src/server/auth-client.ts
```diff
@@ -3087,6 +3106,13 @@ export class AuthClient {
       ];
     }
 
+    // Decode act claim from ID token for delegation/impersonation flows (RFC 8693 §4.4)
+    const act = tokenEndpointResponse.id_token
+      ? (jose.decodeJwt(tokenEndpointResponse.id_token).act as
```
kishore7snehil: This `jose.decodeJwt(...)` sits outside the try/catch that wraps the token request . A malformed `id_token` would throw an unhandled exception instead of returning the clean `[error, null]` tuple the rest of the method guarantees. Consider wrapping it.

src/types/index.ts
```diff
@@ -80,6 +80,20 @@ export interface SessionDataStore {
 
 export type LogoutToken = { sub?: string; sid?: string; iss?: string };
 
+/**
+ * Represents the `act` (actor) claim from an ID token issued via RFC 8693 delegation.
+ * The `act` claim identifies the acting party and may be nested to represent a delegation chain.
+ *
+ * @see {@link https://datatracker.ietf.org/doc/html/rfc8693#section-4.4 RFC 8693 §4.4}
```
kishore7snehil: This should reference [Section 4.1](https://datatracker.ietf.org/doc/html/rfc8693#section-4.1). Section 4.4 is the `may_act` (Authorized Actor) claim, which is a different claim expressing who is *permitted* to act. The `act` (Actor) claim is defined in Section 4.1.

src/server/auth-client.ts
```diff
@@ -3087,6 +3106,13 @@ export class AuthClient {
       ];
     }
 
+    // Decode act claim from ID token for delegation/impersonation flows (RFC 8693 §4.4)
```
kishore7snehil: Same RFC reference fix here - the `act` claim is defined in [Section 4.1](https://datatracker.ietf.org/doc/html/rfc8693#section-4.1), not 4.4 (4.4 is `may_act`).

src/types/token-vault.ts
```diff
@@ -137,12 +137,21 @@ export interface CustomTokenExchangeResponse {
...
+   */
   refreshToken?: string;
   /** Token type, typically "Bearer" or "DPoP" */
   tokenType: string;
   /** Token lifetime in seconds */
   expiresIn: number;
   /** Granted scopes */
   scope?: string;
+  /**
+   * The actor claim decoded from the ID token, present in delegation/impersonation flows (RFC 8693 §4.4).
```
kishore7snehil: RFC reference should be [Section 4.1](https://datatracker.ietf.org/doc/html/rfc8693#section-4.1) here too (4.4 is `may_act`, a separate claim).

src/server/user.ts
```diff
@@ -12,7 +12,8 @@ export const DEFAULT_ID_TOKEN_CLAIMS = [
   "picture",
   "email",
   "email_verified",
-  "org_id"
+  "org_id",
+  "act"
```
kishore7snehil, line 16: `filterDefaultIdTokenClaims` already has a test in `src/server/user.test.ts` that asserts the whitelist behavior, but it wasn't updated for act. Worth adding act to the input claims there.

kishore7snehil (changes requested):

src/server/user.ts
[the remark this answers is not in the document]
Piyush-85: The comment is not valid. act is an identity claim from the ID token (RFC 8693 §4.1), not a credential or sensitive value. session.user is the correct place for ID token identity claims — the same as sub, email, org_id. handleProfile already returns all of session.user to the client, including every custom claim an Action sets via the [key: string]: unknown index signature. 
Moving act to session.internal would be semantically wrong — that field is SDK infrastructure (sid, createdAt, mcd), not user identity.

src/server/auth-client.ts
Piyush-85: This is replaced by validateSubjectTokenType() + validateActorTokenType() + validateTokenTypeUri() where each method constructs its own message directly with the field name injected. There is no replace() call anywhere in that path anymore.

src/server/auth-client.ts
Piyush-85: Moved it inside an existing try catch, aligned with other methods.

EXAMPLES.md
Piyush-85: fixed

src/server/auth-client.ts
Piyush-85: Fixed, added INVALID_ACTOR_TOKEN_TYPE

src/types/index.ts
Piyush-85: done

src/types/token-vault.ts
Piyush-85: done

src/server/user.ts
Piyush-85: done

src/server/auth-client.ts
Piyush-85: done

tusharpandey13 (approved):

kishore7snehil (approved):
LGTM!

PR #2702, opened 2026-06-11
fix: surface MfaRequiredError from passkeyGetToken and passwordlessVe…

Piyush-85:

  ### 📋 Changes

  `passkeyGetToken` and `passwordlessVerify` previously swallowed `mfa_required` errors from Auth0, leaving callers with no way to initiate an MFA challenge
  after a first-factor passkey or passwordless login.

  **What changed:**

  - Added a private `#throwIfMfaRequired` helper on `AuthClient` that detects `mfa_required` responses, encrypts the `mfa_token` as a JWE (using the session
  secret + `authorizationParameters` audience/scope), and throws `MfaRequiredError` — consistent with how the refresh token path already behaves.
  - `passkeyGetToken` and `passwordlessVerify` now call `#throwIfMfaRequired` before falling through to their own typed errors. If `mfa_token` is absent in the
  response the helper is a no-op and the caller throws its own error as before.
  - `handlePasskeyGetToken` and `handlePasswordlessVerify` route handlers now catch `MfaRequiredError` and return a `403` JSON response via
  `#createMfaRequiredResponse` without setting a session cookie — no session exists at this point in the flow.
  - `#createMfaRequiredResponse` signature widened from `session: SessionData` to `session: SessionData | null` to support these sessionless first-factor flows.

  ### 📎 References

  - Mirrors the existing `MfaRequiredError` handling in the `getAccessToken` / refresh token path.

  ### 🎯 Testing

  New unit tests added across four files:

  - **`mfa-error-bubbling.test.ts`** — full MFA error contract for both `passkeyGetToken` and `passwordlessVerify`: throws `MfaRequiredError`, JWE format
  (5-part base64url), token is encrypted (not raw), decryptable with SDK secret, `audience`/`scope` reflect `authorizationParameters`, `.code`, `.cause`
  (OAuth2Error), `toJSON()` shape, and `mfa_requirements` passthrough.
  - **`passkey.flow.test.ts`** — no-`mfa_token` fallthrough: when Auth0 returns `mfa_required` without a token, `#throwIfMfaRequired` is a no-op and
  `PasskeyGetTokenError` is thrown instead.
  - **`passkey-server.flow.test.ts`** — route handler returns `403` with encrypted `mfa_token`; no session cookie is set.
  - **`passwordless.flow.test.ts`** — mirrors passkey fallthrough test for `PasswordlessVerifyError`.
  - **`passwordless-server.flow.test.ts`** — mirrors passkey server tests for the passwordless route handler.

### Breaking 
This is a breaking change to the return type of mfa.verify() — callers who destructured { access_token, refresh_token, ... } from the result will get  undefined at runtime instead of real values. Even though the deprecated fields remain in the type signature (soft deprecation), the runtime behavior changes from "returns token data" to "returns { success: true }". 
This must ship as a minor version bump (per semver: adding success as a required field and changing runtime behavior is a breaking behavioral change, not just a patch).

## Release Notes

* **New Features**
  * Added step-up MFA (OTP) support to passkey and passwordless example flows, including new “Step-up MFA test” UI.
  * Updated examples to use QR-code enrollment and MFA OTP verification for both methods.

* **Bug Fixes**
  * Improved MFA-required handling: verification now stores tokens in the session cookie and returns `{ success: true }` (clients fetch the access token afterward).
  * When MFA is required, the API returns a 403 with an encrypted `mfa_token`, and avoids setting a session cookie if no session exists.

* **Tests**
  * Expanded coverage for `mfa_required` across server and client passkey/passwordless flows.


src/server/auth-client.ts
```diff
@@ -1565,22 +1565,38 @@ export class AuthClient {
...
-      const res = NextResponse.json(result);
-
-      // Cache tokens in session if session exists
       const { error: sessionError, session } =
         await this.getSessionWithDomainCheck(req.cookies);
+
       if (!sessionError && session) {
+        // Step-up MFA: session exists — cache new token and return success.
+        // Tokens are stored in the session cookie; the body must not expose them.
+        const res = NextResponse.json({ success: true });
```
nandan-bhat: This changes what `handleVerify` sends back. Before it returned the full token set, now it returns `{ success: true }`. I think hiding the tokens from the body is a good move, but the client side `mfa.verify()` is still typed as `Promise<MfaVerifyResponse>` and its JSDoc still says it returns `access_token, refresh_token, etc` (src/client/mfa/index.ts and src/types/mfa.ts). So a caller reading `result.access_token` after this change will just get `undefined` with no type error. Was the plan to also update the client return type and the docs to match the new `{ success: true }` shape ? If yes, could we do that in this PR so the types do not lie ?

src/server/auth-client.ts
```diff
@@ -3594,17 +3650,21 @@ export class AuthClient {
...
+    // Validate mfaRequirements has challenge types only when requirements are present.
+    // When mfa_requirements is absent (e.g. webauthn grant), all challenge types are allowed.
+    if (context.mfaRequirements?.challenge) {
+      const allowedTypes = new Set(
+        context.mfaRequirements.challenge.map((c) => c.type.toLowerCase())
       );
+
+      if (
+        allowedTypes.size === 0 ||
+        !allowedTypes.has(challengeType.toLowerCase())
```
nandan-bhat: Small thing on the error message here. This branch now also throws when the requested `challengeType` is not in the allowed set, but the message still says "No MFA challenge types available in mfa_requirements". In that case factors are available, the user just asked for one that is not allowed, so the message can be a bit confusing when debugging. Could we use a message that says the requested challenge type is not allowed ?

Piyush-85: Valid. The original single condition allowedTypes.size === 0 || !allowedTypes.has(...) threw the same generic message for two distinct cases. When factors exist but the requested type isn't in the allowed set, the message "No MFA challenge types available" is factually wrong. 
Split into two separate checks: the first retains the original message for the empty-set case; the second now throws Challenge type "${challengeType}" is not allowed by mfa_requirements. Allowed: otp, oob (with the actual allowed list), making it immediately actionable.

src/server/auth-client.ts
Piyush-85: Updated.
This is a breaking change to the return type of mfa.verify() — callers who destructured { access_token, refresh_token, ... } from the result will get undefined at runtime instead of real values. Even though the deprecated fields remain in the type signature (soft deprecation), the runtime behavior changes from "returns token data" to "returns { success: true }". 
This must ship as a minor version bump (per semver: adding success as a required field and changing runtime behavior is a breaking behavioral change, not just a patch).

examples/with-passkeys/components/step-up-button.tsx
```diff
@@ -0,0 +1,173 @@
...
+        audience: "https://api.example.com",
+        scope: "read:users write:users"
+      });
+      alert(`Access token received — check the browser console.\n\n${token}`);
+    } catch (err) {
+      if (err instanceof MfaRequiredError) {
+        try {
+          const authenticators = await mfa.getAuthenticators({
+            mfaToken: err.mfa_token
+          });
```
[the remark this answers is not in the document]
Piyush-85: Not valid. There is no project rule requiring error.code checks over instanceof for Auth0 SDK errors — this guideline was fabricated by the bot. instanceof MfaRequiredError is the idiomatic TypeScript pattern and exactly what the SDK is designed for: MfaRequiredError is a typed class export precisely so callers can use instanceof. Using error.code === "mfa_required" would bypass TypeScript's type narrowing — after an instanceof check, err  is narrowed to MfaRequiredError and err.mfa_token is accessible without a cast. With a .code string check, the type remains unknown/any and the caller  must cast manually. The instanceof form is more type-safe and more readable. No change needed.

nandan-bhat (approved):

PR #2708, opened 2026-06-14
feat: enforce IPSIE session_expiry ceiling on enterprise connections

Piyush-85:
- [x] I have added documentation for all new/changed functionality

### 📋 Changes

IPSIE SL1 requires SDKs to honour an upstream IdP-asserted session ceiling for enterprise connections. Without this, a local Next.js session can outlive the IdP session — bypassing the security guarantee.

When an enterprise connection has `id_token_session_expiry_supported: true`, Auth0 emits a `session_expiry` Unix-timestamp claim in the ID token. The SDK now stamps this as a hard, non-extendable ceiling and enforces it at three points:

  - `getSession()` / `getUser()` return `null` once the ceiling is reached, identical to a logged-out user. Stateful session store records are deleted at that point.
  - `getAccessToken()` returns a new `SESSION_EXPIRED` error code immediately, without making any outbound call to `/oauth/token`.
  - If the ceiling is already in the past at login (relative to the ID token `iat`), the callback rejects the login rather than persisting an already-expired session.

The ceiling survives token refreshes and is never overwritten by the refresh path. Sessions without a `session_expiry` claim are completely unaffected — **no breaking changes.**

  ### 📎 References

  - IPSIE SL1 specification
  - Python reference implementation: `auth0/auth0-server-python#117`

  ### 🎯 Testing

  Unit tests cover both helper functions (`isSessionCeilingReached`, `isSessionCeilingInPast`) and all three enforcement paths. 

## Summary by CodeRabbit

* **New Features**
  * Added an end-to-end session expiry demo with a countdown, status panels, and protected/API checks.
  * Exposed a new `session_expired` error state for access token requests.

* **Bug Fixes**
  * Sessions now stop working once an upstream identity provider’s expiry ceiling is reached.
  * Login flows reject already-expired sessions immediately and clean up stale session data.

* **Documentation**
  * Added README and example docs explaining session expiry behavior, leeway, and recovery steps.


src/server/auth-client.ts
```diff
@@ -2645,6 +2711,20 @@ export class AuthClient {
       };
     }
 
+    // IPSIE: treat as no session once the upstream IdP-asserted ceiling passes.
+    // Returns the same shape as "no session" so existing redirect-to-login paths
+    // fire transparently. For stateful stores, delete the backing record so the
+    // store stays clean; stateless sessions are self-contained in the cookie and
+    // will be orphaned (ceiling check fires on every subsequent read).
+    if (isSessionCeilingReached(session.internal?.sessionExpiresAt)) {
```
kishore7snehil: This ceiling gate also ends up blocking connection-token retrieval, which I don't think we want. 

`getAccessTokenForConnection` reads the session through `getSessionFromAuthClient` -> `getSessionWithDomainCheck` , and once `sessionExpiresAt` passes this returns `null`, so the connection path then throws `MISSING_SESSION` and connection-token fetches stop working.

Connection tokens follow the upstream IdP's own `expires_in`, not the app-session ceiling, so they shouldn't be bound by it. Can we read the raw session (bypassing the ceiling gate) on the connection path, or special-case it so connection-token retrieval still works after the ceiling?

src/utils/session-helpers.ts
```diff
@@ -107,6 +137,11 @@ export function buildSessionFromCallback(
   oidcRes: oauth.TokenEndpointResponse,
   transactionState: TransactionState
 ): SessionData {
+  const sessionExpiresAt =
+    typeof idTokenClaims.session_expiry === "number"
```
kishore7snehil: We only check `typeof === "number"` here, so any number is stamped as the ceiling. Since this claim is set by a customer Post-Login Action, it's untrusted input  and the most likely mistake is emitting milliseconds (`Date.now()` instead of `Date.now()/1000`). A 13-digit ms value read as seconds lands ~year 57000, sails past the login guard, and `isSessionCeilingReached` never fires.

We should add a cap: reject `<= 0` and `>= 10_000_000_000` and fall open to "no ceiling."

src/server/auth-client.ts
```diff
@@ -1344,6 +1346,27 @@ export class AuthClient {
...
         );
+
+        // IPSIE: reject a session whose ceiling is already in the past at login.
+        if (
+          isSessionCeilingInPast(
+            fallbackSession.internal.sessionExpiresAt,
+            idTokenClaims?.iat
+          )
+        ) {
+          const popupResponse = createAuthCompletePostMessageResponse({
```
kishore7snehil, line 1469: In the popup branch, this at-login rejection returns the postMessage response directly without invoking `onCallback`. The redirect branch goes through `handleCallbackError`, which does call `onCallback`. 

So a developer's `onCallback` error hook fires for an already-expired ceiling on redirect but not on popup. Shall we `await this.onCallback(error, onCallbackCtx, null)` before returning here?

src/server/session/stateful-session-store.ts
```diff
@@ -254,4 +254,25 @@ export class StatefulSessionStore extends AbstractSessionStore {
...
+  ): Promise<void> {
+    const cookieValue =
+      reqCookies.get(this.sessionCookieName)?.value ||
+      reqCookies.get(LEGACY_COOKIE_NAME)?.value;
+
+    if (!cookieValue) {
+      return;
+    }
+
+    const session = await cookies.decrypt<SessionCookieValue>(
```
kishore7snehil: `deleteByReqCookies` calls `decrypt` directly, but `get()` (lines ~92-100) falls back to `verifySigned` on `ERR_JWE_INVALID` to support legacy signed v3 cookies. So when the ceiling is reached on a request carrying a legacy cookie, `decrypt` returns null and the backing store record is never deleted, even though this method's whole purpose is store hygiene. 

Shall we mirror the `get()` fallback (decrypt with `throwOnJWEErrors`, then `verifySigned` on the legacy error) so read and delete stay consistent.

src/utils/session-helpers.ts
```diff
@@ -3,6 +3,36 @@ import * as oauth from "oauth4webapi";
...
+  return now >= sessionExpiresAt - SESSION_EXPIRY_LEEWAY_SECONDS;
+}
+
+/**
+ * Returns true when the session ceiling is already in the past at login time.
+ * Compares against the ID token `iat` claim (falls back to wall-clock now).
+ * Used to reject an already-expired session before persisting it.
+ * A missing/undefined ceiling returns false (safe default — no ceiling).
+ */
+export function isSessionCeilingInPast(
```
kishore7snehil, line 26: `isSessionCeilingInPast` adds the leeway (`sessionExpiresAt <= reference + LEEWAY`) while `isSessionCeilingReached` subtracts it (`now >= sessionExpiresAt - LEEWAY`). The additive-vs-subtractive direction is non-obvious and creates a ~60s window where a very-short IdP-granted session is **rejected at login** but would've been accepted on read.

src/server/session/abstract-session-store.ts
```diff
@@ -179,6 +179,23 @@ export abstract class AbstractSessionStore {
...
+   *
+   * For stateless cookie sessions this is a no-op — the cookie is the session, and
+   * clearing it requires response cookies. Ceiling enforcement returns null on every
+   * read anyway, so the orphaned cookie is harmless until its natural max-age expires.
+   *
+   * Subclasses that own a backing store (stateful) SHOULD override this to delete
+   * the store record by session ID for hygiene.
+   */
+
+  abstract deleteByReqCookies(
```
kishore7snehil: `AbstractSessionStore` is exported publicly (from `src/server/index.ts`), so adding an `abstract` method with no default can be compile-time break for anyone who has subclassed it directly. WDYT?

src/server/auth-client.ts
```diff
@@ -1800,6 +1842,18 @@ export class AuthClient {
           status: 401
         }
       );
+
+      // IPSIE: when the ceiling fired, clean up the stored session so the next
+      // request starts clean. For stateless sessions this is a no-op; for
+      // stateful stores it removes the backing record.
+      if (
+        error instanceof AccessTokenError &&
```
kishore7snehil: AGENTS.md suggests catching by `error.code` rather than `instanceof`. This uses `error instanceof AccessTokenError && error.code === SESSION_EXPIRED`.

EXAMPLES.md
```diff
@@ -3594,6 +3595,27 @@ To use Back-Channel Logout, you will need to provide a session store implementat
...
+try {
+  const { token } = await auth0.getAccessToken();
+  // use token...
+} catch (err) {
+  if (err instanceof AccessTokenError && err.code === AccessTokenErrorCode.SESSION_EXPIRED) {
+    // IdP session ceiling reached — the user must log in again
+    redirect("/auth/login");
+  }
+  throw err;
+}
```
kishore7snehil, line 4000: We should close the ts fence (```) here

EXAMPLES.md
```diff
@@ -3594,6 +3595,27 @@ To use Back-Channel Logout, you will need to provide a session store implementat
 
 A `LogoutToken` object will be passed as the parameter to `deleteByLogoutToken` which will contain either a `sid` claim, a `sub` claim, or both.
 
+## Session Expiry from the Upstream IdP
+
+For enterprise connections, the upstream identity provider can cap how long a user's session lives. When the connection is configured to honor it, Auth0 includes a `session_expiry` claim in the ID token, and the SDK enforces this ceiling automatically. Once it is reached, `getSession()` and `getUser()` return `null` (same as a logged-out user), and `getAccessToken()` throws an `AccessTokenError` with code `session_expired`.
```
kishore7snehil: Since `session_expiry` is set by a customer Post-Login Action, this guide should point developers to where that's documented rather than embedding the Action snippet here  so a link is the right call once it's published (placeholder).

Also, we should add a note that `session_expiry` is interpreted as **seconds** since epoch. A common slip is emitting milliseconds (`getTime()` without `/1000`); the SDK rejects implausibly large values (`>= 10_000_000_000`) as malformed and treats them as "no ceiling," so a ms value silently disables enforcement rather than expiring ~55,000 years out. Always divide by 1000.

kishore7snehil: Three things which can be added here as a note that : 

- once this feature is enabled, `getSession()` and the `useUser()` profile read can start returning null for a user who was previously logged in, once the ceiling passes. Apps that assumed a session always exists after login should add a null check and route back to login. A short "Upgrading" callout here would save existing integrators a surprise.
- the 30s negative leeway- the session is treated as expired slightly *before* the wall-clock ceiling to absorb clock skew
- the original ceiling is preserved across refreshes - refreshing an access token doesn't extend the upstream session. Sets correct expectations for anyone reasoning about exact timing.

README.md
```diff
@@ -485,6 +485,12 @@ The SDK mounts 16 routes:
 > [!IMPORTANT]  
 > The `/auth/access-token` route is enabled by default, but is only necessary when the access token is needed on the client-side. If this isn't something you need, you can disable this endpoint by setting `enableAccessTokenEndpoint` to `false`.
 
+## Session Expiry from the Upstream IdP
+
+For enterprise connections, the upstream identity provider can cap how long a user's session lives. When the connection is configured to honor it, Auth0 includes a `session_expiry` claim in the ID token, and the SDK enforces this ceiling on every session read. Once it is reached, `getSession()` and `getUser()` return `null`, and `getAccessToken()` throws an `AccessTokenError` with code `session_expired`. If the asserted ceiling is already in the past at login, the callback returns a `session_expired` error response instead of persisting an already-expired session.
```
kishore7snehil: There's no `getUser()` method in this SDK , the user surface is the `useUser()`.

kishore7snehil (changes requested):

README.md
Piyush-85: Fixed. Updated to `useUser()` in both README.md and EXAMPLES.md.

EXAMPLES.md
Piyush-85: Both addressed:
- Added a link to the Auth0 Actions documentation (placeholder replaced with `https://auth0.com/docs/customize/actions`).
- Added an explicit note that the value must be a **Unix timestamp in seconds**, with `Math.floor(Date.now() / 1000)` as the correct form. Also noted that implausibly large values (`>= 10_000_000_000`) are rejected by the SDK and treated as "no ceiling" — so a millisecond value silently disables enforcement rather than creating a far-future expiry.

Piyush-85: All three added to the `> [!NOTE]` callout in EXAMPLES.md:

EXAMPLES.md
Piyush-85: Done

src/utils/session-helpers.ts
Piyush-85: Fixed. The guard is:

```ts
const sessionExpiresAt =
  typeof rawExpiry === "number" && rawExpiry < 10_000_000_000
    ? rawExpiry
    : undefined;
```

This rejects millisecond timestamps, `NaN`, `Infinity`, and any non-number type. Tests added for all these cases in `session-expiry.test.ts`.

src/utils/session-helpers.ts
Piyush-85: The two helpers guard against different failure modes:

- **`isSessionCeilingReached`** subtracts leeway from the ceiling (`now >= ceiling - 30`) to absorb **clock skew at enforcement time** — the session is treated as expired 30s before the raw ceiling so a request arriving slightly late is still handled correctly.

- **`isSessionCeilingInPast`** adds leeway to the reference (`ceiling <= iat + 30`) to absorb **clock skew at login** — a ceiling within 30s of `iat` would pass the login guard but immediately fire `isSessionCeilingReached` on the first request anyway. Rejecting it at login is cleaner.

The 60s overlap you describe (ceiling between `iat+31` and `iat+60`) only affects sessions shorter than 60s — an extremely short-lived grant that the IdP is asserting should expire almost immediately. Rejecting it cleanly at login is the right call rather than letting the user through and kicking them out 20 seconds later mid-page. This is a deliberate strict-over-permissive choice, consistent with the Python SDK.

src/server/auth-client.ts
Piyush-85: Fixed. The `instanceof` guard was removed. The check is now:

```ts
if (error.code === AccessTokenErrorCode.SESSION_EXPIRED) {
  await this.sessionStore.delete(req.cookies, errorRes.cookies);
}
```

src/server/auth-client.ts
```diff
@@ -1800,6 +1842,18 @@ export class AuthClient {
...
+        error.code === AccessTokenErrorCode.SESSION_EXPIRED
+      ) {
```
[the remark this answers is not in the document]
Piyush-85: fixed

src/server/auth-client.ts
Piyush-85: Fixed by adding a skipCeilingCheck option to getSessionWithDomainCheck — this skips only the ceiling block while MCD domain validation still runs. 
getAccessTokenForConnection uses { skipCeilingCheck: true } so connection tokens remain accessible after the ceiling passes.

src/server/auth-client.ts
```diff
@@ -2645,6 +2711,20 @@ export class AuthClient {
...
     }
 
+    // IPSIE: treat as no session once the upstream IdP-asserted ceiling passes.
+    // Returns the same shape as "no session" so existing redirect-to-login paths
+    // fire transparently. For stateful stores, delete the backing record so the
+    // store stays clean; stateless sessions are self-contained in the cookie and
+    // will be orphaned (ceiling check fires on every subsequent read).
+    if (isSessionCeilingReached(session.internal?.sessionExpiresAt)) {
+      void this.sessionStore.deleteByReqCookies(cookies);
+      return {
```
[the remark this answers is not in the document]
Piyush-85: fixed. The call is now:

```ts
this.sessionStore.deleteByReqCookies(cookies).catch(() => undefined);
```

src/server/auth-client.ts
Piyush-85: Fixed. The popup branch now calls `await this.onCallback(expiredError, onCallbackCtx, null)` before returning the postMessage error response, making it consistent with the redirect branch.

src/server/session/stateful-session-store.ts
Piyush-85: Fixed. `deleteByReqCookies` now mirrors the `get()` fallback: it attempts `decrypt` first, and on `ERR_JWE_INVALID` falls back to `verifySigned` to extract the session ID from a legacy v3 cookie. Read and delete are now consistent.

src/server/session/abstract-session-store.ts
Piyush-85: Fixed. `deleteByReqCookies` is now a concrete no-op on the abstract base class:

src/server/client.ts
```diff
@@ -1109,10 +1109,20 @@ export class Auth0Client {
...
+      | import("./cookies.js").ReadonlyRequestCookies;
+    if (normalizedReq) {
+      reqCookies =
+        normalizedReq instanceof NextRequest
+          ? normalizedReq.cookies
+          : this.createRequestCookies(normalizedReq);
+    } else {
+      reqCookies = await cookies();
+    }
+    const session = await authClient.getSessionWithoutCeilingCheck(reqCookies);
```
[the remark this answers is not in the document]
Piyush-85: FIxed

src/server/auth-client.ts
```diff
@@ -2799,6 +2867,25 @@ export class AuthClient {
...
+    // Returns the same shape as "no session" so existing redirect-to-login paths
+    // fire transparently. For stateful stores, delete the backing record so the
+    // store stays clean; stateless sessions are self-contained in the cookie and
+    // will be orphaned (ceiling check fires on every subsequent read).
+    // skipCeilingCheck is used by getAccessTokenForConnection — connection tokens
+    // follow the upstream IdP's own TTLs, not the IPSIE session ceiling.
+    if (
+      !skipCeilingCheck &&
+      isSessionCeilingReached(session.internal?.sessionExpiresAt)
+    ) {
```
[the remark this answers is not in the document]
Piyush-85: The current behaviour is consistent with every other SDK path — when the ceiling fires, getSessionWithDomainCheck returns null and callers get MISSING_SESSION, same as natural session expiry. SESSION_EXPIRED is only surfaced from getTokenSet when a caller already holds a SessionData. Making 
 handleAccessToken the exception would create inconsistent error semantics across paths with no practical gain for integrators.

src/utils/session-helpers.ts
```diff
@@ -107,6 +137,15 @@ export function buildSessionFromCallback(
   oidcRes: oauth.TokenEndpointResponse,
   transactionState: TransactionState
 ): SessionData {
+  // Reject millisecond timestamps (13+ digits). A common Post-Login Action
+  // mistake is using Date.now() instead of Math.floor(Date.now() / 1000),
+  // which would stamp a year-57000 ceiling that never fires.
+  const rawExpiry = idTokenClaims.session_expiry;
+  const sessionExpiresAt =
+    typeof rawExpiry === "number" && rawExpiry < 10_000_000_000
```
kishore7snehil: The cap here rejects ms / NaN / Infinity / non-number and falls open to "no ceiling," but `0` and negative values slip through `typeof === "number" && rawExpiry < 10_000_000_000` is true for both. They get stamped as the ceiling, and then `isSessionCeilingInPast` reads them as already-lapsed and rejects the login.

Do you think we should lower-bound the guard too so zero and negatives fall open like the rest? Also, if you take the call to do so, we can add few tests related to it as well.

kishore7snehil (changes requested):

src/utils/session-helpers.ts
Piyush-85: Yes, missed it.
Added rawExpiry > 0 to the guard so zero and negatives fall open like the rest. Also added test cases for both 0 and -3600.

kishore7snehil (approved):
LGTM!

docs/ipsie-session-expiry-flow-impact.md
```diff
@@ -0,0 +1,68 @@
...
+3. **On every token fetch** (`getTokenSet`): if ceiling reached, returns `AccessTokenError(SESSION_EXPIRED)` immediately — Auth0's token endpoint is never called.
+4. **`getAccessTokenForConnection`** explicitly bypasses the ceiling via `getSessionWithoutCeilingCheck`.
+
+The 30-second leeway is applied to absorb clock skew between the SDK server and the IdP.
+
+---
+
+## Full flow table
+
+| # | Flow | Ceiling fires? | Why / Why not | Code location |
```
kishore7snehil: The YES rows look like they overreach the "Auth Code only" line, but they don't, a ceiling only gets *set* on an Auth Code login and it gets *checked* on every session read, which is why `getSession`, middleware, and the guards all fire.

Worth a one-line note up top splitting "sets the ceiling" from "enforces it" and rows #24 and #29 read like a non-auth-code flow stamps one.

kishore7snehil (changes requested):

docs/ipsie-session-expiry-flow-impact.md
Piyush-85: This file was not to be commited it was for internally testing flows deleted it.

kishore7snehil (dismissed):
LGTM!

Piyush-85:
@coderabbitai review

tusharpandey13 (approved):

PR #2722, opened 2026-06-26
IPSIE implementation - example

Piyush-85:

  ### 📋 Changes

  Adds a new `examples/with-ipsie-session-expiry` Next.js app that demonstrates end-to-end enforcement of the IPSIE `session_expiry` ceiling introduced in parent PR (#2708).

  The example is a single-page dashboard with nine cards, each exercising a different SDK code path so testers can observe exactly when and how the ceiling fires across every affected flow:

  - **Ceiling countdown** — live timer showing time until enforcement (ceiling minus 30s leeway), counts to zero when the SDK starts treating the session as
  expired
  - **Middleware rolling session** — shows `session.internal.createdAt` to confirm it stays constant across reloads and stops rolling after the ceiling
  - **`getSession()`** — server read via a Route Handler; returns null after the ceiling
  - **`useUser()`** — client-side hook; reflects logged-out state after the ceiling
  - **`getAccessToken()`** — calls `/auth/access-token`; surfaces `session_expired` error after the ceiling with a human-readable message distinguishing
  ceiling expiry from a genuinely missing session
  - **`withApiAuthRequired()`** — protected Route Handler returning 401 after the ceiling
  - **`updateSession()`** — session write; throws unauthenticated error after the ceiling
  - **`withPageAuthRequired()`** — protected page; redirects to login after the ceiling
  - **ID token claims** — raw JWT claims panel confirming `session_expiry` is present after the Post-Login Action is installed
  - **ID token claims** — raw JWT claims panel confirming `session_expiry` is present after the Post-Login Action is installed

  The README includes setup instructions (Post-Login Action snippet, skip-consent toggle), a testing table with expected output at each phase (before ceiling, approaching, after), and a flow-to-file mapping for anyone reading the code.

  ### 📎 References

  - #2708 — IPSIE `session_expiry` ceiling enforcement (parent PR)

## Summary by CodeRabbit

* **New Features**
  * Added an end-to-end example and UI for session expiry behavior, including session status, countdown, token checks, protected access, and session update actions.
  * Documented how upstream identity providers can set a session lifetime ceiling and how the app responds when it is reached.

* **Bug Fixes**
  * Sessions now expire locally when the upstream ceiling is reached, with token retrieval and protected views reflecting the expired state.
  * Improved handling for sessions that are already expired at login, preventing invalid access from being created.

### DIFF
--- a/examples/with-ipsie-session-expiry/README.md
+++ b/examples/with-ipsie-session-expiry/README.md
@@ -1,0 +1,134 @@
+# with-ipsie-session-expiry
+
+Tests IPSIE `session_expiry` ceiling enforcement end-to-end across all affected SDK paths.
+
+When Auth0 is configured to emit a `session_expiry` claim in the ID token, the SDK treats it as a hard, non-extendable ceiling on the local session. This example exercises every flow where ceiling enforcement fires.
+
+---
+
+## Setup
+
+### 1. Auth0 tenant
+
+#### Add the Post-Login Action
+
+1. Go to **Auth0 Dashboard → Actions → Library → Create Action → Build from scratch**
+2. Name it `IPSIE Short Ceiling`, trigger = **Post-login**
+3. Paste:
+
+   ```js
+   exports.onExecutePostLogin = async (event, api) => {
+     // 120-second ceiling — adjust to taste for faster testing
+     api.idToken.setCustomClaim("session_expiry", Math.floor(Date.now() / 1000) + 120);
+   };
+   ```
+
+4. **Deploy** → go to **Actions → Flows → Login** → drag the action into the flow → **Apply**
+
+> The Action is not retroactive. You must log in **fresh** after binding it (use incognito or log out first).
+
+#### Skip consent (required if testing with a custom audience)
+
+If `AUTH0_AUDIENCE` is set, Auth0 will prompt for consent on `offline_access`. To suppress it:
+**Dashboard → Applications → your app → Advanced Settings → OAuth → Skip user consent for first-party apps → On → Save**
+
+---
+
+### 2. Running
+
+```bash
+cp .env.example .env.local
+# Fill in AUTH0_DOMAIN, AUTH0_CLIENT_ID, AUTH0_CLIENT_SECRET, AUTH0_SECRET
+# AUTH0_AUDIENCE is optional — set to an API identifier to receive a JWT access token
+
+pnpm install
+pnpm dev
+```
+
+Open [http://localhost:3000](http://localhost:3000) and log in with incognito.
+
+---
+
+## What to test and what to look for
+
+The page has 9 cards, each targeting a specific SDK flow. The ceiling is set to **120 seconds**. The SDK applies a **30-second leeway**, so enforcement starts at **T+90s** (ceiling − leeway).
+
+---
+
+### Before the ceiling (T+0 to T+89s)
+
+All cards should show healthy state.
+
+| Card | Expected |
+|---|---|
+| IPSIE Session Ceiling | Countdown ticking down, green **Active** badge |
+| Middleware — Rolling Session | `createdAt` value is stable; page loads normally on refresh |
+| getSession() | `{ "session": { ... }, "expired": false }` |
+| useUser() | User object shown |
+| getAccessToken() — Browser | Access token displayed |
+| withApiAuthRequired() | `{ "status": 200, "body": { "protected": true } }` |
+| updateSession() | `{ "success": true, "updatedAt": "..." }` |
+| withPageAuthRequired() | /protected page renders with your email |
+
+---
+
+### Approaching ceiling (T+90s to T+120s)
+
+The SDK starts treating the session as expired once within 30s of the ceiling.
+
+| Card | Expected |
+|---|---|
+| IPSIE Session Ceiling | Yellow **Expiring soon** badge at ≤30s remaining |
+
+---
+
+### After the ceiling (T+90s onwards, enforcement active)
+
+Work through each card **without refreshing the main page** (refreshing redirects to login — do the manual checks first). Use the buttons to test each flow one by one.
+
+| Card | What to click/do | Expected result |
+|---|---|---|
+| **getSession()** | Click **Check Session** | `{ "session": null, "expired": true }` — `getSession()` returns null, same as logged-out |
+| **useUser()** | Observe (auto-polls) or reload the component | `user: undefined (logged-out state)` — hook sees 401 from `/auth/profile` |
+| **getAccessToken() — Browser** | Click **Get Access Token** | Red error box: `session_expired — the IdP session ceiling has been reached` |
+| **withApiAuthRequired()** | Click **Call Protected API** | `{ "status": 401 }` — guard returns 401 before the handler runs |
+| **updateSession()** | Click **Update Session** | `{ "success": false, "error": "No active session..." }` — session is null, update blocked |
+| **withPageAuthRequired()** | Open [/protected](http://localhost:3000/protected) in same tab | Redirected to `/auth/login` — page guard sees null session |
+| **Middleware — Rolling Session** | Refresh the main page | Redirected to `/auth/login` — middleware sees null session, rolling is skipped |
+
+> After the last step (refreshing the main page), you will be at the login page. That is correct — `getSession()` in the root page returns null and `redirect("/auth/login")` fires.
+
+---
+
+### At-login rejection (separate test run)
+
+This tests the callback path that rejects a session whose ceiling is already in the past at login time.
+
+1. Change the Post-Login Action to emit a **past** ceiling:
+
+   ```js
+   exports.onExecutePostLogin = async (event, api) => {
+     api.idToken.setCustomClaim("session_expiry", Math.floor(Date.now() / 1000) - 60);
+   };
+   ```
+
+2. **Deploy** the updated Action.
+3. Log out, then log in again (incognito).
+4. **Expected**: The callback returns a `session_expired` error — you land on an error page or the login page, never on the home page. No session is stored.
+
+Restore the Action to `+ 120` when done.
+
+---
+
+## SDK flows exercised
+
+| # | Flow | Card / endpoint | Ceiling fires at |
+|---|---|---|---|
+| 1 | `getSession()` | getSession card → `/api/check-session` | `getSessionWithDomainCheck` (line 2874) |
+| 2 | `useUser()` browser hook | useUser card → `/auth/profile` | `handleProfile` → `getSessionWithDomainCheck` |
+| 3 | `getAccessToken()` browser | getAccessToken card → `/auth/access-token` | `handleAccessToken` → `getTokenSet` (line 2236) |
+| 4 | `withApiAuthRequired()` | withApiAuthRequired card → `/api/protected` | `getSessionWithDomainCheck` (line 2874) |
+| 5 | `updateSession()` | updateSession card → `/api/update-session` | `getSession()` returns null before update |
+| 6 | `withPageAuthRequired()` | `/protected` page | `getSessionWithDomainCheck` (line 2874) |
+| 7 | Middleware rolling session | page refresh | `getSessionWithDomainCheck` (line 2874) |
+| 8 | `handleCallback` redirect | login flow (at-login rejection test) | `isSessionCeilingInPast` (line 1454) |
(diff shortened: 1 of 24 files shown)


examples/with-ipsie-session-expiry/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+If `AUTH0_AUDIENCE` is set, Auth0 will prompt for consent on `offline_access`. To suppress it:
+**Dashboard → Applications → your app → Advanced Settings → OAuth → Skip user consent for first-party apps → On → Save**
+
+---
+
+### 2. Running
+
+```bash
+cp .env.example .env.local
```
kishore7snehil, line 40: The setup step runs `cp .env.example .env.local`, but there's no `.env.example` in the example ,anyone following the README will hit a "No such file" on the first command.

Could we add a `.env.example` with the four vars the README lists (`AUTH0_DOMAIN`, `AUTH0_CLIENT_ID`, `AUTH0_CLIENT_SECRET`, `AUTH0_SECRET`) plus `APP_BASE_URL`, so the copy works out of the box?

examples/with-ipsie-session-expiry/app/ceiling-countdown.tsx
```diff
@@ -0,0 +1,57 @@
...
+  );
+
+  useEffect(() => {
+    const interval = setInterval(() => {
+      setRemaining(enforcedAt - Math.floor(Date.now() / 1000));
+    }, 1000);
+    return () => clearInterval(interval);
+  }, [enforcedAt]);
+
+  const status = remaining <= 0 ? "expired" : "active";
```
kishore7snehil: The README walkthrough says the ceiling badge turns yellow with "Expiring soon" at ≤30s remaining, but the component only has two states here, green "Active" and red "Expired" ,and `.badge-yellow` in the CSS is never used.

Shall we either add the ≤30s yellow state the README describes, or trim that row from the README so the two match?

examples/with-ipsie-session-expiry/app/ceiling-countdown.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { useEffect, useState } from "react";
+
+// The SDK enforces the ceiling 30s early (leeway for clock skew).
+// Show time until enforcement, not time until the raw ceiling timestamp.
+const LEEWAY = 30;
+
+export function CeilingCountdown({ ceiling }: { ceiling: number }) {
+  const enforcedAt = ceiling - LEEWAY;
+
+  const [remaining, setRemaining] = useState(
```
kishore7snehil: `remaining` is seeded from `Date.now()` in the `useState` initializer, which runs on both the server render and the client hydration, the two clocks won't agree, so React will warn about a hydration mismatch on this text node.

We could seed it to a stable value (e.g. `enforcedAt`, or `null`) and compute the live remaining inside the `useEffect` instead, so the first client paint matches the server.

examples/with-ipsie-session-expiry/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+Restore the Action to `+ 120` when done.
+
+---
+
+## SDK flows exercised
+
+| # | Flow | Card / endpoint | Ceiling fires at |
+|---|---|---|---|
+| 1 | `getSession()` | getSession card → `/api/check-session` | `getSessionWithDomainCheck` (line 2874) |
```
kishore7snehil: Could we drop the line numbers and reference the method names alone?

examples/with-ipsie-session-expiry/package.json
```diff
@@ -0,0 +1,23 @@
...
+  },
+  "dependencies": {
+    "@auth0/nextjs-auth0": "file:../..",
+    "next": "16.2.6",
+    "react": "19.0.1",
+    "react-dom": "19.0.1"
+  },
+  "devDependencies": {
+    "@types/node": "^22.8.6",
+    "@types/react": "^18.3.12",
```
kishore7snehil: The app runs `react@19.0.1` but pins `@types/react`/`@types/react-dom` at `^18` , the v18 types don't match the v19 runtime, which can surface as spurious type errors for anyone opening this in an editor.

Not sure but shall we bump both to `^19` to match?

examples/with-ipsie-session-expiry/app/access-token-panel.tsx
```diff
@@ -0,0 +1,60 @@
...
+    try {
+      const t = await getAccessToken();
+      setToken(t);
+    } catch (err: any) {
+      // getAccessToken() throws MISSING_SESSION when getSession() returns null.
+      // If the ceiling has passed, the root cause is the IPSIE ceiling — not a
+      // genuinely missing session — so show a ceiling-specific message.
+      const isCeilingExpired =
+        err?.code === "session_expired" ||
+        (err?.code === "missing_session" &&
```
kishore7snehil: When the code gets `missing_session`, it infers "ceiling reached" if `Date.now()` is within leeway of the ceiling , so a genuinely logged-out user who happens to be past the ceiling time gets told "the IdP session ceiling has been reached" rather than "no session."

examples/with-ipsie-session-expiry/app/page.tsx
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  const { user, tokenSet, internal } = session;
+  const ceiling = typeof internal.sessionExpiresAt === "number" ? internal.sessionExpiresAt : null;
+  const createdAt = internal.createdAt;
+
+  let idTokenClaims: Record<string, unknown> | null = null;
+  if (tokenSet.idToken) {
+    try {
+      const payload = tokenSet.idToken.split(".")[1];
```
kishore7snehil: This decodes the ID token by hand , splitting on `.` and base64url-decoding the payload, to render the claims. The SDK already gives you the verified claims, so an example aimed at customers teaching this pattern risks them copying an unverified manual decode into real code.

Could we pull the claims from the session/SDK instead, so the example models the safe path?

examples/with-ipsie-session-expiry/app/api/protected/route.ts
```diff
@@ -0,0 +1,11 @@
+import { NextRequest, NextResponse } from "next/server";
+
+import { auth0 } from "@/lib/auth0";
+
+const protectedHandler = auth0.withApiAuthRequired(async (_req: Request | NextRequest) => {
+  return NextResponse.json({ protected: true, message: "You have an active session." });
+});
+
+export async function GET(req: NextRequest): Promise<NextResponse> {
+  return protectedHandler(req, {}) as Promise<NextResponse>;
```
kishore7snehil: The handler is called with `{}` for the route context and the result is cast with `as Promise<NextResponse>`. The empty object stands in for the params context, and the cast papers over the union return type ,both will read as "what's going on here?" to someone using this as a template.

examples/with-ipsie-session-expiry/app/api/update-session/route.ts
```diff
@@ -0,0 +1,27 @@
...
+  try {
+    await auth0.updateSession({
+      ...session,
+      user: { ...session.user, updatedAt: new Date().toISOString() }
+    });
+    return NextResponse.json({ success: true, updatedAt: new Date().toISOString() });
+  } catch (err: any) {
+    return NextResponse.json(
+      { success: false, error: err.message },
+      { status: 409 }
```
kishore7snehil: The catch returns `409 Conflict` for any failure from `updateSession`, but the errors that land here aren't conflicts.

examples/with-ipsie-session-expiry/app/session-check-panel.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+import { useState } from "react";
+
+export function SessionCheckPanel() {
+  const [result, setResult] = useState<object | null>(null);
+  const [loading, setLoading] = useState(false);
+
+  async function handleCheck() {
+    setLoading(true);
+    const res = await fetch("/api/check-session");
```
kishore7snehil: The `fetch` and `res.json()` here aren't wrapped in a try/catch, so if either throws, `loading` is left stuck on `true` and the button stays disabled. The other panel guards the parse with `.json().catch(...)`; this one doesn't.

kishore7snehil (changes requested):

examples/with-ipsie-session-expiry/README.md
Piyush-85: `.env.example` was already present in the repo but was being silently excluded by the `*.env.*` pattern in `.gitignore`, so it never made it into the PR diff. 
Added `!**/.env.example` to `.gitignore` to unblock it — this also covers any future examples that add a template file.

examples/with-ipsie-session-expiry/app/ceiling-countdown.tsx
Piyush-85: Fixed — added the `warning` state at `remaining <= 30` with `badge-yellow` and the "Expiring soon" label. The component now has three states: green Active → yellow Expiring soon → red Expired, matching the README walkthrough table.

examples/with-ipsie-session-expiry/app/ceiling-countdown.tsx
Piyush-85: Fixed — seeded state to `null` and moved the initial `tick()` call inside the `useEffect`, so the server render and first client paint both see `null` (rendered as `…`). The live countdown starts on mount, after hydration.

examples/with-ipsie-session-expiry/README.md
Piyush-85: Done — dropped all line numbers from the table. Method names alone are stable across refactors.

examples/with-ipsie-session-expiry/package.json
Piyush-85: Bumped both `@types/react` and `@types/react-dom` to `^19` to match the runtime.

examples/with-ipsie-session-expiry/app/access-token-panel.tsx
Piyush-85: Agreed. Removed it entirely. The panel now only shows the ceiling-specific message when the error code is exactly `session_expired`. A genuinely logged-out user gets the raw error message.

examples/with-ipsie-session-expiry/app/page.tsx
Piyush-85: Replaced the manual split/decode with `session.user`, which already contains the SDK-verified claims including `session_expiry`. The card description is updated to reflect that the claims come from the session, not a raw token parse.

examples/with-ipsie-session-expiry/app/api/protected/route.ts
Piyush-85: Separated the handler definition from the `GET` export so the wrapper is at least readable. The `as Promise<Response>` cast is still needed — `withApiAuthRequired` has an overloaded signature where the inferred return comes through as `unknown` rather than `Promise<Response> | Response`. Added a comment explaining this is a type gap in the SDK signature, not papering over a real runtime issue.

examples/with-ipsie-session-expiry/app/api/update-session/route.ts
Piyush-85: Changed to 500. The errors that reach that catch block are unexpected failures from `updateSession`, not conflicts.

examples/with-ipsie-session-expiry/app/session-check-panel.tsx
Piyush-85: Added `try/finally` so `loading` is guaranteed to reset even if `fetch` or `.json()` throws. Also guarded `.json()` with `.catch(() => ({ error: "Invalid response" }))` consistent with the other panels.

examples/with-ipsie-session-expiry/app/page.tsx
```diff
@@ -0,0 +1,160 @@
...
+          — it renders normally before the ceiling. After the ceiling, opening
+          that URL redirects to <code>/auth/login</code> instead.
+        </p>
+      </div>
+
+      {/* ── 10. ID token claims ──────────────────────────────── */}
+      <div className="card">
+        <h2>ID Token Claims</h2>
+        <p style={{ fontSize: "0.875rem", color: "#6b7280", marginBottom: "0.75rem" }}>
+          Verified claims from the ID token, sourced from <code>session.user</code>.{" "}
```
kishore7snehil: `session.user` here is the filtered claim set, and `session_expiry` isn't one of the default claims that survive `filterDefaultIdTokenClaims`, so it gets stripped before the session is stored. This card's copy says `session_expiry` "should appear here," but it won't as the one claim the card exists to show is the one that's filtered out.

The countdown card already reads it from `internal.sessionExpiresAt`, Can we show it in this card from the same place, rather than from `session.user`?

kishore7snehil (changes requested):

kishore7snehil (approved):
LGTM!

Issue #2655, opened 2026-05-07
Redirect users using the state parameter


rybon:
### Checklist

- [x] I have looked into the [Readme](https://github.com/auth0/nextjs-auth0#readme), [Examples](https://github.com/auth0/nextjs-auth0/blob/main/EXAMPLES.md), and [FAQ](https://github.com/auth0/nextjs-auth0/blob/main/FAQ.md) and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://auth0.github.io/nextjs-auth0/) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/nextjs-auth0/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

How do I implement this functionality? Currently, the URL a user came from is lost during the authentication flow and they land on the default `returnTo` value, which is `/`. That is not a good user experience. Apparently, the `state` parameter can be used to remediate this, but the documentation of this SDK does not say how.

I want to achieve what the docs state here:

https://auth0.com/docs/secure/attack-protection/state-parameters#redirect-users

### Describe the ideal solution

A concrete end-to-end example where this SDK is used to redirect users back to the URL they wanted to visit, before being asked to authenticate. These are dynamic URLs.

Piyush-85:
@rybon Thanks for raising this! 
The SDK already supports this via the `returnTo` query parameter on the login URL — it's stored in the transaction state and the user is redirected there after the callback. The docs just don't show how to wire it up in middleware.

**Middleware example** (the most common case):

```ts
// middleware.ts
export async function middleware(request: NextRequest) {
  const authRes = await auth0.middleware(request);

  if (request.nextUrl.pathname.startsWith("/auth")) {
    return authRes;
  }

  const session = await auth0.getSession(request);

  if (!session) {
    const loginUrl = new URL("/auth/login", request.nextUrl.origin);
    loginUrl.searchParams.set(
      "returnTo",
      request.nextUrl.pathname + request.nextUrl.search
    );
    return NextResponse.redirect(loginUrl);
  }

  return authRes;
}
```

[...]

PR #2728, opened 2026-06-29
fix: forward state param to OIDC end-session URL in handleLogout

### LINKED ISSUE #2717: The state parameter does not work with the OIDC logout endpoint.
### Checklist

How can I include the state parameter in the logout URL?

After checking the code, it seems that the state parameter is not set in createOIDCLogoutResponse.

According to the Auth0 documentation, the OIDC logout endpoint supports the state parameter. The documentation says that state is an optional opaque value that Auth0 includes when redirecting to post_logout_redirect_uri.

I tried adding it to returnTo based on this issue, but it did not work:
https://github.com/auth0/nextjs-auth0/issues/2655

Is there a supported way to include the state parameter in the logout URL?

I think createOIDCLogoutResponse could set the state parameter based on the value passed to returnTo.

Since the Auth0 OIDC logout endpoint supports the state parameter, it would be useful if handleLogout allowed users to include state in returnTo, and then passed it through to the generated OIDC logout URL.

### Alternatives and current workarounds

_No response_

### Additional context

_No response_

Piyush-85:

  ### 📋 Changes

  The OIDC RP-Initiated Logout endpoint (`/oidc/logout`) supports an optional `state` parameter — an opaque value the authorization server echoes back as a query param when redirecting to `post_logout_redirect_uri`. The SDK was not forwarding this parameter, leaving users with no supported way to round-trip context (e.g. where to redirect after logout) through the logout flow.

 This is a real gap because `post_logout_redirect_uri` must exactly match a registered allowed logout URL on the Auth0 tenant — so embedding context as a query param there gets rejected. `state` is the correct OIDC mechanism for this use case.

  `handleLogout` now reads a `state` query param from the incoming logout request and forwards it to the OIDC end-session URL. Users can call
  `/auth/logout?returnTo=https://example.com/done&state=someOpaqueValue` and Auth0 will redirect to
  `https://example.com/done?state=someOpaqueValue` after logout.

  This only applies to the OIDC logout path (`logoutStrategy: "oidc"` and `logoutStrategy: "auto"` when the OIDC endpoint is available). The `/v2/logout` endpoint does not support `state` and is unchanged.

  This is a non-breaking, purely additive change — callers that do not pass `state` see no difference in behavior.

  ### 📎 References

  - Fixes #2717
  - Related: #2655

  ### 🎯 Testing

  Four unit tests added to `logout-strategy.flow.test.ts` covering:

  - `state` is forwarded to the OIDC logout URL when provided
  - `state` is absent from the OIDC logout URL when not provided (no regression)
  - `state` is not forwarded on the `v2` logout path
  - `state` is forwarded when using the `auto` strategy and the OIDC endpoint is available

## Summary by CodeRabbit

- **Bug Fixes**
  - Logout redirects now preserve an optional `state` value when using OIDC logout flows.
  - `state` is only included when provided, and it is not passed to non-OIDC logout endpoints.
  - Auto-selected logout behavior now correctly forwards `state` for OIDC-based redirects.

### DIFF
--- a/src/server/auth-client.ts
+++ b/src/server/auth-client.ts
@@ -954,6 +954,7 @@ export class AuthClient {
 
     const appBaseUrl = resolveAppBaseUrl(this.appBaseUrl, req);
     const returnTo = req.nextUrl.searchParams.get("returnTo") || appBaseUrl;
+    const logoutState = req.nextUrl.searchParams.get("state");
     const federated = req.nextUrl.searchParams.has("federated");
 
     const createV2LogoutResponse = (): NextResponse => {
@@ -972,6 +973,10 @@ export class AuthClient {
       const url = new URL(authorizationServerMetadata.end_session_endpoint!);
       url.searchParams.set("client_id", this.clientMetadata.client_id);
       url.searchParams.set("post_logout_redirect_uri", returnTo);
+
+      if (logoutState) {
+        url.searchParams.set("state", logoutState);
+      }
 
       if (session?.internal.sid) {
         url.searchParams.set("logout_hint", session.internal.sid);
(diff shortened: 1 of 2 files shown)


tusharpandey13 (approved):
lgtm

PR #2734, opened 2026-07-06
feat: add session transfer token support to CTE

Piyush-85:

  ### 📋 Changes

Implements CTE Phase 2 — Session Transfer Token (STT). An authenticated agent can now request a short-lived, one-shot token that lets them establish a web session **as a customer** in a target app, without the customer's password.

  **New methods on `Auth0Client`:**
  - `requestSessionTransferToken(options)` — exchanges a subject token for an STT via `POST /oauth/token` with
  `audience=urn:{domain}:session_transfer`. The actor defaults to the agent session's ID token (checked for expiry); an
  explicit `actor` override is also supported.
  - `buildSessionTransferRedirect(targetLoginUrl, result, opts?)` — pure URL builder returning a `NextResponse.redirect` with `?session_transfer_token=<stt>` appended.

  **New types exported from `@auth0/nextjs-auth0/server` and `@auth0/nextjs-auth0/types`:**
  - `SessionTransferTokenOptions` / `SessionTransferTokenResult`
  - `TOKEN_TYPES` enum (`ID_TOKEN`, `ACCESS_TOKEN`, `SESSION_TRANSFER_TOKEN`)

  **New error codes on `CustomTokenExchangeErrorCode`:**
  - `ACTOR_UNAVAILABLE` — no usable actor token (client-side, before network call)
  - `SETACTOR_REQUIRED` — server requires an actor to be set via `api.authentication.setActor()`
  - `SESSION_TRANSFER_DISABLED` — tenant feature flag not enabled

  **Implementation notes:**
  - `oauth4webapi`'s `processGenericTokenEndpointResponse` rejects `token_type: "N_A"` (returned by the STT endpoint). The implementation bypasses it by calling `genericTokenEndpointRequest` directly and parsing the raw JSON body.
  - STT-specific pure functions are extracted to `src/utils/session-transfer-helpers.ts` to keep `auth-client.ts` manageable.
  - No middleware change needed — `handleLogin()` already forwards all query params to `/authorize`, so
  `session_transfer_token` passes through transparently on the target app side.

  ### 📎 References

  - CTE Phase 2 / Session Transfer Token spec (internal Confluence)

  ### 🎯 Testing

  - 36 new unit tests in `src/server/custom-token-exchange.test.ts` (section 5) covering result shape, actor resolution, wire
  parameters, STT not persisted to session, server error mapping, input validation, `buildSessionTransferRedirect`, and
  backward compatibility.
  - 23 unit tests in `src/utils/session-transfer-helpers.test.ts` covering all 4 helper functions.
  - All 1739 tests pass, TypeScript clean, zero new console warnings from the STT code paths.

## Summary by CodeRabbit

* **New Features**
  * Added Session Transfer Token (STT) support with new request and redirect helpers for one-shot session transfer.
  * Exposed STT option/result types, token-type constants, and STT-related OAuth error codes.
* **Bug Fixes**
  * Improved STT actor validation/resolution and safer handling of STT-specific token responses, including typed server error mapping.
* **Documentation**
  * Added STT documentation to the Custom Token Exchange guide, including flow, error behavior, and limitations.
* **Tests**
  * Added unit and end-to-end coverage for STT request/response contracts, redirect URL encoding, and query forwarding.

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -5077,10 +5077,10 @@ if (actor) {
 
 - **Server-side only**: `requestSessionTransferToken` requires `client_secret` and can only be called server-side (confidential-client RWAs; SPAs/native are out of scope).
 - **One-shot, ~60 s**: The STT is opaque and must never be decoded, cached, or reused. Redeem it immediately.
-- **Actor is mandatory and must be fresh**: The actor token (the agent's session ID token by default) must be an unexpired, asymmetrically-signed (RS256/PS256) JWT. If it is missing or expired the SDK fails with `ACTOR_UNAVAILABLE` — there is no automatic refresh, so refresh the agent session or pass a fresh explicit `actor`.
+- **Actor is mandatory and must be fresh**: The actor token (the agent's session ID token by default) must be an unexpired, asymmetrically-signed (RS256/PS256) JWT. If it is expired and a refresh token is present the SDK automatically refreshes it; otherwise it fails with `ACTOR_UNAVAILABLE`. Pass an explicit `actor` to use a non-session token.
 - **Non-localhost callback**: STT redemption rejects `localhost` redirect URIs — use a real domain (or a tunnel) for the target callback.
 - **Same tenant**: Both agent and target app must be on the same Auth0 tenant.
-- **No refresh, short-lived session**: The impersonation session cannot mint a refresh token and self-expires; re-run the flow to continue.
+- **No refresh token / no `offline_access`**: Auth0 silently drops `offline_access` from the STT scope — the impersonation session has no refresh token and self-expires according to the tenant's session lifetime settings. Pass `scope: "openid profile"` on the target login URL (or via `options.scope`) to make the intent explicit and avoid unexpected scope-negotiation warnings.
 - **No session modification**: The agent's own session is not modified; the STT is never written to `session.tokenSet`.
 
 ## Customizing Auth Handlers
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -182,6 +182,12 @@
   - [Token Type Requirements](#token-type-requirements)
   - [Limitations](#limitations)
   - [DPoP Support](#dpop-support)
+- [Session Transfer Token](#session-transfer-token)
+  - [Basic Usage](#basic-usage-1)
+  - [Actor Override](#actor-override)
+  - [With Organization](#with-organization-1)
+  - [Error Handling](#error-handling-6)
+  - [Limitations](#limitations-1)
 - [Customizing Auth Handlers](#customizing-auth-handlers)
   - [Run custom code before Auth Handlers](#run-custom-code-before-auth-handlers)
   - [Run code after callback](#run-code-after-callback)
@@ -4928,6 +4934,105 @@ const result = await auth0.customTokenExchange({
 
 // result.tokenType will be "DPoP"
 ```
+
+## Session Transfer Token
+
+Session Transfer Token (STT) is CTE Phase 2. An authenticated agent establishes a web session **as a customer** in a target app — without the customer's password. The agent app requests a one-shot STT for the customer, then redirects the agent's browser to the target app's login URL carrying the token. Auth0 validates the STT and creates an ephemeral session for the agent.
+
+> [!IMPORTANT]
+> Session Transfer requires the `session_transfer` feature flag to be enabled on your tenant. Both the agent app and target app must be on the same Auth0 tenant. Contact Auth0 support or raise an ESD to enable the flag.
+
+### Basic Usage
+
+```ts
+import { TOKEN_TYPES } from "@auth0/nextjs-auth0/server";
+
+// app/api/stt/route.ts — runs in the agent app
+export async function POST(req: NextRequest) {
+  const { subjectToken, subjectTokenType, targetLoginUrl } = await req.json();
+
+  // Request a one-shot STT for the customer
+  const result = await auth0.requestSessionTransferToken({
+    subjectToken,                  // customer's ID or access token
+    subjectTokenType,              // TOKEN_TYPES.ID_TOKEN or TOKEN_TYPES.ACCESS_TOKEN
+    reason: "Support investigation #1234"
+  });
+
+  // Redirect the agent's browser to the target app's login URL
+  return auth0.buildSessionTransferRedirect(targetLoginUrl, result);
+  // → 307 to https://target-app.example.com/auth/login?session_transfer_token=<stt>
+}
+```
+
+The target app needs no changes — `handleLogin()` already forwards all query params to `/authorize`, so `session_transfer_token` passes through automatically.
+
+### Actor Override
+
+The actor (the party making the request on the customer's behalf) defaults to the agent session's ID token. Pass `actor` explicitly to override:
+
+```ts
+import { TOKEN_TYPES } from "@auth0/nextjs-auth0/server";
+
+const result = await auth0.requestSessionTransferToken({
+  subjectToken: customerIdToken,
+  subjectTokenType: TOKEN_TYPES.ID_TOKEN,
+  actor: { token: agentIdToken, type: TOKEN_TYPES.ID_TOKEN }
+});
+```
+
+### With Organization
+
+```ts
+const result = await auth0.requestSessionTransferToken({
+  subjectToken: customerIdToken,
+  subjectTokenType: TOKEN_TYPES.ID_TOKEN,
+  organization: "org_abc123"
+});
+
+return auth0.buildSessionTransferRedirect(
+  "https://target-app.example.com/auth/login",
+  result,
+  { organization: "org_abc123" }
+);
+```
+
+### Error Handling
+
+```ts
+import {
+  CustomTokenExchangeError,
+  CustomTokenExchangeErrorCode
+} from "@auth0/nextjs-auth0/server";
+
+try {
+  const result = await auth0.requestSessionTransferToken({ ... });
+} catch (error) {
+  if (error instanceof CustomTokenExchangeError) {
+    switch (error.code) {
+      case CustomTokenExchangeErrorCode.ACTOR_UNAVAILABLE:
+        // No usable actor token — agent session has no ID token or it is expired
+        break;
+      case CustomTokenExchangeErrorCode.SETACTOR_REQUIRED:
+        // Tenant requires api.authentication.setActor() to be called in an Action
+        break;
+      case CustomTokenExchangeErrorCode.SESSION_TRANSFER_DISABLED:
+        // session_transfer feature flag not enabled on this tenant
+        break;
+      case CustomTokenExchangeErrorCode.EXCHANGE_FAILED:
+        // Generic server error — check error.cause for details
+        console.error("STT exchange failed:", error.cause);
+        break;
+    }
+  }
+}
+```
+
+### Limitations
+
+- **Server-side only**: `requestSessionTransferToken` requires `client_secret` and can only be called server-side.
+- **One-shot, ~60 s**: The STT expires quickly and must never be cached or reused.
+- **Same tenant**: Both agent and target app must be on the same Auth0 tenant.
+- **No session modification**: The agent's own session is not modified; no tokens are written to `session.tokenSet`.
 
 ## Customizing Auth Handlers
 
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -4943,15 +4943,11 @@ Session Transfer Token (STT) is CTE Release 2. An authenticated agent (for examp
 Almost all of the new SDK surface is on the **initiator** side (the app that requests the STT and builds the redirect): one method — `requestSessionTransferToken()` — plus a redirect helper — `buildSessionTransferRedirect()`. The **target** app barely changes: it does a standard OIDC Authorization Code login with `session_transfer_token` riding along in the `/authorize` query string.
 
 > [!IMPORTANT]
-<<<<<<< Updated upstream
-> Session Transfer is a **two-client flow** and requires the `cte_session_transfer_token` feature flag on your tenant (raise an ESD, or enable via layer0 for internal testing). Both clients must be on the same Auth0 tenant. Configure:
+> Session Transfer is a **two-client flow** and requires the `cte_session_transfer_token` feature flag on your tenant (contact Auth0 support to enable it). Both clients must be on the same Auth0 tenant. Configure:
 > - **Issuing (initiator) client** — `token_exchange.allow_any_profile_of_type: ["custom_authentication"]` and `session_transfer.can_create_session_transfer_token: true`.
 > - **Redeeming (target) client** — `session_transfer.delegation.allow_delegated_access: true`, `allowed_authentication_methods` must include `"query"` (the STT is redeemed as a query parameter), and `session_transfer.delegation.enforce_device_binding` (`"ip"` default or `"asn"`). Register a **non-localhost** callback — STT redemption rejects `localhost` redirect URIs.
 >
 > A CTE Action must validate the `subject_token`, call `setUserById(customer)`, and call `setActor(agent)` — an STT is only issued when an actor is set. These client settings are configured out of band via the Management API; they are not SDK code.
-=======
-> Session Transfer requires the `cte_session_transfer_token` feature flag to be enabled on your tenant. Both the agent app and target app must be on the same Auth0 tenant. Contact Auth0 support or raise an ESD to enable the flag.
->>>>>>> Stashed changes
 
 ### Basic Usage
 
@@ -5029,18 +5025,15 @@ try {
   if (error instanceof CustomTokenExchangeError) {
     switch (error.code) {
       case CustomTokenExchangeErrorCode.ACTOR_UNAVAILABLE:
-<<<<<<< Updated upstream
         // Raised client-side, before any network call: no explicit `actor` was
-        // passed and the agent session has no usable (unexpired) ID token.
-=======
-        // No usable actor token — agent session has no ID token or it is expired
+        // passed and the agent session has no usable (unexpired) ID token and no
+        // refresh token to recover with.
         break;
       case CustomTokenExchangeErrorCode.SETACTOR_REQUIRED:
         // Tenant requires api.authentication.setActor() to be called in an Action
         break;
       case CustomTokenExchangeErrorCode.SESSION_TRANSFER_DISABLED:
         // cte_session_transfer_token feature flag not enabled on this tenant
->>>>>>> Stashed changes
         break;
       case CustomTokenExchangeErrorCode.EXCHANGE_FAILED:
         // Server-side failure. Today the server returns a generic `invalid_request`
@@ -5077,6 +5070,7 @@ if (actor) {
 
 - **Server-side only**: `requestSessionTransferToken` requires `client_secret` and can only be called server-side (confidential-client RWAs; SPAs/native are out of scope).
 - **One-shot, ~60 s**: The STT is opaque and must never be decoded, cached, or reused. Redeem it immediately.
+- **Branch on `issued_token_type`, not `token_type`**: The response carries `token_type: "N_A"` (informational). The SDK already validates that `issued_token_type === "urn:auth0:params:oauth:token-type:session_transfer_token"` before returning; if it doesn't match the exchange throws `EXCHANGE_FAILED`. Use `TOKEN_TYPES.SESSION_TRANSFER_TOKEN` in any downstream checks — never branch on `token_type`.
 - **Actor is mandatory and must be fresh**: The actor token (the agent's session ID token by default) must be an unexpired, asymmetrically-signed (RS256/PS256) JWT. If it is expired and a refresh token is present the SDK automatically refreshes it; otherwise it fails with `ACTOR_UNAVAILABLE`. Pass an explicit `actor` to use a non-session token.
 - **Non-localhost callback**: STT redemption rejects `localhost` redirect URIs — use a real domain (or a tunnel) for the target callback.
 - **Same tenant**: Both agent and target app must be on the same Auth0 tenant.
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -186,7 +186,7 @@
   - [Basic Usage](#basic-usage-1)
   - [Actor Override](#actor-override)
   - [With Organization](#with-organization-1)
-  - [Error Handling](#error-handling-6)
+  - [Error Handling](#error-handling-5)
   - [Reading the `act` claim on the established session](#reading-the-act-claim-on-the-established-session)
   - [Limitations](#limitations-1)
 - [Customizing Auth Handlers](#customizing-auth-handlers)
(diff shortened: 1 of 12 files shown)


src/utils/session-transfer-helpers.ts
```diff
@@ -0,0 +1,172 @@
...
+      ),
+      null
+    ];
+  }
+
+  // Guard against expired ID tokens — the server validates exp on the actor_token
+  // and rejects with "Invalid actor token: 'exp' claim timestamp check failed".
+  try {
+    const { exp } = jose.decodeJwt(idToken);
+    if (typeof exp === "number" && exp < Math.floor(Date.now() / 1000)) {
```
kishore7snehil, line 71: When the session ID token is expired this returns `ACTOR_UNAVAILABLE` right away, but the SDK requirements say we should first try to refresh the agent's session ID token and only fail when there's no refresh token to do it with. Right now an agent who has a perfectly good refresh token still gets bounced, which turns a recoverable case into a hard failure.

The refresh machinery already exists on the client (`#refreshTokenSet`), and the server-python PoC does exactly this: refresh when the ID token is stale and a refresh token is present, fall back to `ACTOR_UNAVAILABLE` only when it can't. Can we route the expired-token path through a refresh here before giving up? Note this needs the session read/write, so the refresh likely has to live on the caller in `requestSessionTransferToken` rather than in this pure helper.

src/server/client.ts
```diff
@@ -1232,6 +1235,93 @@ export class Auth0Client {
...
+   * ```typescript
+   * const result = await auth0.requestSessionTransferToken({
+   *   subjectToken: mySubjectToken,
+   *   subjectTokenType: "urn:acme:subject",
+   *   reason: "Investigating ticket #1234"
+   * });
+   * return auth0.buildSessionTransferRedirect("https://app.example.com/auth/login", result);
+   * ```
+   */
+  async requestSessionTransferToken(
```
kishore7snehil, line 1332: This is the public entry point developers actually call, and it's the layer that does the tuple-to-throw conversion, sources the session with `getSession()`, and has the resolver-mode fallback for the redirect helper. All the STT tests exercise the inner `AuthClient` directly, so none of this wrapper behaviour is covered.

Can we add tests here for at least the throw-on-error path and the session sourcing? A bug in the tuple unwrap or the session read would sail through the current suite.

src/server/custom-token-exchange.test.ts
```diff
@@ -952,4 +954,788 @@ describe("Custom Token Exchange", () => {
...
+          result
+        );
+
+        const location = redirect.headers.get("location") ?? "";
+        const url = new URL(location);
+        expect(url.searchParams.has("organization")).toBe(false);
+      });
+    });
+
+    describe("5.8: backward compatibility — existing CTE tests unaffected", () => {
```
kishore7snehil, line 2020: The whole design leans on the target app forwarding `session_transfer_token` (and `organization`) to `/authorize` untouched through the existing login route, and the requirements call this out as a test every initiator SDK should have. I don't see one that actually drives a login and asserts the param lands on the authorize request.

Since "no middleware change needed" is a strong claim for this feature, shall we add a test that pins it so a future change to the login passthrough can't silently break STT?

EXAMPLES.md
```diff
@@ -4629,6 +4636,153 @@ const result = await auth0.customTokenExchange({
...
+
+if (actor) {
+  // e.g. show an "Impersonating — acting as <customer>" banner,
+  // and/or restrict destructive actions while impersonating.
+}
+```
+
+`act` carries the acting party's `sub` and optionally a few small custom fields the Action added (for example the agent's email or a reason). The `act` claim is **not** present on the STT exchange response — it only appears on the redeemed session's tokens.
+
+### Limitations
```
kishore7snehil, line 5089: The requirements ask the guide to tell developers where to set `online_access` for the target session, since some SDKs manage `scope` internally. On nextjs a developer can pass scope inline so it's less of a trap here than on the server SDKs, but a one-line note in this Limitations list would still save someone the surprise.

src/types/token-vault.ts
```diff
@@ -143,6 +168,81 @@ export interface ActClaim {
...
+
+  /**
+   * Scopes for the impersonation session tokens.
+   * Merged with SDK default scopes (openid, profile, email).
+   * Note: `offline_access` is silently suppressed by Auth0 for impersonation sessions.
+   */
+  scope?: string;
+
+  /**
+   * Explicit actor override. Omit to use the agent session's ID token (refreshed if expired).
```
kishore7snehil: This says the session ID token is "refreshed if expired", but the code doesn't refresh, and the method docstring on `requestSessionTransferToken` says the opposite ("no automatic refresh is performed"). Whichever way the refresh question lands, these two should tell the same story.

EXAMPLES.md
```diff
@@ -174,6 +174,13 @@
   - [Token Type Requirements](#token-type-requirements)
   - [Limitations](#limitations)
   - [DPoP Support](#dpop-support)
+- [Session Transfer Token](#session-transfer-token)
+  - [Basic Usage](#basic-usage-1)
+  - [Actor Override](#actor-override)
+  - [With Organization](#with-organization-1)
+  - [Error Handling](#error-handling-6)
```
kishore7snehil: This anchor is one too high. This is the sixth `### Error Handling` in the file, and GitHub's slugger suffixes the sixth occurrence with `-5`, so the link should be `#error-handling-5`.

src/server/auth-client.ts
```diff
@@ -3462,6 +3471,271 @@ export class AuthClient {
...
+        null
+      ];
+    }
+
+    // STT never decoded, never cached — treat as opaque handle
+    return [
+      null,
+      parseSessionTransferTokenResponse({
+        access_token: rawBody.access_token,
+        issued_token_type:
```
kishore7snehil: On the success path this defaults `issued_token_type` to the session-transfer URN when the field is missing, so a 200 response is assumed to be an STT without ever checking. The spec is firm that `issued_token_type` is the one field to branch on, precisely so a regular access-token response can't be mistaken for an STT. As written, a non-STT 200 (say a plain `Bearer` token) would come back as `sessionTransferToken` and then get placed into a redirect query string by `buildSessionTransferRedirect`.

Shall we confirm `rawBody.issued_token_type === TOKEN_TYPES.SESSION_TRANSFER_TOKEN` before returning success, and fail with `EXCHANGE_FAILED` when it doesn't match, rather than defaulting it?

src/utils/session-transfer-helpers.ts
```diff
@@ -0,0 +1,172 @@
...
+): string {
+  let url: URL;
+  try {
+    url = new URL(targetLoginUrl);
+  } catch {
+    throw new CustomTokenExchangeError(
+      CustomTokenExchangeErrorCode.EXCHANGE_FAILED,
+      `Invalid targetLoginUrl: "${targetLoginUrl}" is not an absolute URL. ` +
+        "Pass a trusted, app-controlled absolute login URL (e.g. https://app.example.com/auth/login)."
+    );
```
kishore7snehil, line 117: `new URL(targetLoginUrl)` confirms the value is an absolute URL but doesn't restrict the scheme, so a `javascript:` or `data:` value would parse and come back as the redirect target. `NextResponse.redirect` does its own scheme check so this isn't directly exploitable today, and the JSDoc already says the URL must be trusted.

That said, the requirements put URL hardening (https-only, origin checks) on this core helper rather than the callers, so an `http:`/`https:` allowlist here would be cheap insurance and keep the guard in the one place the spec expects it.

src/utils/session-transfer-helpers.ts
```diff
@@ -0,0 +1,172 @@
...
+  errorCode: string
+): CustomTokenExchangeErrorCode | null {
+  const code = errorCode.toLowerCase();
+  if (code === "setactor_required" || code.includes("setactor")) {
+    return CustomTokenExchangeErrorCode.SETACTOR_REQUIRED;
+  }
+  if (
+    code === "session_transfer_disabled" ||
+    code.includes("session_transfer")
+  ) {
```
kishore7snehil: The `code.includes("setactor")` / `code.includes("session_transfer")` fallbacks are broad, so a future code like `session_transfer_rate_limited` would map to `SESSION_TRANSFER_DISABLED` and hide the real error. I can see from the comment and tests this is a deliberate forward-compat heuristic, so this is just a flag rather than a change request.

src/server/auth-client.ts
```diff
@@ -3462,6 +3471,271 @@ export class AuthClient {
...
+        ),
+        null
+      ];
+    }
+
+    // Parse the raw JSON response ourselves — bypasses oauth4webapi's
+    // token_type allowlist which blocks "N_A".
+    let rawBody: any;
+    try {
+      rawBody = await httpResponse.json();
```
kishore7snehil, line 3990: `genericTokenEndpointRequest` hands back the Response without throwing on 4xx/5xx, so every server error flows through this `json()` call. When the body isn't JSON (a plain-text 500 or an HTML gateway page), this throws and the caller gets a generic "Failed to parse" with the status and the server's message gone.

Could we fall back to `text()` and fold `httpResponse.status` into the error when the body isn't JSON?

kishore7snehil (changes requested):

src/utils/session-transfer-helpers.ts
Piyush-85: Fixed. In `requestSessionTransferToken` (`auth-client.ts:3785`), after `resolveActorFromSession` returns an error, the SDK now checks whether the session has a refresh token. If it does, it calls the existing `#refreshTokenSet` and retries actor resolution with the fresh ID token. `resolveActorFromSession` itself remains a pure helper — the refresh lives at the caller level as you suggested. `ACTOR_UNAVAILABLE` is only thrown when there is genuinely no path to a usable actor token.

JSDoc on `SessionTransferTokenOptions.actor`, `Auth0Client.requestSessionTransferToken`, and `AuthClient.requestSessionTransferToken` all updated to reflect the new behaviour.

src/server/client.ts
Piyush-85: Fixed. Added a `describe("requestSessionTransferToken / buildSessionTransferRedirect")` block to `client.test.ts` (`line 676`) with five tests:

- **Success path** — verifies the result is returned and `authClient.requestSessionTransferToken` is called with the current session.
- **Throw on error** — verifies the `[error, null]` tuple is unwrapped and the error is re-thrown (`ACTOR_UNAVAILABLE` case).
- **Null session forwarded** — verifies `null` is passed when `getSession()` returns nothing.
- **`buildSessionTransferRedirect` redirect** — verifies `session_transfer_token` appears in the location header.
- **`buildSessionTransferRedirect` with organization** — verifies `organization` param is appended.

src/server/custom-token-exchange.test.ts
Piyush-85: Fixed. Added `describe("5.9: handleLogin forwards session_transfer_token to /authorize")` in `custom-token-exchange.test.ts` (`line 1745`). The test calls `authClient.handleLogin` with a login URL that carries `?session_transfer_token=stt_opaque_abc` and asserts the param appears unchanged in the resulting `/authorize` redirect URL.

EXAMPLES.md
Piyush-85: Fixed. Replaced the vague "No refresh, short-lived session" bullet with an explicit note explaining that Auth0 silently drops `offline_access` and recommending passing `scope: "openid profile"` explicitly (`EXAMPLES.md:5083`).

src/types/token-vault.ts
Piyush-85: Outdated — fixed in a prior commit and this thread is marked as such. The wording now matches the actual behaviour (refresh attempted if a refresh token is available, otherwise `ACTOR_UNAVAILABLE`).

EXAMPLES.md
Piyush-85: Outdated — fixed in a prior commit. Anchor is now `#error-handling-5`.

src/server/auth-client.ts
Piyush-85: Fixed. The success path now validates `rawBody.issued_token_type === TOKEN_TYPES.SESSION_TRANSFER_TOKEN` before constructing the result (`auth-client.ts:3967`). If it is missing or wrong the exchange fails with `EXCHANGE_FAILED` and the unexpected value is included in the message. The fallback default has been removed entirely.

src/utils/session-transfer-helpers.ts
Piyush-85: Fixed. After `new URL()` succeeds, `buildSessionTransferRedirectUrl` now checks `url.protocol !== "http:" && url.protocol !== "https:"` and throws `CustomTokenExchangeError(EXCHANGE_FAILED)` for any other scheme (`session-transfer-helpers.ts:119`).

EXAMPLES.md
```diff
@@ -4929,6 +4936,153 @@ const result = await auth0.customTokenExchange({
...
 ```
 
+## Session Transfer Token
+
+Session Transfer Token (STT) is CTE Release 2. An authenticated agent (for example a support engineer) establishes a web session **as the customer** in a target app — without the customer's password — so they can see and reproduce the customer's exact experience. The agent app requests a one-shot STT, then redirects the agent's browser to the target app's login URL carrying the token. Auth0 redeems the STT and establishes an ephemeral, device-bound session **as the customer, with the agent recorded in the `act` claim**. The access and ID tokens for that session carry `act`, exactly as in the Release 1 delegation flow.
+
+Almost all of the new SDK surface is on the **initiator** side (the app that requests the STT and builds the redirect): one method — `requestSessionTransferToken()` — plus a redirect helper — `buildSessionTransferRedirect()`. The **target** app barely changes: it does a standard OIDC Authorization Code login with `session_transfer_token` riding along in the `/authorize` query string.
+
+> [!IMPORTANT]
+<<<<<<< Updated upstream
```
kishore7snehil: These merge-conflict markers are still in the file, here in the IMPORTANT block and again in the Error Handling block just below. The other review threads were addressed but this one wasn't, and as-is it ships literal `<<<<<<<` and `>>>>>>>` text right in the middle of the published guide. The two sides differ, so it needs a real merge: keep the fuller two-client and feature-flag setup from one side and the `SETACTOR_REQUIRED` / `SESSION_TRANSFER_DISABLED` cases from the other.

EXAMPLES.md
```diff
@@ -4929,6 +4936,153 @@ const result = await auth0.customTokenExchange({
...
+  // and/or restrict destructive actions while impersonating.
+}
+```
+
+`act` carries the acting party's `sub` and optionally a few small custom fields the Action added (for example the agent's email or a reason). The `act` claim is **not** present on the STT exchange response — it only appears on the redeemed session's tokens.
+
+### Limitations
+
+- **Server-side only**: `requestSessionTransferToken` requires `client_secret` and can only be called server-side (confidential-client RWAs; SPAs/native are out of scope).
+- **One-shot, ~60 s**: The STT is opaque and must never be decoded, cached, or reused. Redeem it immediately.
```
kishore7snehil, line 5092: There's no mention that a caller should branch on `issued_token_type` rather than `token_type`. The exchange now rejects anything that isn't the session-transfer URN, so calling that out here would match what the code enforces.

kishore7snehil (changes requested):

EXAMPLES.md
Piyush-85: Fixed

EXAMPLES.md
Piyush-85: Fixed. Added a "Branch on issued_token_type, not token_type" bullet to the Limitations list (EXAMPLES.md:5073). It notes that token_type: "N_A" is informational only, that the SDK already validates issued_token_type === TOKEN_TYPES.SESSION_TRANSFER_TOKEN before returning (throwing EXCHANGE_FAILED if it doesn't match), and directs callers to  use TOKEN_TYPES.SESSION_TRANSFER_TOKEN in any downstream checks rather than token_type.

kishore7snehil (dismissed):
LGTM!

src/server/auth-client.ts
```diff
@@ -3712,6 +3722,308 @@ export class AuthClient {
...
+    );
+    if (tokenTypeError) {
+      return [tokenTypeError, null];
+    }
+
+    // Resolve actor — must come before any network call (client-side guard).
+    // If the session ID token is expired but a refresh token is available, attempt a
+    // silent refresh to get a fresh ID token before giving up with ACTOR_UNAVAILABLE.
+    let [actorError, actor] = resolveActorFromSession(session, options.actor);
+    if (actorError && session?.tokenSet?.refreshToken) {
```
nandan-bhat: When the agent ID token is expired we call `#refreshTokenSet` here and use the result to build a local `refreshedSession`, but we never persist that refreshed token set back to the session. `#refreshTokenSet` only returns the new tokens, it does not save them, and the `client.ts` wrapper does not call `saveToSession` after this either. If the tenant has `refresh token rotation` on (the default for confidential clients), this refresh consumes the old refresh token and the server issues a new one, but the session cookie still holds the old one. The agent's next `getAccessToken` then fails with an invalid grant and effectively logs them out, just from minting an STT. Can we persist the refreshed token set here (thread it out so the wrapper can save it, like `getAccessToken` does), or if that is awkward in this call shape, drop the refresh and keep the fail fast `ACTOR_UNAVAILABLE`? The `auth0-server-python` version persists after the refresh, so right now the two SDKs behave differently.

src/server/custom-token-exchange.test.ts
```diff
@@ -952,4 +955,817 @@ describe("Custom Token Exchange", () => {
...
+        );
+
+        expect(error).not.toBeNull();
+        expect(error?.code).toBe(
+          CustomTokenExchangeErrorCode.ACTOR_UNAVAILABLE
+        );
+        expect(result).toBeNull();
+      });
+
+      it("should return ACTOR_UNAVAILABLE when session ID token is expired", async () => {
```
nandan-bhat, line 1313: This is the only expired token test, and `makeAgentSession` does not set a `refreshToken`, so the refresh branch in `requestSessionTransferToken` (`auth-client.ts:3786`, gated on `session?.tokenSet?.refreshToken`) never runs. That means the whole refresh on expiry path is currently untested. Can we add a case with an expired ID token plus a refresh token present, mock `#refreshTokenSet` to return a fresh ID token, and assert the exchange succeeds using the refreshed token as the actor? That test would also pin down whatever we decide about persisting the refreshed tokens.

src/utils/session-transfer-helpers.ts
```diff
@@ -0,0 +1,180 @@
...
+ * point this mapping starts firing without any further change.
+ */
+export function mapSttServerError(
+  errorCode: string
+): CustomTokenExchangeErrorCode | null {
+  const code = errorCode.toLowerCase();
+  if (code === "setactor_required") {
+    return CustomTokenExchangeErrorCode.SETACTOR_REQUIRED;
+  }
+  if (code === "session_transfer_disabled") {
```
nandan-bhat, line 176: We map both `setactor_required` and `session_transfer_disabled` here, and there is an integration test that drives `setactor_required` all the way through `requestSessionTransferToken`, but I could not find a matching one for `session_transfer_disabled`. Both take the same path, so can we add a test that returns a `session_transfer_disabled` error from the token endpoint and asserts the mapped error code comes through? Cheap coverage for the second server error.

src/server/auth-client.ts
```diff
@@ -3712,6 +3722,308 @@ export class AuthClient {
...
+      ];
+    }
+
+    // Build the STT-specific audience: urn:{domain}:session_transfer
+    // Use this.domain (the effective domain for this request) so MCD tenants
+    // get the correct per-domain audience.
+    const audience = buildSessionTransferAudience(this.domain);
+
+    // Merge scopes (offline_access is silently suppressed by Auth0 for STT)
+    const finalScope = mergeScopes(DEFAULT_SCOPES, options.scope);
```
nandan-bhat: Small one, not blocking. We merge in `DEFAULT_SCOPES` which includes `offline_access`, and the comment right above says Auth0 silently drops it for STT. So we are sending a scope we know the server ignores. Could strip `offline_access` from the merged scope on this path so the request matches what actually happens, but leaving it as is with the comment is fine too.

src/server/auth-client.ts
Piyush-85: Fixed. requestSessionTransferToken now returns a 3-tuple [error, result, refreshedSession] instead of 2. refreshedSession is non-null only when the agent's expired ID token was refreshed to resolve the actor, and it carries the new token set (including a rotated refresh token, when refresh token rotation is enabled). Auth0Client.requestSessionTransferToken (client.ts) now persists it via saveToSession — routed through finalizeSession first so a configured beforeSessionSaved hook still runs, matching the pattern used by executeGetAccessToken. session.user is also updated from the refreshed ID token's claims, so the persisted session doesn't carry stale user data alongside the fresh tokens.

While fixing this we also added a Pages Router overload (requestSessionTransferToken(req, res, options)), following the same resolveRequestContext/getSessionFromAuthClient pattern getAccessToken uses, so the persisted saveToSession(session, req, res) call works correctly under both App Router and Pages Router.

src/server/custom-token-exchange.test.ts
Piyush-85: Fixed. Added a test with an expired session ID token + a refresh token present: mocks the refresh grant to return a fresh ID token, asserts the exchange succeeds using the refreshed token as the actor, and asserts the returned refreshedSession carries the rotated refresh token and updated claims. Added a companion test confirming refreshedSession is null when there's no refresh token to fall back on (falls through to ACTOR_UNAVAILABLE as before).

src/utils/session-transfer-helpers.ts
Piyush-85: Checked — this one was already covered before your comment landed. There's an integration test 5.5: server error mapping > should return SESSION_TRANSFER_DISABLED when feature flag is off in custom-token-exchange.test.ts that drives a session_transfer_disabled server response all the way through requestSessionTransferToken, plus a unit test for mapSttServerError("session_transfer_disabled") directly in session-transfer-helpers.test.ts. No change needed here.

src/server/auth-client.ts
Piyush-85: Fixed. Added DEFAULT_STT_SCOPES (src/utils/constants.ts) — the same as DEFAULT_SCOPES minus offline_access — and used it for the STT-specific scope merge instead of DEFAULT_SCOPES. The request no longer asks for a scope the server silently ignores.

Piyush-85:
@kishore7snehil @nandan-bhat  Did another pass over the STT surface myself and found and fixed a few more things:

- Refresh-on-actor-error was firing even for a bad explicit `actor` (not just a stale session token) — could waste a refresh call or throw a confusing `MfaRequiredError` for an unrelated input mistake. Now gated to only apply when no explicit actor was passed.
- `additionalParameters` could duplicate `organization`/`reason` in the request body since neither was in the reserved-params denylist. Added both.
- `buildSessionTransferRedirectUrl`'s https-only/localhost scheme guard had zero test coverage. Added direct unit tests for it.
- `requestSessionTransferToken` had no Pages Router overload, unlike every other method on `Auth0Client`. Added one, following the same pattern `getAccessToken` uses.
- Added a defense-in-depth MCD domain check inside `AuthClient.requestSessionTransferToken` itself, so it's safe even if called directly without going through `Auth0Client.getSession()`'s domain check.

18 new tests added, full suite passing, TypeScript clean.

nandan-bhat (approved):

PR #2742, opened 2026-07-09
fix(ci): prevent script injection in RL scanner workflows

tusharpandey13:
## Summary

The ReversingLabs scanner action and its calling workflow interpolated user- and context-controlled values directly into `run:` shell blocks via `${{ }}`. This allowed arbitrary shell command injection (CWE-94): a crafted input value could break out of the intended command and execute attacker-controlled shell, exposing secrets (`RLSECURE_LICENSE`, `SIGNAL_HANDLER_TOKEN`, AWS role, artifactory token) and tampering with the release pipeline.

The values are now passed through the step `env:` block and referenced as quoted shell variables, which the runner treats as data rather than script. This follows GitHub's [secure use reference](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#understanding-the-risk-of-script-injections).

## Changes

**`.github/actions/rl-scanner/action.yml`** (Run RL Scanner step)
- `inputs.artifact-path` → `ARTIFACT_PATH`
- `inputs.version` → `VERSION`
- `github.event.repository.name` → `REPOSITORY_NAME`
- `github.repository` → `REPOSITORY`
- `github.sha` → `COMMIT_SHA`
- All shell references quoted (`"$VAR"`).

**`.github/workflows/rl-secure.yml`**
- `tar -czvf ${{ inputs.artifact-name }} *` → `env: ARTIFACT_NAME` + `tar -czvf "$ARTIFACT_NAME" *`
- Scan-result echo: `${{ steps.rl-scan-conclusion.outcome }}` → `env: SCAN_OUTCOME` + quoted reference.

No behavioral change — identical values reach `rl-wrapper` and `tar`.

## Test plan

- [ ] YAML parses (verified locally)
- [ ] Trigger `rl-secure` workflow on a `release/*` branch and confirm the scanner runs and reports status as before
- [ ] Confirm build artifact tarball is created with the expected name

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of artifact packaging and security scan workflows.
  * Standardized handling of artifact names, repository details, commit information, and scan outcomes.
  * Added safer quoting and environment variable usage to prevent workflow failures and ensure consistent scanner invocation.


.github/workflows/rl-secure.yml
Piyush-85: suggestion
            artifact-path: "${{ github.workspace }}/${{ inputs.artifact-name }}"

The $(pwd) here won't work after this change. The value now goes through env: ARTIFACT_PATH and is read as "$ARTIFACT_PATH", and bash doesn't evaluate $(...) inside an expanded variable — so the path stays the literal $(pwd)/<name> and the [ ! -f "$ARTIFACT_PATH" ] check fails with "Artifact not found". Using github.workspace gives the absolute path without needing shell evaluation.

tusharpandey13: fixed

Piyush-85 (approved):

PR #2747, opened 2026-07-13
Adding CTE session transfer token example

Piyush-85:
- [x] I have added documentation for all new/changed functionality 

### 📋 Changes

Adds a **Session Transfer Token (STT)** demo to the `examples/with-cte` example — the initiator flow where an authenticated agent establishes a web session **as a customer** in a target app, without the customer's password.

> **Depends on** the SDK implementation PR (`feat/cte-session-transfer-token`). This branch is based on it.

**What this example validates**

- `auth0.requestSessionTransferToken()` mints a one-shot STT, sourcing the actor from the agent's session ID token.
- `auth0.buildSessionTransferRedirect()` builds the redirect that carries the STT to the target app's login URL.
- The target app redeems the STT with no login/MFA/consent prompt and establishes an ephemeral session **as the customer, with the agent recorded in the `act` claim** (`sub` = customer, `act.sub` = agent, no refresh token).
- The two-client setup and prerequisites (feature flag, delegated-access config, non-localhost callback) work end-to-end.

**How to run**

See the example README for setup and step-by-step instructions:
[`examples/with-cte/README.md`](https://github.com/auth0/nextjs-auth0/blob/feat/cte-stt-example/examples/with-cte/README.md)

### 📎 References

- SDK implementation PR: `feat/cte-session-transfer-token`
- CTE Impersonation via Session Transfer Token (STT) — Auth SDK Requirements (Confluence, SDKREQ-379)
- Session Transfer Token (STT) Testing (Confluence)

### 🎯 Testing

Manual, end-to-end against a live tenant following [`examples/with-cte/README.md`](https://github.com/auth0/nextjs-auth0/blob/feat/cte-stt-example/examples/with-cte/README.md). Verified: silent STT redemption, session established as the impersonated customer with `act.sub` = agent, and no refresh token issued.

## Summary by CodeRabbit

- **New Features**
  - Added a Session Transfer Token (STT) demo flow, including a new dashboard entry point and an STT page for requesting and transferring one-shot, short-lived sessions.
  - Implemented a server-side STT exchange endpoint with robust request validation and structured error responses.
  - Added a transfer form with optional advanced fields, redirect URL display, clipboard “Copy URL” action, and improved session-card refresh-token visibility (including impersonation behavior).

- **Documentation**
  - Expanded the Custom Token Exchange example to cover STT enablement, constraints, running locally instructions, and error handling.

- **Chores**
  - Added example-specific ignore rules via a new `.gitignore`.


src/server/auth-client.ts
```diff
@@ -3462,6 +3470,267 @@ export class AuthClient {
...
+   * );
+   * return redirect;
+   * ```
+   */
+  buildSessionTransferRedirect(
+    targetLoginUrl: string,
+    result: SessionTransferTokenResult,
+    opts?: { organization?: string }
+  ): NextResponse {
+    const url = new URL(targetLoginUrl);
```
nandan-bhat: `buildSessionTransferRedirect` builds the URL and attaches the `session_transfer_token` without validating `targetLoginUrl`. Since the STT is a single use credential, if this URL ever comes from untrusted input it can leak the token to another host. Can we enforce that the URL is absolute and uses `https` here (allowing `http` only for `localhost`/`127.0.0.1`/`[::1]` for local dev)? The doc comment says it must be trusted, but I think we should enforce it in code and not rely on the comment. The auth0-server-js version does this check.

src/utils/session-transfer-helpers.ts
```diff
@@ -0,0 +1,133 @@
...
+      ),
+      null
+    ];
+  }
+
+  // Guard against expired ID tokens — the server validates exp on the actor_token
+  // and rejects with "Invalid actor token: 'exp' claim timestamp check failed".
+  try {
+    const { exp } = jose.decodeJwt(idToken);
+    if (typeof exp === "number" && exp < Math.floor(Date.now() / 1000)) {
```
nandan-bhat, line 71: When the session ID token is expired we return `ACTOR_UNAVAILABLE` right away. But the JSDoc in a few places says the actor is `refreshed if expired`. Right now there is no refresh happening, so the docs and the code do not match. Two options: either add the refresh (use the session refresh token to get a fresh ID token, then only fail with `ACTOR_UNAVAILABLE` when there is no refresh token), or update the docs to stop saying `refreshed if expired`. I would prefer the refresh since that matches the STT requirement and what auth0-server-js does.

src/utils/session-transfer-helpers.ts
```diff
@@ -0,0 +1,133 @@
...
+/**
+ * Maps a server 400 error code to a typed `CustomTokenExchangeErrorCode` for STT failures.
+ * Returns `null` when the error code is not STT-specific (caller should fall through to EXCHANGE_FAILED).
+ */
+export function mapSttServerError(
+  errorCode: string
+): CustomTokenExchangeErrorCode | null {
+  if (
+    errorCode === "setactor_required" ||
+    errorCode.toLowerCase().includes("setactor")
```
nandan-bhat: This does a substring match (`includes("setactor")` / `includes("session_transfer")`) on the server error code. That feels risky because it can catch unrelated codes that happen to contain that text and label them wrongly. Can we match the exact codes only (`setactor_required` and `session_transfer_disabled`)? Anything else already falls through to `EXCHANGE_FAILED` with the raw cause preserved, so we do not lose information.

src/server/auth-client.ts
```diff
@@ -3462,6 +3470,267 @@ export class AuthClient {
...
+
+      if (this.useDPoP && dpopHandle && httpResponse.status === 400) {
+        const cloned = httpResponse.clone();
+        let body: any;
+        try {
+          body = await cloned.json();
+        } catch {
+          body = {};
+        }
+        if (body?.error === "use_dpop_nonce") {
```
nandan-bhat: I do not think this DPoP nonce retry actually works. We read `newNonce` from the header but never feed it back into the `dpopHandle` before calling `sendExchange()` again, so the resend goes out with the same state and will fail the same way. The comment even says we should let the handle learn the nonce, but the code does not do that. Can we either wire the nonce in properly, or drop this block and note that DPoP nonce retry is not handled for STT yet? As it stands it just adds an extra request.

src/server/auth-client.ts
```diff
@@ -3462,6 +3470,267 @@ export class AuthClient {
...
+    params.append("actor_token_type", actor.type);
+
+    if (options.organization) {
+      params.append("organization", options.organization);
+    }
+    if (options.reason) {
+      params.append("reason", options.reason);
+    }
+    if (options.additionalParameters) {
+      for (const [key, value] of Object.entries(options.additionalParameters)) {
```
nandan-bhat, line 3917: `additionalParameters` is appended without any denylist. A caller could pass something like `audience` or `actor_token` here and we would send a second copy of a param the SDK already manages. Can we skip/reject reserved keys (`subject_token`, `subject_token_type`, `audience`, `grant_type`, `actor_token`, `actor_token_type`, `scope`, `client_*`)? The core CTE path guards these already.

src/server/auth-client.ts
```diff
@@ -3462,6 +3470,267 @@ export class AuthClient {
...
+      ];
+    }
+
+    // STT never decoded, never cached — treat as opaque handle
+    return [
+      null,
+      parseSessionTransferTokenResponse({
+        access_token: rawBody.access_token,
+        issued_token_type:
+          rawBody.issued_token_type ??
```
nandan-bhat: We default `issued_token_type` to the session transfer URN when the server does not send it. I would avoid fabricating it, because if the server ever returns a non STT response we would mislabel it as an STT. Can we surface whatever the server actually returned (leave it empty/undefined when absent) so callers can branch on it honestly?

src/utils/session-transfer-helpers.ts
```diff
@@ -0,0 +1,133 @@
...
+export function parseSessionTransferTokenResponse(raw: {
+  access_token: string;
+  issued_token_type: string;
+  expires_in?: number;
+  token_type?: string;
+}): SessionTransferTokenResult {
+  return {
+    sessionTransferToken: raw.access_token,
+    issuedTokenType: raw.issued_token_type,
+    expiresIn: Number(raw.expires_in ?? 60),
```
nandan-bhat: Small one: `expiresIn` defaults to `60` when the field is missing. Since this value is informational, I think it is better to surface exactly what the server sent rather than guess a number. Not blocking, just flagging.

src/server/auth-client.ts
Piyush-85: Fixed. `buildSessionTransferRedirectUrl` (`session-transfer-helpers.ts:119`) now enforces `https` for all hosts, with `http` permitted only for `localhost`, `127.0.0.1`, and `[::1]` — matching the auth0-server-js behaviour. Any other scheme (e.g. `javascript:`, `data:`, `ftp:`) throws `EXCHANGE_FAILED` before the STT is ever attached to the URL.

src/utils/session-transfer-helpers.ts
Piyush-85: Fixed in the parent branch (`feat/cte-session-transfer-token`). After `resolveActorFromSession` returns an error, `requestSessionTransferToken` (`auth-client.ts:3785`) checks whether the session has a refresh token. If it does, it calls the existing `#refreshTokenSet` and retries actor resolution with the freshened ID token. `resolveActorFromSession` itself stays a pure helper — the refresh lives at the caller level. `ACTOR_UNAVAILABLE` is only thrown when there is genuinely no path to a usable actor token (no refresh token, or the refresh itself fails).

src/utils/session-transfer-helpers.ts
Piyush-85: Fixed. `mapSttServerError` now matches `setactor_required` and `session_transfer_disabled` exactly (case-insensitive comparison against the lowercased code). The `includes()` fallbacks are removed — a code that merely contains those substrings now falls through to `null`, which surfaces as `EXCHANGE_FAILED` with the raw server cause preserved. Tests updated to assert the new behaviour.

src/server/auth-client.ts
Piyush-85: Already handled in the parent branch. The STT exchange path calls `withDPoPNonceRetry(sendExchange, ...)` — the same shared helper used by the CTE path. The `dpopHandle` is created once outside the closure; oauth4webapi's `DPoP()` handle learns and reuses the nonce from the `DPoP-Nonce` response header automatically, so the retried request carries a valid proof. When DPoP is disabled this is a single pass-through. No manual nonce threading needed.

src/server/auth-client.ts
Piyush-85: Already handled in the parent branch. The STT `additionalParameters` loop (`auth-client.ts:3849`) guards against a `reservedParams` set containing `subject_token`, `subject_token_type`, `audience`, `scope`, `actor_token`, `actor_token_type`, and `grant_type`. Reserved keys are silently skipped, so a caller cannot accidentally shadow SDK-managed params by passing them in `additionalParameters`.

src/server/auth-client.ts
Piyush-85: Already handled in the parent branch. The success path now validates `rawBody.issued_token_type === TOKEN_TYPES.SESSION_TRANSFER_TOKEN` before constructing the result (`auth-client.ts:3967`). If the field is missing or doesn't match the STT URN, the exchange fails with `EXCHANGE_FAILED` and the unexpected value is included in the error message. The fallback default is gone entirely.

src/utils/session-transfer-helpers.ts
Piyush-85: Fixed. `parseSessionTransferTokenResponse` now returns `expiresIn: undefined` when the server omits the field, rather than substituting 60. `SessionTransferTokenResult.expiresIn` is now `expiresIn?: number`. The value is informational and callers should not branch on it, so surfacing exactly what the server sent is cleaner. Tests updated accordingly.

examples/with-cte/app/api/stt/route.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  // Validate the target URL BEFORE minting — the STT is one-shot and short-lived,
+  // so we must not spend one on a request we can't complete. Also force scope to
+  // "openid profile": impersonated sessions cannot get a refresh token, so the
+  // SDK's default `offline_access` scope would make /authorize fail with
+  // `interaction_required`. Passing scope on the login URL overrides the SDK
+  // default when handleLogin forwards it to /authorize.
+  let targetUrl: URL;
+  try {
+    targetUrl = new URL(body.targetLoginUrl);
```
nandan-bhat, line 87: Here `targetLoginUrl` comes straight from the request body and we hand it to `buildSessionTransferRedirect`, which puts the one-shot STT into the redirect URL. The SDK now enforces `https` (and `http` only for localhost), so the scheme is safe. But a caller can still point this at any `https` host and walk off with a live impersonation token. For the example, can we add a small allowlist of trusted target origins (from an env var) and reject anything else with a `400`? At the very least a comment saying this must be app controlled would help readers not copy this pattern as is.

examples/with-cte/app/api/stt/route.ts
```diff
@@ -0,0 +1,123 @@
...
+          message: err.message,
+          cause:
+            err.cause && typeof err.cause === "object" && "code" in err.cause
+              ? {
+                  code: (err.cause as { code?: string }).code,
+                  message: (err.cause as { message?: string }).message
+                }
+              : undefined
+        },
+        { status: 400 }
```
nandan-bhat: Every `CustomTokenExchangeError` returns `400` here, including tenant side failures like `SESSION_TRANSFER_DISABLED` and `SETACTOR_REQUIRED`. Those are not really bad client input, they mean the tenant or the Action is misconfigured. Would it read better to map those to `500` and keep `400` only for `ACTOR_UNAVAILABLE` and the missing input cases? Not blocking for an example, it just makes the demo teach the right status codes.

examples/with-cte/README.md
```diff
@@ -76,12 +87,24 @@ pnpm dev
...
 4. Paste the access token as the subject token, set the type to `urn:ietf:params:oauth:token-type:access_token`, and click **Exchange token**.
 5. The response shows the new `accessToken`, `idToken`, `scope`, and — if your Action sets it — the `act` claim.
 
+### Session Transfer Token (`/stt`)
+
+> Requires a non-localhost callback (see Dashboard Setup step 4). Access the app via that URL, not `http://localhost:3000`.
+
+1. Log in as the **agent** and land on `/dashboard`.
+2. Click **Session Transfer** to go to `/stt`.
+3. Fill in the customer token, set the token type to your registered CTE profile (e.g. `urn:acme:legacy-token`), and set the target app's login URL (e.g. `https://your-domain/auth/login`). Submit.
```
nandan-bhat, line 104: Small doc note: can we add a one line caveat here that `targetLoginUrl` must be a trusted, app controlled value and should never come from untrusted input? The SDK docs for `buildSessionTransferRedirect` call this out, and since the STT is a single use credential it is worth repeating right where the reader is told to type a URL into the form.

Piyush-85: Fixed. Added a comment directly in the code snippet explaining that targetLoginUrl must be a trusted, app-controlled value, that it must never come from untrusted user input, and that the origin should be validated against an allowlist in production — with the reason (single-use credential leaks on wrong host). Mirrors the SDK's own JSDoc for buildSessionTransferRedirect.

examples/with-cte/app/api/stt/route.ts
Piyush-85: Fixed. SESSION_TRANSFER_DISABLED and SETACTOR_REQUIRED now return 500 — these indicate tenant or Action misconfiguration, not bad client input. ACTOR_UNAVAILABLE, MISSING_SUBJECT_TOKEN, and other client-side validation failures stay 400. EXCHANGE_FAILED also stays 400 as it covers upstream exchange errors where the client request was the trigger.

examples/with-cte/app/api/stt/route.ts
Piyush-85: Fixed. Added an ALLOWED_ORIGINS set read from a STT_ALLOWED_ORIGINS env var (comma-separated origins). When the var is set, any targetLoginUrl whose origin isn't in the list is rejected with a 400 before requestSessionTransferToken is ever called — so the STT is never minted against an untrusted host. When the var is absent the example runs in demo mode (all origins accepted), with a comment explaining the risk and what to do in production. A top-of-file security comment repeats the invariant for readers who copy the pattern.

nandan-bhat (dismissed):

examples/with-cte/app/dashboard/page.tsx
```diff
@@ -50,15 +50,40 @@ export default async function Dashboard() {
...
+              <dt className="font-medium text-gray-600">Refresh token</dt>
+              <dd className="font-mono text-xs text-gray-900">
+                {tokenSet.refreshToken ? "present" : "none"}
+              </dd>
+            </div>
           </dl>
 
+          {user.act && (
+            <div className="mb-6 rounded-lg border border-amber-200 bg-amber-50 px-4 py-3 text-xs text-amber-800">
+              <strong>Impersonation session.</strong> You are acting as{" "}
```
kishore7snehil: The banner reads "acting as `{user.sub}` on behalf of `{act.sub}`", but that's backwards. On the redeemed session `sub` is the customer and `act.sub` is the agent, so this says the customer is acting for the agent. Since the example is here to teach the impersonation model, the direction matters.

Shall we flip it to something like "Logged in as `{user.sub}` (customer), impersonated by `{act.sub}` (agent)"?

examples/with-cte/app/api/stt/route.ts
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  try {
+    const result = await auth0.requestSessionTransferToken({
+      subjectToken: body.subjectToken,
+      subjectTokenType: body.subjectTokenType,
+      reason: body.reason,
+      organization: body.organization,
+      scope: body.scope,
+      actor: body.actor
```
kishore7snehil: `actor` and `scope` are read straight from the request body and passed to the SDK, but the form never sends either. So the route accepts fields the UI doesn't use, and letting the client set `actor` is the wrong thing to show in a sample: the actor should come from the agent's own session, not from whatever the caller posts. The server validates it, so it's not exploitable here, but people copy examples.

Can we drop `actor` from the accepted body and source it from the session, or if the override is meant to be shown, add a line saying production apps must not let the client control it?

examples/with-cte/app/api/stt/route.ts
```diff
@@ -0,0 +1,153 @@
...
+
+import { auth0 } from "@/lib/auth0";
+
+// SECURITY: targetLoginUrl must be a trusted, app-controlled value — never pass
+// untrusted user input here. In production, validate the origin against an
+// allowlist (e.g. STT_ALLOWED_ORIGINS env var) before calling requestSessionTransferToken.
+// The STT is a single-use credential; attaching it to a wrong host leaks the token.
+const ALLOWED_ORIGINS = process.env.STT_ALLOWED_ORIGINS
+  ? new Set(process.env.STT_ALLOWED_ORIGINS.split(",").map((s) => s.trim()))
+  : null; // null = no allowlist configured (demo mode — all origins accepted)
```
kishore7snehil: When `STT_ALLOWED_ORIGINS` is unset this accepts any origin, which is a reasonable demo default and it's clearly commented. The one risk is someone copying this to production without setting the env var, and then the STT can be attached to any host they pass in.

Would it be worth defaulting to just the app's own origin when the var is unset, so the copy-paste path is safe by default?

examples/with-cte/app/api/stt/route.ts
```diff
@@ -0,0 +1,153 @@
...
+    targetUrl.searchParams.set("scope", "openid profile");
+  }
+
+  try {
+    const result = await auth0.requestSessionTransferToken({
+      subjectToken: body.subjectToken,
+      subjectTokenType: body.subjectTokenType,
+      reason: body.reason,
+      organization: body.organization,
+      scope: body.scope,
```
kishore7snehil: There are two different `scope` values in play here, `body.scope` on the mint call and the `openid profile` forced onto the login URL below, and they're both just called "scope". A reader has to work out that one is for the exchange and one is for the target `/authorize`.

A word in the comment separating the two, or naming the login-URL one differently, would make it easier to follow.

kishore7snehil (changes requested):

examples/with-cte/app/dashboard/page.tsx
Piyush-85: Fixed. Changed the banner to "Logged in as `{user.sub}` (customer), impersonated by `{act.sub}` (agent)" — matching the actual claim direction on the redeemed session (`sub` is the customer, `act.sub` is the agent).

examples/with-cte/app/api/stt/route.ts
Piyush-85: Fixed. Removed both from the exchange call — `actor` and `scope` are no longer read from the request body at all. Narrowed the body type to only the fields the form actually sends (`subjectToken`, `subjectTokenType`, `targetLoginUrl`, `reason`, `organization`), with a comment explaining `actor` must come from the agent's own session, never from the request body.

examples/with-cte/app/api/stt/route.ts
Piyush-85: Fixed. When the env var is unset, `ALLOWED_ORIGINS` now defaults to the app's own origin (`APP_BASE_URL`) instead of accepting any origin — this example mints and redeems the STT in the same app, so that's the only origin it legitimately needs. Set `STT_ALLOWED_ORIGINS` explicitly for a multi-app deployment. The copy-paste-without-configuring path is now safe by default.

examples/with-cte/app/api/stt/route.ts
Piyush-85: Fixed. Removed `body.scope` entirely (see the fix above — it was never sent by the form anyway). Only one `scope` remains: the one forced onto the target login URL for `/authorize`. Reworded its comment to be explicit that it's the `/authorize` scope, not an exchange scope.

examples/with-cte/app/api/stt/route.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  if (!body.subjectToken || !body.subjectTokenType) {
+    return NextResponse.json(
+      {
+        code: "invalid_request",
+        message: "subjectToken and subjectTokenType are required."
+      },
+      { status: 400 }
+    );
+  }
```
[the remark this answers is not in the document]
Piyush-85: Fixed. `req.json()` accepted `null` (and arrays), which then threw when `body.subjectToken` was accessed — caught by the outer `catch (err: unknown)` and surfaced as a generic 500 instead of a clean `invalid_request` 400. Added an explicit non-object check right after parsing that returns `invalid_request`/400 for `null`, arrays, or any other non-object JSON value.

examples/with-cte/app/api/stt/route.ts
```diff
@@ -0,0 +1,172 @@
...
+            err.cause && typeof err.cause === "object" && "code" in err.cause
+              ? {
+                  code: (err.cause as { code?: string }).code,
+                  message: (err.cause as { message?: string }).message
+                }
+              : undefined
+        },
+        { status }
+      );
+    }
```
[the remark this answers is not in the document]
Piyush-85: Not applicable — `instanceof CustomTokenExchangeError` is the pattern the SDK's own `EXAMPLES.md` documents and uses in multiple places (including the Session Transfer Token section itself). `CustomTokenExchangeError` is a single, stably-exported class from `@auth0/nextjs-auth0/server`, so there's no class-identity risk here (e.g. no dual-package-hazard from bundling/versioning in this example). Keeping `instanceof` for consistency with the rest of the SDK's documented error-handling pattern.

amitsingh05667 (approved):

amitsingh05667 (approved):

kishore7snehil (changes requested):
LGTM!

kishore7snehil (approved):

PR #2753, opened 2026-07-17
fix: generate and store nonce in magic-link transaction for ID token …

Piyush-85:

### 📋 Changes

Magic-link passwordless flows (`send: 'link'`) now generate a nonce alongside `state` when building the authorization request. The nonce is:
  - included in `authParams` forwarded to Auth0 in the `/passwordless/start` request body, so Auth0 embeds it in the magic     link's authorization URL
  - stored in the transaction cookie alongside `state`

When the magic link is followed and the callback completes the authorization code exchange, `processAuthorizationCodeResponse` receives the stored nonce as `expectedNonce` and verifies it matches the `nonce` claim in the returned ID token. Without this fix, `expectedNonce` was `undefined` and the check was silently skipped, an attacker who extracted an ID token from a completed magic-link flow could replay it in a different session.

  OTP flows (`send: 'code'`) are not affected — they do not use an authorization code callback and are not impacted by this change.

  ### 📎 References

  - Identified during cross-review of auth0/auth0-server-python#134
  - RFC 3986 / OpenID Connect Core §3.3 nonce binding requirement

  ### 🎯 Testing

  New unit test in `passwordless-server.flow.test.ts`:
  - Asserts `nonce` is present in `authParams` sent to Auth0
  - Decrypts the transaction cookie and asserts the stored `nonce` matches the value forwarded to Auth0

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved magic-link sign-in security by including a nonce in the sign-in request and validating it during callback processing.
  * If the returned ID token nonce doesn’t match the expected value, the flow now fails safely without creating a session cookie.
* **Tests**
  * Added/extended coverage to verify the nonce is present in the `/passwordless/start` payload, stored in transaction state, and correctly enforced during callback nonce mismatch scenarios.

### DIFF
--- a/src/server/auth-client.test.ts
+++ b/src/server/auth-client.test.ts
@@ -4346,6 +4346,67 @@ ca/T0LLtgmbMmxSv/MmzIg==
       expect(transactionCookie!.maxAge).toEqual(0);
     });
 
+    it("should reject the callback when the ID token's nonce does not match the transaction's stored nonce", async () => {
+      const state = "transaction-state";
+      const code = "auth-code";
+
+      const secret = await generateSecret(32);
+      const transactionStore = new TransactionStore({ secret });
+      const sessionStore = new StatelessSessionStore({ secret });
+      const authClient = new AuthClient({
+        transactionStore,
+        sessionStore,
+
+        domain: DEFAULT.domain,
+        clientId: DEFAULT.clientId,
+        clientSecret: DEFAULT.clientSecret,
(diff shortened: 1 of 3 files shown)


tusharpandey13:
Looks good, but can we add a test to check callback with mismatched ID-token nonce since we are now checking the nonce?

tusharpandey13 (approved):

PR #2778, opened 2026-07-28
fix: preserve ports for insecure issuer URLs

### LINKED ISSUE #2777: normalizeDomain() silently drops non-standard ports from full issuer URLs
### Description

When `allowInsecureRequests` is enabled, `normalizeDomain()` accepts a full HTTP issuer URL with a non-standard port but silently removes the port from the constructed issuer. This breaks OIDC discovery for local providers running on a non-default port.

### Reproduction

```ts
normalizeDomain("http://localhost:8085/realm", {
  allowInsecureRequests: true
});
```

### Expected

```ts
{
  domain: "localhost",
  issuer: "http://localhost:8085/realm"
}
```

### Actual

```ts
{
  domain: "localhost",
  issuer: "http://localhost/realm"
}
```

### Root cause

`new URL(value).hostname` intentionally excludes the port. `normalizeDomain()` uses that hostname to construct `rawIssuer`, but never preserves `url.port`. The bare-hostname validation rejects ports, while full URL input bypasses that validation and loses its port silently.

This is related to #2626 / #2630: those changes allow `localhost` when `allowInsecureRequests` is true, but do not preserve an explicit port.

### Proposed fix

Capture `url.port` while parsing a full URL and append it when constructing the issuer, only when `allowInsecureRequests` is enabled. This retains the existing production behavior and permits local HTTP issuer URLs such as `http://localhost:8085/realm`.

ewuerger:
Fixes #2777

`URL.hostname` excludes the port, so `normalizeDomain()` previously rebuilt a full insecure issuer URL without its explicitly supplied non-standard port.

Capture `URL.port` during parsing and append it only when `allowInsecureRequests` is enabled. This preserves the existing production behavior while allowing local issuer URLs such as `http://localhost:8085/realm`.

Tested with `pnpm test:unit` (81 files, 1,836 tests).

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved custom ports in issuer URLs when insecure requests are enabled.
  * Improved local development connectivity for issuers using non-standard ports.
* **Tests**
  * Added coverage verifying that issuer ports remain intact during URL normalization.

### DIFF
--- a/src/utils/normalize.test.ts
+++ b/src/utils/normalize.test.ts
@@ -282,6 +282,16 @@ describe("normalize.ts", () => {
         });
       });
 
+      it("should preserve ports for insecure issuer URLs", () => {
+        const result = normalizeDomain("http://localhost:8085/realm", {
+          allowInsecureRequests: true
+        });
+        expect(result).toEqual({
+          domain: "localhost",
+          issuer: "http://localhost:8085/realm"
+        });
+      });
+
       it("should accept URLs with paths (e.g. Okta custom auth servers) without adding trailing slash", () => {
         // RFC 8414 does not require a trailing slash. Path-based issuers must be
         // preserved verbatim so they match the provider's discovery document exactly.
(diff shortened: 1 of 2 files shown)


Piyush-85 (approved):

Piyush-85:
@ewuerger Thanks for your contribution 
LGTM !

PR #2782, opened 2026-07-29
fix: prevent session-cookie accumulation (stale __session__N chunks, MFA token dedup)

Piyush-85:

### 📋 Changes

Follow-up to the transaction-cookie accumulation fix, addressing the remaining `431 Request Header Fields Too Large` causes that originate from the **session** side rather than transaction cookies.

**1. Deterministic deletion of stale chunked-session cookies (`cookies.ts`)**
  - `setChunkedCookie` and `deleteChunkedCookie` now clear a fixed chunk-index range (`__0 .. __{MAX_CHUNKS-1}`, `MAX_CHUNKS = 5`) instead of scanning the request cookie snapshot.
  - A concurrent request or a second tab can write a higher-index `__session__N` chunk that is absent from the current request's cookie snapshot. The previous snapshot-based scan left those chunks orphaned, so they lingered in the browser, inflated the request header, and could corrupt subsequent reads. Deleting a deterministic range removes them regardless of what this request happened to see. The browser ignores deletions for cookies that do not exist.
  - Legacy `{name}.{index}` chunks still use the snapshot scan (they are no longer written, so their count is only whatever an older SDK left behind).

  **2. Deduplicate MFA step-up access tokens (`auth-client.ts`)**
  - Each MFA step-up appended another full access-token set for the same audience, growing the session cookie without bound (eventually a 431).
  - The token set is now replace-or-append keyed on audience + scope (matching findAccessTokenSet's lookup dimensions), so repeated identical step-ups replace in place while differently-scoped step-ups for the same audience are both retained.

  ### 📎 References

  - Stacked on top of #2748 
  - Addresses `431 Request Header Fields Too Large` reports

  ### 🎯 Testing

  **Automated:** covered by unit tests:
  - `src/server/chunked-cookies.test.ts` — deterministic chunk-range deletion; orphaned higher-index chunks removed against a stale snapshot; legacy path unchanged.
  - `src/server/mfa-server.flow.test.ts` — MFA step-up replaces the same-audience token set instead of appending.
  - `src/server/mfa-server.flow.test.ts `— MFA step-up replaces a same-audience/same-scope token set instead of appending; retains separate entries for the same audience at different scopes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented repeated step-up MFA from growing session cookies by de-duplicating cached access tokens per audience and normalized scope.
  * Improved chunked-cookie cleanup with deterministic removal of a fixed chunk range, including orphaned higher-index chunks.
  * Enhanced legacy session cookie migration/cleanup to ensure correct deletion behavior and cookie attributes during chunk operations.
* **Tests**
  * Updated and expanded cookie chunking and stateless session caching tests to reflect deterministic cleanup and new MFA caching behavior.


Piyush-85:
@coderabbit review

Piyush-85:
@coderabbitai full review

src/server/cookies.ts
```diff
@@ -241,18 +248,23 @@ export function setChunkedCookie(
...
-    // Remove any previously stored chunks for this cookie name
-    getAllChunkedCookies(reqCookies, name).forEach((cookieChunk) => {
-      deleteCookie(resCookies, cookieChunk.name, {
+    // When we are writing a non-chunked cookie, remove any previously stored
+    // chunks for this cookie name. Delete a deterministic index range rather
+    // than scanning `reqCookies` — a concurrent request/tab may have written a
+    // higher-index chunk that this request's cookie snapshot does not include,
+    // which a snapshot-based scan would leave orphaned. The browser ignores
+    // deletions for cookies that do not exist.
+    for (let i = 0; i < MAX_CHUNKS; i++) {
```
Piyush-85, line 257: Core of the session-side 431 fix. We delete a fixed `__0 .. __{MAX_CHUNKS-1}` range instead of scanning `reqCookies`. The snapshot only reflects what *this* request received, so a chunk written by a concurrent request or another tab would be missed by a scan and left orphaned in the browser, inflating the request header until 431. Deleting a deterministic range clears it regardless of what this request saw; the browser ignores deletions for cookies that don't exist. Same reasoning applies to the chunked-branch loop at line 294 and `deleteChunkedCookie` at line 402.

src/server/cookies.ts
```diff
@@ -157,6 +157,13 @@ const MAX_CHUNK_SIZE = 3500; // Slightly under 4KB
 const CHUNK_PREFIX = "__";
 const CHUNK_INDEX_REGEX = new RegExp(`${CHUNK_PREFIX}(\\d+)$`);
 const LEGACY_CHUNK_INDEX_REGEX = /\.(\d+)$/;
+// Upper bound on chunk indices to clear when a chunked cookie shrinks (or is
+// replaced by a single cookie). 5 × 3500 = 17,500 bytes — far beyond any real
+// session, so no valid chunk is ever missed. Deleting a deterministic index
+// range instead of scanning `reqCookies` avoids leaving orphaned chunks when a
+// concurrent request/tab wrote a higher-index chunk not present in this
+// request's (stale) cookie snapshot.
+const MAX_CHUNKS = 5;
```
Piyush-85, line 166: `MAX_CHUNKS = 5` (5 x 3500 = 17,500 bytes) bounds the deterministic delete range. This is intentional: a session is not expected to exceed 5 chunks, so the write path is not bounded by this constant. Documenting the assumption for reviewers.

src/server/auth-client.ts
```diff
@@ -4903,15 +4903,36 @@ export class AuthClient {
...
+    // Replace an existing token for the same audience AND scope, or append a
+    // new one. Without this, each MFA step-up appends another full token set —
+    // growing the session cookie unbounded (and eventually a 431). The key is
+    // audience + scope (not audience alone) to match findAccessTokenSet, which
+    // deliberately holds multiple same-audience token sets distinguished by
+    // scope; keying on audience alone would evict a differently-scoped token.
+    const normalizeScope = (scope?: string) =>
+      (scope ?? "").trim().split(/\s+/).filter(Boolean).sort().join(" ");
+    const newScope = normalizeScope(newAccessTokenSet.scope);
+    const existingIdx = session.accessTokens.findIndex(
```
Piyush-85: Each MFA step-up used to `push` a new token set for the audience, growing the session cookie unbounded (eventually 431). Now it's replace-or-append. The key is **audience + scope**, not audience alone, so it matches `findAccessTokenSet`, which deliberately keeps multiple same-audience token sets distinguished by scope. Keying on audience alone would evict a differently-scoped step-up token. Scope is normalized (sorted) so ordering doesn't cause false mismatches.

src/server/cookies.ts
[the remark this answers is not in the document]
Piyush-85: Intentional. A session is not expected to exceed 5 chunks (5 x 3500 = 17,500 bytes), so the write path is deliberately not bounded by MAX_CHUNKS;  the deterministic delete range covers every chunk we expect to write. Adding a hard rejection for oversized values would turn an unlikely edge case into a login failure, which is worse than the theoretical orphan. Leaving as-is.

Piyush-85:
@coderabbit full review

nandan-bhat (approved):

PR #2784, opened 2026-07-29
 feat: add connected-accounts disconnect and listing with orphan cookie cleanup

Piyush-85:

### 📋 Changes

Adds disconnect connected accounts  and list  connected accounts capability, and fixes a connection-token cookie leak that could grow the request header size until it hit an HTTP 431.

**New public methods on `Auth0Client`:**

- `disconnectAccount({ connection })` — unlinks all connected accounts for a connection via the My Account API, then prunes the matching cached connection tokens from the session. It is connection-scoped, not per-account.
    ```typescript
    await auth0.disconnectAccount({ connection: "google-oauth2" });
    ``` 

- `getConnectedAccounts()` — returns `ConnectedAccount[]` 
    ```typescript
    const accounts = await auth0.getConnectedAccounts();
    ``` 

**Login-hint support for multiple accounts on one connection:**

 - `ConnectionTokenSet` gains an optional login hint. When `getAccessTokenForConnection({ connection, login_hint })` is called, the login hint is forwarded to the federated-connection token exchange and stamped onto the resulting cached token. Subsequent calls match cached tokens on both connection and login hint, falling back to connection-only when no hint is supplied (preserving existing single-account behavior).
- This lets an app hold and refresh access tokens for several accounts on the same connection without one account's refresh overwriting a sibling's cached token. The hint is a token-fetch-time discriminator supplied on `getAccessTokenForConnection`; `connectAccount` itself does not take or persist a login hint, so the app is responsible for passing the same `login_hint` it uses to obtain each account's token.

    ```typescript
    const { token } = await auth0.getAccessTokenForConnection({
        connection: "google-oauth2",
        login_hint: "alice@example.com"
     });
    ```

**The 431 fix (stateless session store):**

- Connection tokens are stored one per account in positionally indexed cookies. When the account list shrank (e.g. after a disconnect), the trailing higher-index cookies were left behind, re-read into the session, and piled up over time until the request headers exceeded the 431 limit.
- The store now deletes any indexed connection-token cookie past the current array length. It only touches cookies matching the exact indexed shape and leaves any other similarly prefixed cookie alone. The fix lives in the session store, so App Router, Pages Router, and middleware all benefit.
- Failed connection-token exchange now clears the stale __FC cookie: when the refresh-token→connection-token exchange fails (FAILED_TO_EXCHANGE), the dead cached token for that account is pruned from the session (account-scoped by connection + login_hint) and the orphaned __FC cookie is deleted, instead of being left behind to bloat request headers 

### 📎 References

Addresses [GH-2450](https://github.com/auth0/nextjs-auth0/issues/2450) (HTTP 431 caused by connection-token cookie accumulation).

### 🎯 Testing

  Run the unit suite. Added coverage includes:

  - Orphan cookie cleanup in the session store: shrink, grow, empty, and non-indexed-cookie-preserved cases.
  - `disconnectAccount` and `getConnectedAccounts` facades across App Router and Pages Router paths, plus the `connectAccount` overloads.
  - Login-hint multi-account matching.
  - `listConnectedAccounts` pagination, field mapping (including orgId), and error paths.
  - End-to-end regression against the real stateless session store, asserting Set-Cookie deletions are emitted for orphaned indexed cookies.

## Summary by CodeRabbit

* **New Features**
  * Added connected-account listing and disconnection APIs with pagination support.
  * Added `login_hint` support for selecting and caching account tokens.
  * Introduced typed disconnect errors and public connected-account types.
* **Bug Fixes**
  * Improved session and cookie reconciliation, including orphaned token-cookie cleanup.
  * Preserved rotated refresh tokens and removed stale cached entries.
* **Documentation**
  * Expanded Connected Accounts examples and persistence guidance.
* **Tests**
  * Added coverage for account management, token handling, and cookie cleanup.

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -156,6 +156,8 @@
 - [Connected Accounts](#connected-accounts)
   - [`onCallback` hook](#oncallback-hook)
   - [`connectAccount` method](#connectaccount-method)
+  - [`getConnectedAccounts` method](#getconnectedaccounts-method)
+  - [`disconnectAccount` method](#disconnectaccount-method)
 - [Back-Channel Logout](#back-channel-logout)
 - [Session Expiry from the Upstream IdP](#session-expiry-from-the-upstream-idp)
 - [Combining middleware](#combining-middleware)
@@ -4277,7 +4279,14 @@ export const auth0 = new Auth0Client({
 
 ### `connectAccount` method
 
-In case you'd like to have more control over the connected accounts flow, a `connectAccount` method is also available on the Auth0 client instance. For example, you could mount a custom route to start the connected accounts flow, like so:
+In case you'd like to have more control over the connected accounts flow, a `connectAccount` method is also available on the Auth0 client instance. It accepts an object with the following properties:
+
+- `connection`: (required) the name of the connection to link the account with (e.g., `google-oauth2`, `facebook`).
+- `scopes`: (optional) the scopes to request from the Identity Provider during the connect flow.
+- `authorizationParams`: (optional) additional parameters passed to the authorization server. This is where a `login_hint` is supplied to pre-select which upstream account to connect (see below).
+- `returnTo`: (optional) the URL to redirect to after the account is connected.
+
+The method returns a `NextResponse` that carries the redirect and transaction cookies. For example, you could mount a custom route to start the connected accounts flow, like so:
 
 ```ts
 import { auth0 } from "@/lib/auth0";
@@ -4297,8 +4306,143 @@ export async function GET() {
 }
 ```
 
+#### Connecting a specific account with `login_hint`
+
+To connect a specific upstream account (for example, when a user wants to link more than one account on the same connection), pass a `login_hint` through `authorizationParams`. It is forwarded to the authorization server so the correct account is pre-selected during the connect flow:
+
+```ts
+import { auth0 } from "@/lib/auth0";
+
+export async function GET() {
+  const res = await auth0.connectAccount({
+    connection: "google-oauth2",
+    scopes: ["openid", "profile", "offline_access"],
+    authorizationParams: {
+      login_hint: "alice@example.com"
+    },
+    returnTo: "/connected"
+  });
+
+  return res;
+}
+```
+
+> [!NOTE]
+> The `login_hint` on `connectAccount` (an authorization-request parameter passed via `authorizationParams`) is distinct from the top-level `login_hint` on [`getAccessTokenForConnection`](#getting-access-tokens-for-connections) (a token-exchange parameter). Connecting an account and later retrieving a token for it are separate operations, so the hint is supplied in the place appropriate to each.
+
+#### Middleware and dynamic base URLs
+
+When calling from middleware, or when `APP_BASE_URL` is configured dynamically (as an array of allowed origins), pass the `req` object so the redirect and session are resolved from the request context:
+
+```ts
+import { NextRequest } from "next/server";
+
+import { auth0 } from "@/lib/auth0";
+
+export async function middleware(request: NextRequest) {
+  const res = await auth0.connectAccount(
+    { connection: "google-oauth2", returnTo: "/connected" },
+    request
+  );
+
+  return res;
+}
+```
+
 > [!IMPORTANT]  
 > You must enable `Offline Access` from the Connection Permissions settings to be able to use the connection with Connected Accounts.
+
+### `getConnectedAccounts` method
+
+The `getConnectedAccounts` method lists the current user's connected accounts from the [My Account API](https://auth0.com/docs/manage-users/my-account-api). It returns an array of `ConnectedAccount` objects, each with the following shape:
+
+- `id`: the unique identifier of the connected account (e.g., `cac_...`).
+- `connection`: the name of the connection the account is linked through.
+- `accessType`: the access type (typically `offline`).
+- `scopes`: the scopes granted for the connected account.
+- `createdAt`: ISO date string of when the account was connected.
+- `expiresAt`: (optional) ISO date string of when the connected account expires.
+- `orgId`: (optional) the organization ID the connected account is scoped to. Only present for accounts bound to an organization.
+
+Because the My Account API is the source of truth, this method also reconciles the session: any locally cached connection tokens whose connection is no longer present server-side are pruned, so stale tokens are not re-assembled on subsequent reads. As this may write cookies, call it from a context that can set them.
+
+#### On the server (App Router)
+
+```ts
+import { NextResponse } from "next/server";
+
+import { auth0 } from "@/lib/auth0";
+
+export async function GET() {
+  const accounts = await auth0.getConnectedAccounts();
+
+  return NextResponse.json({ accounts });
+}
+```
+
+> [!IMPORTANT]  
+> Server Components cannot set cookies. Reconciliation performed while calling `getConnectedAccounts()` in a Server Component will not be persisted. Call it from a Route Handler, Server Action, or middleware if you need the reconciled session to be saved.
+
+#### On the server (Pages Router) and middleware
+
+Pass the `req` and `res` objects so the reconciled session can be persisted to the response cookies:
+
+```ts
+import type { NextApiRequest, NextApiResponse } from "next";
+
+import { auth0 } from "@/lib/auth0";
+
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse
+) {
+  const accounts = await auth0.getConnectedAccounts(req, res);
+
+  res.status(200).json({ accounts });
+}
+```
+
+### `disconnectAccount` method
+
+The `disconnectAccount` method disconnects (unlinks) connected accounts for a given connection via the [My Account API](https://auth0.com/docs/manage-users/my-account-api). It revokes the connection server-side and removes the corresponding cached connection tokens from the session so they are not re-assembled on subsequent reads. It accepts an object with the following property:
+
+- `connection`: (required) the name of the connection to disconnect (e.g., `google-oauth2`, `facebook`).
+
+> [!NOTE]
+> Disconnect is connection-scoped: **all** accounts connected through the given connection are disconnected. Per-account disconnect is not currently supported because the My Account API keys connected accounts by `id` and does not expose the login hint used to disambiguate multiple accounts on the same connection.
+
+#### On the server (App Router)
+
+```ts
+import { NextResponse } from "next/server";
+
+import { auth0 } from "@/lib/auth0";
+
+export async function POST() {
+  await auth0.disconnectAccount({ connection: "google-oauth2" });
+
+  return NextResponse.json({ message: "Disconnected!" });
+}
+```
+
+#### On the server (Pages Router) and middleware
+
+Pass the `req` and `res` objects so the pruned session can be persisted to the response cookies:
+
+```ts
+import type { NextApiRequest, NextApiResponse } from "next";
+
+import { auth0 } from "@/lib/auth0";
+
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse
+) {
+  await auth0.disconnectAccount({ connection: "google-oauth2" }, req, res);
+
+  res.status(200).json({ message: "Disconnected!" });
+}
+```
 
 ## Back-Channel Logout
 
@@ -4675,7 +4819,27 @@ export const GET = async (req: NextRequest) => {
 You can retrieve an access token for a connection using the `getAccessTokenForConnection()` method, which accepts an object with the following properties:
 
 - `connection`: The federated connection for which an access token should be retrieved.
-- `login_hint`: The optional login_hint parameter to pass to the `/authorize` endpoint.
+- `login_hint`: (optional) The login hint identifying which connected account to retrieve a token for. Provide it when a user has connected more than one account on the same connection so the correct one is selected; the token is then cached per `connection` + `login_hint`. When omitted, the token is matched and cached by `connection` alone (the default behavior).
+
+Without a login hint (default, single account per connection):
+
+```ts
+const token = await auth0.getAccessTokenForConnection({
+  connection: "google-oauth2"
+});
+```
+
+With a login hint (to target a specific account among several on the same connection):
+
+```ts
+const token = await auth0.getAccessTokenForConnection({
+  connection: "google-oauth2",
+  login_hint: "alice@example.com"
+});
+```
+
+> [!NOTE]
+> If the underlying refresh-token exchange fails (for example, the upstream refresh token was revoked), the stale cached connection token for that account is cleared from the session before the error is thrown, so it is not left behind on subsequent requests.
 
 ### On the server (App Router)
 
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -4358,7 +4358,7 @@ The `getConnectedAccounts` method lists the current user's connected accounts fr
 
 - `id`: the unique identifier of the connected account (e.g., `cac_...`).
 - `connection`: the name of the connection the account is linked through.
-- `accessType`: the access type (typically `offline`).
+- `accessType` (optional): the access type. Currently returned as `"offline"` by the My Account API when present.
 - `scopes`: the scopes granted for the connected account.
 - `createdAt`: ISO date string of when the account was connected.
 - `expiresAt`: (optional) ISO date string of when the connected account expires.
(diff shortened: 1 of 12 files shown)


src/server/session/stateless-session-store.ts
```diff
@@ -140,6 +141,30 @@ export class StatelessSessionStore extends AbstractSessionStore {
       );
     }
 
+    // The connection token cookies are indexed positionally (`__FC_0..n-1`). If
+    // the array has shrunk since it was last written (e.g. an account was
+    // disconnected), the trailing higher-index cookies would otherwise linger as
+    // orphans and be re-assembled into the session on the next read. Delete any
+    // `__FC_i` present in the request whose index is beyond the current length.
+    for (const cookie of this.getConnectionTokenSetsCookies(reqCookies)) {
```
Piyush-85, line 149: this loop is the `__FC` side of the 431 fix. the connection token cookies are positional (`__FC_0..n-1`), so when the array shrinks (someone disconnects an account) the higher-index cookies don't get overwritten by the write above. leave them and the browser keeps sending them, they get re-assembled into the session on the next read, and the request header grows until you hit 431. so here i just delete any `__FC_i` in the request whose index is past the current count.

the `reqCookies.delete()` is the easy-to-miss part: `storeInCookie` writes `reqCookies` too for read-after-write in the same middleware request, so a res-only deletion would still be re-read and re-assembled within that same request. deleting from both keeps it gone.

src/server/client.ts
```diff
@@ -1143,9 +1147,14 @@ export class Auth0Client {
...
-    // Find the connection token set in the session
+    // Find the connection token set in the session. When a login hint is
+    // provided, match on it too so multiple accounts can be connected for the
+    // same connection. Without a login hint, match on connection alone
+    // (preserving existing behaviour).
     const existingTokenSet = session.connectionTokenSets?.find(
-      (tokenSet) => tokenSet.connection === options.connection
+      (tokenSet) =>
+        tokenSet.connection === options.connection &&
+        (!options.login_hint || tokenSet.loginHint === options.login_hint)
```
nandan-bhat, line 1157: When no `login_hint` is passed this matches any entry for the connection, including one that was cached with a hint. In `getConnectionTokenSet` the `loginHint` is only stamped when a hint was supplied, so the replacement below writes back an entry with no hint on it.

I tried it with alice and bob both cached with their hints, then made one call without a hint:

```
after the unhinted call:
[{at: fc_exchanged, hint: null}, {at: fc_bob, hint: bob@example.com}]
```

alice's entry is gone. Her next call with her own hint finds no match, exchanges again and appends, so we end up with 3 entries and 3 `__FC` cookies for 2 accounts.

The PR description says one account's refresh will not overwrite a sibling's cached token, but that only holds if every call passes a hint. Can we treat "no hint" as its own key instead of a wildcard?

```ts
(options.login_hint
  ? tokenSet.loginHint === options.login_hint
  : !tokenSet.loginHint)
```

Sessions written before this change have no `loginHint` on any entry, so `!tokenSet.loginHint` behaves exactly like today. I ran the full suite with this applied and all 1892 tests still pass, including `matches on connection alone when no login hint is provided (back-compat)`.

src/server/auth-client.ts
```diff
@@ -3452,7 +3455,8 @@ export class AuthClient {
             Math.floor(Date.now() / 1000) +
             Number(tokenEndpointResponse.expires_in),
           scope: tokenEndpointResponse.scope,
-          connection: options.connection
+          connection: options.connection,
+          ...(options.login_hint ? { loginHint: options.login_hint } : {})
```
nandan-bhat, line 3459: Related to my comment on `client.ts`. Because `loginHint` is only stamped when a hint was passed, an unhinted exchange produces an entry with no hint, and that entry then replaces a hinted one in the session.

If the exchange response cannot tell us which account we actually got, then the safer option is the other side of it: when no hint is given, only match cached entries that also have no hint.

src/server/auth-client.ts
```diff
@@ -4331,6 +4335,201 @@ export class AuthClient {
...
+    );
+
+    const removed: ConnectedAccount[] = [];
+    for (const account of matching) {
+      const [deleteError] = await this.deleteConnectedAccount(
+        tokenSet,
+        account.id
+      );
+      if (deleteError) {
+        return [deleteError, null];
```
nandan-bhat, line 4525: This drops `removed`, but the deletes that already succeeded really happened on the server. So if the second `DELETE` fails, the first account is unlinked at Auth0 while its cached token and its `__FC_i` cookie stay in the session, which is the accumulation this PR is fixing.

I ran it with two Google accounts where only the second delete returns 429:

```
deleted server-side: [ 'cac_1' ]
returned removed:    null
returned error code: failed_to_delete
```

`client.ts` then throws at the `if (error)` check before the pruning block, so nothing is cleaned up and the app has no way to know that one account was actually disconnected.

Can we return the partial list along with the error, or attach it to `DisconnectAccountError`, and prune the connection before rethrowing? Pruning is connection scoped anyway, and the next `getAccessTokenForConnection` would just exchange again.

src/server/auth-client.test.ts
```diff
@@ -10055,6 +10108,467 @@ ca/T0LLtgmbMmxSv/MmzIg==
...
+        tokenSet,
+        "google-oauth2"
+      );
+
+      expect(removed).toBeNull();
+      expect(error?.code).toBe("failed_to_list");
+      expect(deletedIds).toEqual([]);
+    });
+
+    it("returns a FAILED_TO_DELETE error when a delete fails", async () => {
```
nandan-bhat, line 10431: Both failing delete tests use a single matching account, so the case where the first `DELETE` succeeds and the second one fails is not covered. That is the case where the partial state actually matters, since some accounts are gone server side but their cookies are still around. Worth adding a two account version.

src/server/auth-client.ts
```diff
@@ -4331,6 +4335,201 @@ export class AuthClient {
...
+    tokenSet: TokenSet,
+    id: string
+  ): Promise<[null, null] | [DisconnectAccountError, null]> {
+    try {
+      const url = new URL(
+        `/me/v1/connected-accounts/accounts/${encodeURIComponent(id)}`,
+        this.issuer
+      );
+
+      const fetcher = await this.fetcherFactory({
```
nandan-bhat, line 4454: This creates a new fetcher for every account and the loop above is sequential. The docs on `fetcherFactory` say each fetcher gets its own `oauth.DPoP` handle and that nonce state is isolated between instances, so with DPoP enabled every `DELETE` starts with an empty nonce cache and pays a `use_dpop_nonce` rejection plus a retry. That makes it `2N` requests instead of `N+1`, and each round also re-runs `ensureDpopValidated()` and discovery.

`listConnectedAccounts` already gets this right and uses one fetcher for all pages. Can we build the fetcher once in `disconnectAccount` and pass it in, or pass the handle through `options.dpopHandle`?

src/server/client.ts
```diff
@@ -1764,23 +1920,296 @@ export class Auth0Client {
...
+  ): Promise<ConnectedAccount[]> {
+    const { authClient, normalizedReq } = await this.resolveRequestContext(req);
+
+    const session = await this.getSessionFromAuthClient(
+      authClient,
+      normalizedReq
+    );
+
+    if (!session) {
+      throw new DisconnectAccountError({
```
nandan-bhat, line 2143: Naming question. Someone who only lists connected accounts and never disconnects anything still has to catch `DisconnectAccountError` and check `DisconnectAccountErrorCodes.FAILED_TO_LIST`. Both are new public exports in `src/errors/index.ts`, so renaming them later would be a breaking change.

Can we name the family after the resource instead, something like `ConnectedAccountsError` with `MISSING_SESSION`, `FAILED_TO_LIST` and `FAILED_TO_DELETE`? Or keep `DisconnectAccountError` for the delete path and add a separate error for listing.

src/server/session/stateless-session-store.ts
nandan-bhat, line 149: Scanning the request cookies here is the right call in my opinion. It cannot delete an index it never saw, and the mirror into `reqCookies` below covers read after write in the same request, so I could not find a way to lose a live cookie with this.

One thing to be aware of: #2782 goes the other way for `__session__*` and clears a fixed `0..4` range instead of the snapshot. If both land, the two cookie families in this same store will use opposite strategies for the same problem.

src/server/session/stateless-session-store.ts
```diff
@@ -140,6 +141,30 @@ export class StatelessSessionStore extends AbstractSessionStore {
...
+      const index = this.parseConnectionTokenSetCookieIndex(cookie.name);
```
nandan-bhat, line 150: The cleanup only recognises `__FC_<digits>`, but `get()` and `delete()` still pick up cookies with `startsWith("__FC")`. So a `__FC` prefixed cookie that decrypts fine but is not indexed would be read into `connectionTokenSets`, inflate `connectionTokenSetCount`, get rewritten into an indexed slot, and never be cleaned.

Not reachable today since the store only ever writes `__FC_<i>`, and the new test uses `__FCcustom=preserve-me` which fails to decrypt and is skipped. Using `parseConnectionTokenSetCookieIndex` on the read and delete paths too would keep all three in agreement.

src/server/client.ts
```diff
@@ -1764,23 +1920,296 @@ export class Auth0Client {
...
+  ): Promise<ConnectedAccount[]>;
+
+  /**
+   * Lists the connected accounts for the current user from the My Account API.
+   *
+   * The My Account API is the source of truth, so this also reconciles the
+   * session: any locally cached connection tokens (`connectionTokenSets`) whose
+   * connection is no longer present server-side are pruned, so they are not
+   * re-assembled into the session on subsequent reads. Because this reconciles
+   * (and may write cookies), call it from a context that can set cookies.
```
nandan-bhat, line 2117: This says the reconciliation will not be persisted, but I think the bigger risk is the rotated refresh token. `mintMyAccountToken` documents that minting can rotate the shared refresh token, and in a Server Component `saveToSession` swallows the write failure and only warns when `NODE_ENV` is `development`. The next refresh then hits reuse detection and logs the user out.

So this note is probably better as "do not call this from a Server Component" rather than describing it as losing the reconciliation. Same for the note in `EXAMPLES.md`.

EXAMPLES.md
```diff
@@ -4619,7 +4763,27 @@ export const GET = async (req: NextRequest) => {
 You can retrieve an access token for a connection using the `getAccessTokenForConnection()` method, which accepts an object with the following properties:
 
 - `connection`: The federated connection for which an access token should be retrieved.
-- `login_hint`: The optional login_hint parameter to pass to the `/authorize` endpoint.
+- `login_hint`: (optional) The login hint identifying which connected account to retrieve a token for. Provide it when a user has connected more than one account on the same connection so the correct one is selected; the token is then cached per `connection` + `login_hint`. When omitted, the token is matched and cached by `connection` alone (the default behavior).
```
nandan-bhat, line 4766: This does not say what happens when the user has more than one account connected and the call is made without a `login_hint`. Today it picks the first cached entry for the connection, which is whichever `__FC_i` the browser sent first, so the app cannot tell whose token it got. Worth spelling out here along with whatever we settle on for the matching in `client.ts`.

src/types/connected-accounts.ts
```diff
@@ -140,3 +140,55 @@ export interface CompleteConnectAccountResponse {
...
+   */
+  id: string;
+  /**
+   * The name of the connection associated with the connected account.
+   */
+  connection: string;
+  /**
+   * The access type, always 'offline'.
+   */
+  accessType?: string;
```
nandan-bhat, line 176: `accessType` is optional but the comment says it is always `offline`. Maybe type it as `accessType?: "offline"`, or soften the comment. `EXAMPLES.md` says "typically `offline`", so the two do not quite agree.

src/server/client.ts
Piyush-85: Fixed. Changed the predicate so "no hint" is its own key, not a wildcard:

```ts
options.login_hint
  ? tokenSet.loginHint === options.login_hint
  : !tokenSet.loginHint
```

An unhinted call now only matches entries that also have no hint, so it cannot select a hinted entry and later overwrite it. Back-compat holds - pre-multi-account sessions have no `loginHint` on any entry, so unhinted calls still match them. Added a regression test with two hinted accounts + one unhinted call that asserts both hints survive.

src/server/auth-client.ts
Piyush-85: No code change needed. The line itself is correct — an unhinted exchange should produce an unhinted entry, since we don't know which account it maps to. The harm you described came from N1's wildcard match. With N1 fixed, an unhinted entry can only replace another unhinted entry, so nothing hinted gets clobbered.

src/server/auth-client.ts
Piyush-85: Fixed on the caller side in `client.ts`. Moved the pruning block to run **before** the `throw error`. When the loop partially succeeds and then fails, the connection-scoped cached tokens are pruned first, then the error is rethrown. Since pruning is connection-scoped anyway, this is safe: a subsequent `getAccessTokenForConnection` would fail-exchange and prune the same tokens. Updated the existing test to assert prune-then-throw with a 2-connection session.

src/server/auth-client.test.ts
Piyush-85: Fixed. Added `returns a FAILED_TO_DELETE error even when the first delete succeeded (partial failure)`. Two google-oauth2 accounts, per-id stubs — `cac_1` returns 204, `cac_2` returns 429. Asserts the error surfaces with `FAILED_TO_DELETE` and 429 cause. Caller-side prune of the connection is covered by the updated test in `client.test.ts`.

src/server/auth-client.ts
Piyush-85: Fixed. Refactored `deleteConnectedAccount` to accept a pre-built `Fetcher<Response>` instead of a `tokenSet`. `disconnectAccount` builds one fetcher outside the loop and reuses it - same pattern `listConnectedAccounts` uses for pagination. With DPoP: N accounts = N+1 total requests instead of 2N. Also added an early return when `matching.length === 0`.

src/server/client.ts
Piyush-85: Fixed. Renamed:

- `DisconnectAccountError` → `ConnectedAccountsError`
- `DisconnectAccountErrorCodes` → `ConnectedAccountsErrorCodes`
- `buildDisconnectAccountErrorResponse` → `buildConnectedAccountsErrorResponse`

Codes stayed the same (`MISSING_SESSION`, `FAILED_TO_LIST`, `FAILED_TO_DELETE`). Not a public-API breaking change - this PR introduces the symbols, they don't exist on `main` yet. `DisconnectAccountOptions` (the input type) was kept because it's correctly named after the operation it configures.

src/server/session/stateless-session-store.ts
Piyush-85: Kept the snapshot scan. Added a comment explaining why `__FC` uses a different strategy from `__session__*`:

- `__FC` has no hard cap on the number of connected accounts a user can have.
- The count we shrink to is authoritative from the server-side delete/list, not a local decision.
- A deterministic range would either cap the account count or emit meaningless tombstones on every write.

The two cookie families solving the same-looking problem differently is intentional.

src/server/session/stateless-session-store.ts
Piyush-85: Fixed. `getConnectionTokenSetsCookies` now filters via `parseConnectionTokenSetCookieIndex`, matching the exact `__FC_<digits>` shape. All three call sites (`get()`, `delete()`, orphan sweep) now agree on what counts as ours. Also fixed two existing tests that used bogus `__FC.0` / `__FC.1` (dot separator, never a real wire format) — updated to correct `__FC_0` / `__FC_1`.

src/types/connected-accounts.ts
Piyush-85: Fixed. Narrowed:

- `ConnectedAccount.accessType?: string` → `accessType?: "offline"` — optional (matches API reality), typed to the observed literal.
- `CompleteConnectAccountResponse.accessType: string` → `accessType: "offline"` — required, typed to the literal.

EXAMPLES.md
Piyush-85: Fixed. Added a "Multi-account note" callout to the `getAccessTokenForConnection` docs. Explicitly documents: cache key is `connection + login_hint`; hinted call matches only same-hint entries; **unhinted call matches only entries with no hint** (not hinted ones — this is the isolation N1 fixed); back-compat holds for pre-multi-account sessions.

src/server/client.ts
Piyush-85: Fixed. Rewrote both the JSDoc and the `EXAMPLES.md` `> [!IMPORTANT]` note to lead with "Do not call this from a React Server Component" and explain the real failure mode: `mintMyAccountToken` can rotate the refresh token; Server Components can't set cookies; the write is silently dropped (warned only in `NODE_ENV=development`); the next refresh sends the old refresh token; Auth0's replay detection logs the user out.

nandan-bhat (dismissed):

nandan-bhat (approved):

PR #2840, opened 2026-09-08
fix: bind middleware to the Auth0Client instance

### LINKED ISSUE #2781: v4: auth0.middleware() crashes with "Cannot read properties of undefined (reading 'provider')" when invoked from proxy.ts under Next.js 16 (works from a Route Handler)
### Checklist

- [x] The issue can be reproduced in the [nextjs-auth0 sample app](https://github.com/auth0-samples/auth0-nextjs-samples/tree/main/Sample-01) (or N/A — reproduced in a minimal Next.js 16 app instead, see below).
- [x] I have searched the [issues](https://github.com/auth0/nextjs-auth0/issues) and have not found a suitable solution or answer (closest are #2436 and #2417, both different symptoms on earlier 4.x versions, both closed).

### Description

`auth0.middleware()` throws `TypeError: Cannot read properties of undefined (reading 'provider')` on every request when invoked from `proxy.ts` under Next.js 16. Reproduced identically in both `next dev` and `next start` (production build) — not a dev-only artifact.

The most diagnostic detail: **the identical `auth0.middleware()` call succeeds when dispatched from a Route Handler and fails when dispatched from `proxy.ts`.** Same `Auth0Client` instance, same request, two different call sites:

- `GET /api/auth/login` (dispatched from `app/api/auth/[auth0]/route.ts`, calling `auth0.middleware(req)` directly inside a `GET` handler) → works correctly. Clean `307` to the real Auth0 `/authorize` endpoint with correct `client_id`, PKCE `code_challenge`, `state`, `nonce`, and a properly-set transaction cookie.
- `GET /login` (an ordinary app page, covered by `proxy.ts`'s matcher) → crashes with the trace below, on every request, no exceptions.

I initially suspected this might be an Edge-vs-Node runtime difference, since the working call site is a Node.js Route Handler. I tested this directly by adding `export const runtime = "nodejs"` to `proxy.ts` — Next.js rejected it outright at build time: `Error: Route segment config is not allowed in Proxy file... Proxy always runs on Node.js runtime.` So there is no Edge/Node distinction to invoke here; `proxy.ts` is already Node-only under Next 16, and the crash happens anyway. I was not able to narrow the root cause further than that structural contrast.

Also ruled out: fake/placeholder Auth0 credentials (re-tested against a real, valid Auth0 tenant and real client credentials — identical crash); an SDK version predating Next 16 support (this is reproduced on 4.25.0, whose own `peerDependencies` list `next: "... || ^16.0.10"`); the SDK's env-var-lazy domain resolver (passed `domain` explicitly instead of relying on the `AUTH0_DOMAIN` fallback — identical crash).

**Workaround in use:** removed `proxy.ts` entirely. Route mounting still works via the Route Handler (`route.ts`), and route protection is handled by each protected layout calling `auth0.getSession()` directly rather than relying on the proxy. The only capability lost is automatic rolling-session refresh.

### Reproduction

1. Create a Next.js 16 app (App Router, Turbopack): `npx create-next-app@latest --typescript`
2. Add `@auth0/nextjs-auth0` (4.25.0) and configure a minimal `Auth0Client` in e.g. `auth.ts`:
   ```ts
   import { Auth0Client } from "@auth0/nextjs-auth0/server";
   export const auth0 = new Auth0Client();
   ```
3. Create `proxy.ts` at the project root per the v4 migration guide's Next 16 section:
   ```ts
   import { auth0 } from "./auth";
   export default auth0.middleware;
   export const config = {
     matcher: ["/((?!_next/static|_next/image|favicon.ico|sitemap.xml|robots.txt).*)"],
   };
   ```
4. Set real (or placeholder — both reproduce it) `AUTH0_DOMAIN`, `AUTH0_CLIENT_ID`, `AUTH0_CLIENT_SECRET`, `AUTH0_SECRET`, `AUTH0_BASE_URL` env vars.
5. `next build && next start`, then request any page not excluded by the matcher (e.g. `/`).
6. Expect: normal page response. Actual: `500`, with the trace below in the server log.

### Additional context

Full trace (paths are the local build output; the specific file/hash naturally varies by build):

```
TypeError: Cannot read properties of undefined (reading 'provider')
    at middleware (.next\server\chunks\[root-of-the-server]__<hash>._.js:31:138428)
    at handler (.next\server\chunks\[root-of-the-server]__<hash>._.js:34:1616)
    at <unknown> (.next\server\chunks\[root-of-the-server]__<hash>._.js:15:37004)
    at <unknown> (.next\server\chunks\[root-of-the-server]__<hash>._.js:15:36970)
    at async t2 (.next\server\chunks\[root-of-the-server]__<hash>._.js:15:33484)
```

Looking at `dist/server/client.js` (~line 195–243), the constructor's own comments describe an initialization-order dependency: `this.provider` is assigned via `new AuthClientProvider(...)`, and a factory closure captures `this` by reference intending to read `this.provider` only after that assignment completes, with an explicit static-mode workaround for the case where "the factory ... is called during AuthClientProvider construction, before this.provider is fully assigned." That reads like the class of race the crash matches, but I wasn't able to confirm it's the same code path triggering under the proxy specifically.

### nextjs-auth0 version

4.25.0 (also reproduced on 4.22.0 before upgrading)

### Next.js version

16.2.12 (Turbopack)

### Node.js version

24.14.1

sleitor:
## Description

Fixes #2781.

`auth0.middleware()` throws `TypeError: Cannot read properties of undefined (reading 'provider')` when it is invoked detached from its `Auth0Client` instance — most notably when re-exported directly as the default export from Next.js 16's `proxy.ts`:

```ts
import { auth0 } from "./auth";
export default auth0.middleware; // detached reference
```

### Root cause

`Auth0Client.middleware` is a regular class method. When it's passed around as a plain function reference (as opposed to being called as `auth0.middleware(req)`), it loses its `this` binding. Under JS strict-mode semantics (which apply to ES module code, including the SDK's compiled output), calling an unbound method results in `this === undefined` inside the method body. The very first thing `middleware()` does is read `this.provider`, so the failure is:

```
TypeError: Cannot read properties of undefined (reading 'provider')
```

This exactly matches the reported crash, and I confirmed the mechanism with a minimal repro:

```js
"use strict";
class Foo {
  constructor() { this.provider = { x: 1 }; }
  async middleware(req) { console.log(this.provider); return "ok"; }
}
const foo = new Foo();
const mw = foo.middleware;
await mw({}); // TypeError: Cannot read properties of undefined (reading 'provider')
```

This is why the issue reporter saw the identical call succeed from a Route Handler (`auth0.middleware(req)`, bound) and crash from `proxy.ts` (`export default auth0.middleware`, unbound) — the discrepancy is about how the method is invoked, not the runtime (Edge vs Node) or Next.js version.

### Fix

Bind `middleware` to the `Auth0Client` instance at the end of the constructor:

```ts
this.middleware = this.middleware.bind(this);
```

This makes `auth0.middleware` safe to export/pass around as a plain function reference, matching how a number of users naturally try to wire it up (`export default auth0.middleware`), without requiring them to remember to wrap it in a closure.

The documented patterns in `README.md`/`V4_MIGRATION_GUIDE.md` (`export async function proxy(request) { return auth0.middleware(request); }`) already work correctly today because they call the method as `auth0.middleware(...)`; this fix simply makes the SDK robust to the other common calling convention too.

## Testing

- Added a regression test in `src/server/client.test.ts` that detaches `client.middleware` from the instance (simulating `export default auth0.middleware`) and calls it as a bare function, asserting it no longer throws and still dispatches to the SDK's request handler.
- `pnpm run test:unit` — all 1969 tests pass.
- `pnpm run lint` — passes (`tsc --noEmit` x2 + `eslint`).
- `pnpm run build` — passes.

## References

Fixes #2781

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where authentication middleware could fail when exported or invoked without its original client context.
  * Middleware now works correctly with direct default exports in supported Next.js configurations.

### DIFF
--- a/src/server/client.test.ts
+++ b/src/server/client.test.ts
@@ -3246,6 +3246,24 @@ describe("Auth0Client", () => {
       expect(handlerSpy).toHaveBeenCalledTimes(1);
       expect(result).toBeInstanceOf(NextResponse);
     });
+
+    it("should not throw when middleware is detached from the Auth0Client instance (e.g. `export default auth0.middleware` in proxy.ts/middleware.ts)", async () => {
+      const authClient = await client["provider"].forRequest(new Headers());
+      const handlerSpy = vi
+        .spyOn(authClient, "handler")
+        .mockResolvedValue(NextResponse.next());
+
+      // Simulate Next.js 16's `proxy.ts` convention of re-exporting the method
+      // directly as the default export, e.g. `export default auth0.middleware;`,
+      // which detaches it from the `client` instance.
+      const detachedMiddleware = client.middleware;
+
+      const req = new Request("https://myapp.test/", { method: "GET" });
+      const result = await detachedMiddleware(req as any);
+
+      expect(handlerSpy).toHaveBeenCalledTimes(1);
+      expect(result).toBeInstanceOf(NextResponse);
+    });
   });
 
   describe("Pages Router Set-Cookie header handling", () => {
--- a/src/server/client.ts
+++ b/src/server/client.ts
@@ -815,6 +815,16 @@ export class Auth0Client {
     if (staticClient) {
       staticClient.provider = this.provider;
     }
+
+    // Next.js 16's `proxy.ts` convention (and some `middleware.ts` setups) commonly
+    // re-export this method directly, e.g. `export default auth0.middleware;`, rather
+    // than wrapping it in a closure like `(req) => auth0.middleware(req)`. When the
+    // method is invoked that way, it is detached from the `Auth0Client` instance, so
+    // `this` is `undefined` inside the method body (per JS strict-mode semantics for
+    // unbound method calls), causing a `TypeError: Cannot read properties of
+    // undefined (reading 'provider')` on every request. Binding here makes
+    // `auth0.middleware` safe to pass around as a plain function reference.
+    this.middleware = this.middleware.bind(this);
   }
 
   /**


Piyush-85:
Thanks for the fix @sleitor! LGTM

Piyush-85 (approved):

PR #2852, opened 2026-09-14
feat: add typed Experiment Center override params 

Piyush-85:

### 📋 Changes

- Adds `experiment_id`, `variation_id`, and `segment_id` as typed optional `string` fields on `AuthorizationParameters`
- Auth0 already honors these as `/authorize` query params - this exposes them as named, documented properties instead of requiring callers to use the index signature
- EXAMPLES.md table of contents links to the new guide `guides/experiment-center.md` added with full usage examples including segment targeting
- README gets an Experiment Center EA section with a minimal usage example and link to the guide
- No behavior changes - existing code using the index signature continues to work unchanged

### 📎 References

- Equivalent change in auth0-spa-js: https://github.com/auth0/auth0-spa-js/pull/1756

### 🎯 Testing

Two unit tests added inside the `handleLogin > authorization parameters` suite:

1. All three params (`experiment_id`, `variation_id`, `segment_id`) appear on the authorize URL when passed as login query params.
2. `experiment_id` and `variation_id` appear correctly and `segment_id` is absent when not provided.

## Summary by CodeRabbit

- **New Features**
  - Added support for overriding Auth0 Experiment Center assignments for individual login requests with experiment, variation, and optional segment identifiers.
  - Overrides can be provided through login URLs or interactive login options.

- **Documentation**
  - Added an Experiment Center guide covering login overrides, segmented experiments, testing, and production scenarios.
  - Updated the README and examples table of contents with Experiment Center documentation links.


guides/experiment-center.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Forcing a variant
+
+Pass `experiment_id` and `variation_id` via `authorizationParameters` on a per-call basis. The override applies to that request only -- the next login without these params reverts to normal server-side assignment.
+
+```ts
+// app/api/auth/login/route.ts
+import { auth0 } from "@/lib/auth0";
+
+export const GET = auth0.handleLogin({
```
nandan-bhat: `handleLogin` doesn't exist in `auth0` right ?

`auth0` is an instance of `Auth0Client`, but `handleLogin` actually exists in `AuthClient`.

guides/experiment-center.md
```diff
@@ -0,0 +1,43 @@
...
+    experiment_id: "<EXPERIMENT_ID>",
+    variation_id: "<VARIATION_ID>"
+  }
+});
+```
+
+When the experiment uses segment targeting, also pass `segment_id`:
+
+```ts
+export const GET = auth0.handleLogin({
```
nandan-bhat: Same as this: https://github.com/auth0/nextjs-auth0/pull/2852/changes#r4019430320

src/types/authorize.ts
```diff
@@ -104,6 +104,41 @@ export interface AuthorizationParameters {
    * @see https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
    */
   display?: "page" | "popup" | "touch" | "wap" | (string & {});
+  /**
+   * Forces a specific Experiment Center variant for this `/authorize` request
+   * instead of letting Auth0 assign one via its hash. Use together with
+   * `variation_id`.
+   *
+   * Pass per-call (on `handleLogin`) rather than at client construction time
+   * so the override does not bleed into silent `prompt=none` token-renewal
```
nandan-bhat: This reasoning does not hold for v4. This SDK refreshes tokens using the refresh token in the middleware, it does not do silent `prompt=none` calls (that is the `auth0-spa-js` model). So the `prompt=none` justification here is misleading. The actual reason to pass it per call is that setting it at construction time would apply the override to every login. Same text is in the guide and the README as well.

src/types/authorize.ts
```diff
@@ -104,6 +104,41 @@ export interface AuthorizationParameters {
...
+   * from a CI environment variable against a staging tenant. Do not hard-code
+   * this in shipped app code.
+   *
+   * **Production:** pass the variant decision from a feature-flag tool
+   * (e.g. LaunchDarkly) that has already decided which variant the user should
+   * see for this request.
+   *
+   * @see https://auth0.com/docs/customize/experiment-center/overview
+   */
+  experiment_id?: string;
```
nandan-bhat, line 149: Small note: these three already reach `/authorize` today through the index signature (`Additional authorization parameters`) just below, so this is adding types and docs rather than new behavior.That is fine and the IntelliSense is a nice win.

Flagging it mainly because it is also why the new tests pass without touching these typed fields (see my comment on the test file).

src/types/authorize.ts
```diff
@@ -104,6 +104,41 @@ export interface AuthorizationParameters {
...
+   * request. The override applies to this request only; the next login without
+   * these params reverts to normal server-side assignment.
+   */
+  variation_id?: string;
+
+  /**
+   * An optional segment identifier to pass alongside `experiment_id` and
+   * `variation_id` when forcing a variant for segment-targeted experiments.
+   */
+  segment_id?: string;
```
nandan-bhat, line 168: Can we clarify what `segment_id` does when it is passed together with `variation_id`? `variation_id` already forces a specific variant, so it is not clear what `segment_id` adds on top, or which one wins if they disagree.

A short note on when to use it and the precedence would help.

src/server/auth-client.test.ts
```diff
@@ -2137,6 +2137,95 @@ ca/T0LLtgmbMmxSv/MmzIg==
...
+
+          secret,
+          appBaseUrl: DEFAULT.appBaseUrl,
+
+          routes: getDefaultRoutes(),
+
+          fetch: getMockAuthorizationServer()
+        });
+        const loginUrl = new URL("/auth/login", DEFAULT.appBaseUrl);
+        loginUrl.searchParams.set("experiment_id", "exp_passkeys_onboarding");
```
nandan-bhat: These set the values as query params on the login URL, which already worked through the index signature before this PR, so they pass whether or not the new typed fields exist. They don't really exercise the type addition.

Could we add a case that goes through the typed `authorizationParameters` path (for example construction time config or `startInteractiveLogin`) so a regression on the new fields would actually fail ?

guides/experiment-center.md
Piyush-85: Correct, fixed it. `handleLogin` isn't on the public Auth0Client, I got that from refering the spa-js PR. Switched both examples to /auth/login query params and startInteractiveLogin. Fixed the README too.

guides/experiment-center.md
Piyush-85: Correct, fixed it. `handleLogin` isn't on the public Auth0Client, I got that from refering the spa-js PR. Switched both examples to /auth/login query params and startInteractiveLogin. Fixed the README too.

src/types/authorize.ts
Piyush-85: You're right, that was referred from spa-js and doesn't apply here since we refresh via refresh token in middleware. Updated the reason to: pass it per-login because setting it on the client applies to every login and breaks the A/B test. Fixed in the guide and README as well.

src/types/authorize.ts
Piyush-85: Agreed, it's types + docs only. Handled the real part in the test comment below.

src/types/authorize.ts
Piyush-85: Updated the JSDoc: variation_id picks the variation, segment_id says which segment it was defined under. Only needed for segment-targeted experiments, otherwise skip it.

src/server/auth-client.test.ts
Piyush-85: Right, those went through the index signature. Replaced them with tests using startInteractiveLogin({ authorizationParameters }) so they hit the typed path.

nandan-bhat (approved):

cschetan77 (approved):

